PR #3, opened 2021-05-14
don't attempt to run logic when just rendering display

bo-dun-1:


iamhimateja (approved):

iamhimateja:
> Reference

https://feathery.slack.com/archives/D01V0E5T2NM/p1620974145027200

PR #30, opened 2021-06-08
Fixed carry-over validity bug for Upfront

JakeKo:
After trying to provide an invalid VIN number, the validity of the input never got reset. This meant Upfront forms couldn't submit even if the field value was changed to a valid value.

### DIFF
--- a/src/Form.jsx
+++ b/src/Form.jsx
@@ -504,6 +504,14 @@ function Form({
             return;
         }
 
+        // For each field that changed, reset its validity
+        fieldKeys.forEach((fieldKey) => {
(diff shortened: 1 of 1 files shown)


bo-dun-1 (approved):
thanks for fixing this! did you test the change? and are we worried at all about performance issues since this is called per change?

JakeKo:
> did you test the change?

Yep I tested in hosted flows and the validity was reset correctly.

> and are we worried at all about performance issues since this is called per change?

I didn't encounter any issues with performance - I tried both the VIN field and the Google Maps fields.

PR #43, opened 2021-06-17
array rules logic support

bo-dun-1:


src/utils/formHelperFunctions.js
```diff
@@ -293,20 +300,37 @@ const nextStepKey = (nextConditions, metadata, steps, fieldValues) => {
...
+    // Don't propagate array rules if an equality rule matches or there are no array rules
+    if (newSequence.length === 0 || nextKey) newSequence = stepSequence;
+
+    if ([nextKey].includes(newSequence[0])) {
+        newSequence = newSequence.slice(1);
+    } else if (!nextKey && newSequence.length > 0) {
+        nextKey = newSequence[0];
+        newSequence = newSequence.slice(1);
+    }
+    return [nextKey, newSequence];
```
JakeKo, line 333: nit: prefer to use `{ nextKey, newSequence }` instead of arrays. When we enable typescript support, the array here won't be typed effectively. This also means when we add return values to this function, we don't need to worry about a specific ordering

src/utils/formHelperFunctions.js
```diff
@@ -293,20 +300,37 @@ const nextStepKey = (nextConditions, metadata, steps, fieldValues) => {
...
                 if (rulesMet) newKey = cond.next_step_key;
             }
         });
-    return newKey || defaultKey;
+
+    let nextKey = newKey || defaultKey;
+    // Don't propagate array rules if an equality rule matches or there are no array rules
+    if (newSequence.length === 0 || nextKey) newSequence = stepSequence;
+
+    if ([nextKey].includes(newSequence[0])) {
```
JakeKo, line 327: Wouldn't this just be `nextKey === newSequence[0]`

src/utils/__test__/formHelperFunctions.spec.js
```diff
@@ -476,11 +476,17 @@ describe('formHelperFunctions', () => {
             ];
 
             // Act
-            const actual = nextStepKey(conditions, {
```
JakeKo, line 479: Please add a test to check the newSequence that now gets returned

src/utils/formHelperFunctions.js
bo-dun-1: rip yes lol

JakeKo (approved):

PR #47, opened 2021-06-18
Enabled custom repeated rows

JakeKo:
Summary of the changes:

1. New hydration module to support reactive changes to the active step. Right now the hydration pipeline looks like this: raw step (direct from the server) -> inject rows (number is based on `fieldValues`) -> `activeStep` (which is fit to render). Other hydration transformations can be included in this pipeline. Benefits: automatic dimension calculations, automatic re-render on added and removed rows, cleaner code.
2. All the various methods that set a field value at some point have been updated to take into account repeated values. Most of the helper functions for these kinds of operations are in `formHelperFunctions`. Benefits: `formHelperFunctions` no longer sets any state on its own outside the Form component, just returns new values.


src/utils/hydration.js
```diff
@@ -0,0 +1,195 @@
...
+        } else if (field.row_index > repeatEnd) {
+            return [
+                {
+                    ...field,
+                    row_index:
+                        field.row_index +
+                        (repeatEnd - repeatStart + 1) * (repeatedRowCount - 1),
+                    row_index_end:
+                        field.row_index_end +
+                        (repeatEnd - repeatStart + 1) * (repeatedRowCount - 1)
```
bo-dun-1, line 73: nit: can we decompose some of this math

bo-dun-1 (changes requested):

bo-dun-1 (approved):

PR #63, opened 2021-06-23
Implemented file input auto fill

JakeKo:


src/utils/client.js
```diff
@@ -174,13 +160,24 @@ export default class Client {
         const url = `${API_URL}api/panel/step/submit/file/${userKey}/`;
 
         const formData = new FormData();
+        const normalizeFiles = (file) =>
+            file instanceof File ? file : file.path;
         servars.forEach((servar) => {
             if (servar.file_upload) {
-                formData.append(servar.key, servar.file_upload);
+                formData.append(
```
bo-dun-1, line 167: we can probably decompose this into a "getFileValue" function or something where it returns a file value if it exists. If the value exists, we append it to formData

bo-dun-1 (approved):

PR #78, opened 2021-07-01
firebase sms support

bo-dun-1:


src/utils/formHelperFunctions.js
```diff
@@ -330,14 +335,22 @@ function getFieldError(value, servar) {
 /**
  * Returns the list of DOM nodes corresponding to the field with the provided key.
```
JakeKo, line 336: Change the comment here

src/utils/formHelperFunctions.js
```diff
@@ -330,14 +335,22 @@ function getFieldError(value, servar) {
...
     const singleOrList = formRef.current.elements[fieldKey];
-    const elements =
+    let elements =
         singleOrList instanceof RadioNodeList
             ? Array.from(singleOrList)
             : [singleOrList];
-
-    return elements.filter((e) => e);
+    elements = elements.filter((e) => e);
+    if (index) elements = [elements[index]];
```
JakeKo, line 352: You should explicitly check for `index !== null` since this will fail if `index` is 0.

src/Form.jsx
```diff
@@ -263,6 +285,23 @@ function Form({
                 setRawActiveStep(newStep);
             }
 
+            newStep.servar_fields.forEach((field) => {
+                const servar = field.servar;
+                if (
+                    servar.type === 'phone_number' &&
+                    servar.metadata.send_sms_code
+                ) {
+                    newStep.buttons.forEach((button) => {
```
JakeKo, line 294: Should we be worried about weird behavior if we have multiple submit buttons?

PR #80, opened 2021-07-01
comment fixes

bo-dun-1:


src/Form.jsx
```diff
@@ -291,12 +291,14 @@ function Form({
                     servar.type === 'phone_number' &&
                     servar.metadata.send_sms_code
                 ) {
+                    let setCaptcha = false;
                     newStep.buttons.forEach((button) => {
```
JakeKo, line 295: I'd recommend `newStep.buttons.find(b => b.link === 'submit')` and then doing logic based on whether the return value is defined

JakeKo (approved):

src/Form.jsx
bo-dun-1: makes sense

PR #99, opened 2021-07-15
conditionally load firebase

bo-dun-1:


src/Form.jsx
```diff
@@ -94,6 +92,8 @@ function Form({
     const [stepSequence, setStepSequence] = useState([]);
     const [sequenceIndex, setSequenceIndex] = useState(0);
 
+    const [firebase, setFirebase] = useState(null);
```
JakeKo, line 95: Do we need to be careful about checking for null when we use `firebase`?

bo-dun-1: imo we'd rather hard fail if firebase isn't there and we expect it to be there, since that's indicative of a more serious breakage in our upstream assumptions

bo-dun-1: right now it'd be if we had a login field in a form that had sms authentication but the firebase integration is turned on. which shouldn't be allowed anyways

JakeKo (approved):

PR #129, opened 2021-08-02
Implemented conversion of files to blobs upon loading form

JakeKo:


src/Form.jsx
```diff
@@ -202,15 +201,24 @@ function Form({
         return newValues;
     };
 
-    const updateSessionValues = (session, fieldVals) => {
+    async function updateSessionValues(session, fieldVals) {
+        // Convert files of the format { url, path } to blobs
+        const blobs = (
```
bo-dun-1, line 206: this function runs in more places than initial load, so this probably isn't the best place to convert s3 urls (only on initial session load, so fetchSession might make more sense). also ideally loading these files doesn't need to be blocking so the form can load first while the files download.

bo-dun-1: it is also loading the files sequentially here, which also isn't ideal

src/utils/formHelperFunctions.js
```diff
@@ -464,6 +464,28 @@ function shouldElementHide({ fields, values, element }) {
         : !matchValues.includes(value);
 }
 
+/**
+ * Files coming from Feathery servers are provided as objects { path, url }
+ * We need to convert the url into a blob for standardized handling
+ * We recurse to handle repeated fields
+ */
+async function convertToBlob(maybeBlob) {
```
bo-dun-1, line 472: if we instead do conversion of only the session data fetched from feathery servers, we know all of those are urls and no need to do hacky type detection here

bo-dun-1 (changes requested):
do we not need to make any additional changes to handle blob data when rendering in a file field? are we still passing back URLs when submitting data back to the servers? can you also include your testing plan here (including testing correct values being passed back in user provided lifecycle functions)

src/Form.jsx
```diff
@@ -880,7 +895,7 @@ function Form({
         );
         let submitPromise = null;
         if (featheryFields.length > 0)
-            submitPromise = client.submitStep(featheryFields);
+            submitPromise = client.submitStep(featheryFields, filePathMap);
```
bo-dun-1, line 898: do we need to pass filePathMap directly from updateSessionValues since this is the same render and state may not have been updated yet?

bo-dun-1 (changes requested):

src/Form.jsx
JakeKo: I didn't see a way for it to be in the same render since loading session values and submitting would definitely happen distinctly

bo-dun-1: whoops didn't see that this was in submit - u right

bo-dun-1 (approved):

PR #140, opened 2021-08-12
Misc fixes for file uploaders

JakeKo:


src/Form.jsx
```diff
@@ -137,7 +137,7 @@ function Form({
         activeStep
     ]);
 
-    function addRepeatedRow(fieldValuesArg = fieldValues) {
+    function addRepeatedRow({ values }) {
```
JakeKo, line 140: This should default to `{ values = fieldValues }` since you call it with an empty object in the `onClick` handler

src/Form.jsx
```diff
@@ -151,19 +151,27 @@ function Form({
...
+        return {
+            values: updateFieldValues(updatedValues),
+            fieldIDs,
+            fieldKeys
+        };
     }
 
-    function removeRepeatedRow(repeatRowIndex, fieldValuesArg = fieldValues) {
-        if (isNaN(repeatRowIndex)) return;
+    function removeRepeatedRow({ index, values }) {
```
JakeKo, line 173: This should also probably default to `{ index, values = fieldValues }`

src/Form.jsx
```diff
@@ -1173,8 +1186,28 @@ function Form({
...
+                                    break;
+                                case 'remove_repeated_row':
+                                    data = removeRepeatedRow({
+                                        index: field.repeat
+                                    });
+                                    break;
+                                default:
+                                    break;
+                            }
+                            if (data) {
```
JakeKo, line 1203: `data` will always be defined here - you probably want to check for `data.fieldKeys.length > 0` if you're checking if any servars changed

bo-dun-1 (approved):

PR #141, opened 2021-08-13
Fixed width bug

JakeKo:


src/components/Bootstrap.jsx
```diff
@@ -63,17 +64,15 @@ function BootstrapField({
                 style={{
                     position: 'relative',
                     height: `${styles.field_height}${styles.field_height_unit}`,
-                    width: `${styles.field_width}${styles.field_width_unit}`,
-                    maxWidth: '100%'
+                    width: '100%'
```
bo-dun-1, line 67: can you check that both px and percent width units look good for these fields with the updated css? i remember there was a specific reason why i replicated the width definition at each level but i can't remember right now.

src/components/Bootstrap.jsx
```diff
@@ -63,17 +64,15 @@ function BootstrapField({
...
             >
                 <ReactForm.Control
                     id={servar.key}
                     pattern={pattern}
                     style={{
-                        maxWidth: '100%',
-                        height: `${styles.field_height}${styles.field_height_unit}`,
-                        width: `${styles.field_width}${styles.field_width_unit}`,
+                        height: '100%',
+                        width: '100%',
```
bo-dun-1, line 75: does 100% height play well with a label?

bo-dun-1 (approved):

PR #144, opened 2021-08-13
refactor integration initialization

bo-dun-1:


src/utils/init.js
```diff
@@ -61,28 +64,105 @@ function init(apiKey, options = {}) {
...
+        TagManager.initialize({
+            gtmId: gtm.api_key,
+            dataLayer: { userId: initState.userKey }
+        });
+    }
+
+    const fb = integrations.firebase;
+    if (fb) {
+        let firebase = global.firebase;
+        if (!firebase) {
```
JakeKo, line 95: `await` doesn't block execution outside its scope - I think you want to wait for this to finish before you continue on Line 112. You can wrap this all with a Promise:
```js
const firebase = await new Promise((resolve) => {
    if (!global.firebase) {
        await dynamicImport(...);
        ...
        resolve(global.firebase);
    } else {
        resolve(global.firebase);
    }
})

JakeKo (changes requested):

JakeKo (approved):

PR #169, opened 2021-08-24
make tooltip font size a little smaller

bo-dun-1:


src/utils/styles.js
```diff
@@ -621,9 +622,12 @@ function getFieldStyles(field) {
                 styles.setStyle('placeholderFocus', 'display', 'none');
             }
 
-            styles.apply('tooltip', 'font_size', (a) => ({
+            styles.apply('tooltipIcon', 'font_size', (a) => ({
                 width: `${a}px`
             }));
+            styles.apply('tooltip', 'font_size', (a) => ({
+                fontSize: `${a - 2}px`
```
JakeKo, line 629: Try replacing this with `fontSize: smaller` and see how that looks

bo-dun-1: https://github.com/feathery-org/feathery-react/pull/171

PR #185, opened 2021-09-08
Resolve highlighting last pasted digit

ckrueger6:
From thread: https://feathery.slack.com/archives/C02AYGFN03Y/p1631054670004100?thread_ts=1630618279.031800&cid=C02AYGFN03Y

### DIFF
--- a/src/elements/fields/PinInput.jsx
+++ b/src/elements/fields/PinInput.jsx
@@ -15,6 +15,7 @@ function SingleOtpInput({
     onChange,
     onKeyDown,
     onInput,
+    paste,
     onPaste,
     onFocus,
     onBlur,
@@ -32,6 +33,9 @@ function SingleOtpInput({
         if (inputEl && focus) {
             inputEl.focus();
             inputEl.select();
+            if (paste) {
+                inputEl.selectionStart = inputEl.selectionEnd;
+            }
         }
     }, [focus, input]);
 
@@ -69,6 +73,7 @@ function SingleOtpInput({
 
 function OtpInput({ onChange, onClick, value, field, inlineError }) {
     const [activeInput, setActiveInput] = useState(0);
+    const [pasted, setPasted] = useState(false);
     const [rawValue, setRawValue] = useState(value.toString().split(''));
 
     // Helper to return OTP from input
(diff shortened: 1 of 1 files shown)


src/elements/fields/PinInput.jsx
```diff
@@ -186,9 +192,11 @@ function OtpInput({ onChange, onClick, value, field, inlineError }) {
                     onChange={handleOnChange}
                     onKeyDown={handleOnKeyDown}
                     onInput={handleOnInput}
+                    paste={pasted}
                     onPaste={handleOnPaste}
                     onFocus={(e) => {
```
bo-dun-1, line 197: what if the user starts typing again after pasting but before re-focusing and setting paste back to false?

src/elements/fields/PinInput.jsx
```diff
@@ -33,6 +34,9 @@ function SingleOtpInput({
             inputEl.focus();
             inputEl.select();
         }
+        if (paste) {
```
bo-dun-1, line 37: should this be wrapped in the previous if statement

bo-dun-1: also can this behavior be replicated by just not calling inputEl.select()?

bo-dun-1 (changes requested):

src/elements/fields/PinInput.jsx
ckrueger6: That's fine, it'll go back to autohighlighting so they can keep typing (and overwriting if need be)

src/elements/fields/PinInput.jsx
ckrueger6: Putting it inside the if makes sense to me, but select and focus both seem to do the same thing, and not calling either of them without doing the selection space thing I'm doing makes it very unclear where your cursor is after pasting

src/elements/fields/PinInput.jsx
ckrueger6: correction: it will do that now, your comment made me go looking for the edge case and it's handled now, thanks!

bo-dun-1 (approved):

PR #205, opened 2021-09-23
Fixed required status on repeat_trigger elements

JakeKo:


src/Form.jsx
```diff
@@ -114,15 +115,23 @@ function Form({
...
+                      values: fieldValues
+                  })
+                : null,
+        [rawActiveStep, repeatChanged]
+    );
+
     // Create the fully-hydrated activeStep by injecting repeated rows
     // and pre-calculating styles.
     // Note: Other hydration transformations can also be included here
     const activeStep = useMemo(() => {
```
bo-dun-1, line 133: should you include repeatedRowCount as a dependency in this useMemo? Are there any race conditions introduced between these two usememos?

src/Form.jsx
```diff
@@ -1413,6 +1433,7 @@ function Form({
                             );
                             controlElement = (
                                 <MaskedTextField
+                                    {...commonProps}
                                     {...fieldProps}
```
bo-dun-1, line 1437: can we change this to textFieldProps and make commonProps -> fieldProps?

bo-dun-1 (approved):

PR #213, opened 2021-09-30
increase specificity of css

bo-dun-1:


src/elements/basic/__test__/__snapshots__/ButtonElement.spec.js.snap
```diff
@@ -2,7 +2,7 @@
 
 exports[`ButtonElement renders a button with custom styles 1`] = `
 <button
-  className="css-100jsw4-ButtonElement btn btn-primary"
+  className="css-3azwfl-ButtonElement btn btn-primary"
```
JakeKo, line 5: Where are these className changes coming from?

JakeKo (approved):

src/elements/basic/__test__/__snapshots__/ButtonElement.spec.js.snap
bo-dun-1: autogenerated from emotion

PR #215, opened 2021-10-05
Implement auto-scroll functionality

harishgadiya:


src/form/Form.jsx
```diff
@@ -152,6 +153,19 @@ function Form({
...
+    const scrollToRef = (ref) => window.scrollTo({
+        top: 0,
+        left: ref?.current?.offsetTop,
+        behavior: 'smooth'
+    });
+    useEffect(() => {
+        const nextKeyName = activeStep?.next_conditions[0]?.next_step_key;
+        const currentKey = activeStep?.key;
+        if (shouldScrollToTop && (currentKey !== nextKeyName)) {
+            scrollToRef(formRef);
```
JakeKo, line 165: Do you need to reset `scrollToTop` at all here?

src/form/GooglePlaces.jsx
```diff
@@ -47,6 +48,7 @@ export default function GooglePlaces({
 
             // Check if address is valid
             if (address) {
+                setShouldScrollToTop(false);
```
JakeKo, line 51: Instead of setting `scrollToTop` in this embedded component, I think I'd prefer that you move this check to `fieldOnChange` and check the `trigger` property to determine if Google Maps caused the field change. That way we can keep the `scrollToTop` checks in roughly the same place

JakeKo (changes requested):
There is an issue with auto-scrolling on some steps (discussing offline)

JakeKo (approved):

PR #244, opened 2021-11-08
Adding useHotKeys module for better keyPress event handling

avnkailash:


avnkailash:
Do not merge it. I want to test the change locally and then merge it. Code-wise changes are all done and complete!

src/elements/fields/TextField.jsx
```diff
@@ -89,6 +90,11 @@ function TextField({
 }) {
   const servar = element.servar;
   const inputType = fieldProps.as === 'textarea' ? 'textarea' : 'input';
+
+  useHotkeys('enter', (e) =>
```
bo-dun-1, line 94: when is this event triggered? before we would specify the element on which we're listening to keydown but now that it's in a use effect, what happens?

avnkailash: useHotKeys hook will listen for all key interactions on the element. When the text field is in focus, the hook will capture the key events and call the callback function(2nd argument). This technically replaces the onKeyDown event handler.

src/elements/fields/PinInputField.jsx
```diff
@@ -41,6 +34,42 @@ function SingleOtpInput({
...
 
   applyStyles.applyBorders('field', focus ? 'selected_' : '', false);
+
+  // Handle cases of backspace, delete, left arrow, right arrow, space
+  useHotkeys(
+    'enter, backspace, delete, left, right, space',
+    (e, handler) => {
+      switch (handler.key) {
+        case 'enter':
+          console.log('You pressed enter!');
```
bo-dun-1, line 44: this doens't look like final code

src/elements/fields/TextField.jsx
```diff
@@ -89,6 +90,15 @@ function TextField({
 }) {
   const servar = element.servar;
   const inputType = fieldProps.as === 'textarea' ? 'textarea' : 'input';
+
+  useHotkeys(
+    'enter',
+    (e) => (inputType === 'textarea' ? e.stopPropagation() : null),
```
bo-dun-1, line 96: can this instead be inputType === 'textarea' && e.stopPropagation()

bo-dun-1: also have you tested that enter successfully creates newlines for textareas on forms and won't attemp to submit that step? but will submit for text fields?

src/form/Form.jsx
```diff
@@ -1012,6 +1013,16 @@ function Form({
     } else handleRedirect({ metadata });
   };
 
+  useHotkeys('enter', (e) => {
```
bo-dun-1, line 1016: have u verified the submit functionality still works as expected?

bo-dun-1 (changes requested):

src/elements/fields/PinInputField.jsx
avnkailash: Fixed it. Was testing the enter key event and missed changing the line.

src/elements/fields/TextField.jsx
avnkailash: Done. Changed the logic to use && operator. Yes I have tested the key event and validated that it behaves as expected in the text area

src/form/Form.jsx
avnkailash: Yes, I have tested. I had to add an option to make it behave correctly and consistently.

src/elements/fields/PinInputField.jsx
avnkailash: As per my testing of the existing PinInput field, I am not doing any specific action except for preventing the default behavior. Should we change that behavior?

avnkailash:
As per my testing of the existing PinInput field, I am not doing any specific action except for preventing the default behavior. Should we change that behavior?

package.json
```diff
@@ -68,11 +68,13 @@
     "@emotion/styled": "^11.0.0",
     "@fingerprintjs/fingerprintjs": "^3.1.0",
     "google-libphonenumber": "^3.2.22",
+    "imask": "^6.2.2",
```
bo-dun-1, line 71: why are we adding imask here when we have react-imask?

bo-dun-1:
> As per my testing of the existing PinInput field, I am not doing any specific action except for preventing the default behavior. Should we change that behavior?

You should test out the existing key listeners, e.g. enter to submit steps with only 1 field

bo-dun-1 (changes requested):

src/elements/fields/PinInputField.jsx
```diff
@@ -60,7 +92,7 @@ function SingleOtpInput({
         ref={input}
         value={value || ''}
         onChange={onChange}
-        onKeyDown={onKeyDown}
+        // onKeyDown={onKeyDown}
```
bo-dun-1, line 95: don't leeave commented out code in production code

src/elements/fields/TextField.jsx
```diff
@@ -89,6 +90,15 @@ function TextField({
 }) {
   const servar = element.servar;
   const inputType = fieldProps.as === 'textarea' ? 'textarea' : 'input';
+
+  useHotkeys(
+    'enter',
+    (e) => (inputType === 'textarea' && e.stopPropagation()),
```
bo-dun-1, line 96: why not just enable this only on inputs and remove this if check? also this is currently buggy, enter will submit text areas even though it shouldn't. did you test this code path?

bo-dun-1 (changes requested):
this pr needs to be rebased

package.json
avnkailash: This has been removed.

src/elements/fields/PinInputField.jsx
avnkailash: done

src/elements/fields/TextField.jsx
avnkailash: Modified this logic to capture only the text area. `enter` events and stop the propagation.

PR #245, opened 2021-11-09
allow set steps and progress

bo-dun-1:


src/form/Form.jsx
```diff
@@ -936,6 +944,33 @@ function Form({
...
+              activeStep,
+              metadata.elementIDs[0],
+              metadata.elementType
+            ),
+            triggerType: metadata.elementType,
+            triggerAction: metadata.trigger
+          });
+          if (stepChanged) return;
+        }
         handleRedirect({ metadata });
```
ckrueger6, line 974: just want to confirm we still want this handleRedirect to run after the onskip if onskip is true

bo-dun-1: yep! the form will navigate to the next step either way, we additionally optionally run the onSkip callback function if it's passed in

ckrueger6 (approved):

PR #283, opened 2022-01-04
Enforce button widths, unique key and id

drewj:


src/elements/fields/ButtonGroupField.jsx
```diff
@@ -23,19 +23,23 @@ function ButtonGroupField({
...
         {servar.metadata.options.map((opt, index) => {
           const imageUrl = servar.metadata.option_images[index];
+          const fieldStyle = applyStyles.getTarget('field');
+          // enforce width
+          fieldStyle.minWidth = fieldStyle.width;
+          fieldStyle.maxWidth = fieldStyle.width;
           return (
             <div
-              id={servar.key}
+              id={`${servar.key}-option-${index + 1}`}
```
bo-dun-1, line 32: should the id unique string be consistent with the key? also is it necessary to use index+1 here but index below and include opt as part of the key

i would also recommend testing the setErrors api https://docs.feathery.io/react/onsubmit#context-properties
to see if setting errors on button groups worked before and/or still works

we look up elements by their html id via servar.key https://github.com/feathery-org/feathery-react/blob/master/src/utils/formHelperFunctions.js#L303 to set default html5 errors but this may have not even been working for button groups before since these are divs, not inputs

src/elements/fields/ButtonGroupField.jsx
```diff
@@ -23,19 +23,23 @@ function ButtonGroupField({
       >
         {servar.metadata.options.map((opt, index) => {
           const imageUrl = servar.metadata.option_images[index];
+          const fieldStyle = applyStyles.getTarget('field');
```
bo-dun-1, line 26: since we allow the user to apply different styles at different breakpoints from the dashboard, applyStyles is middleware we use for translating styles stored in the json dict to css styles, handling inheritance, breakpoints, and mobile compatibility. The button group styles are set at https://github.com/feathery-org/feathery-react/blob/master/src/elements/fields/index.jsx#L104, which is where minWidth and maxWidth should be set too (otherwise won't account for breakpoints)

Happy to give more context here if helpful!

bo-dun-1 (changes requested):

bo-dun-1 (approved):

PR #286, opened 2022-01-06
Drewj/initial values

drewj:


bo-dun-1 (approved):
lgtm! small nit would be to add a comment to applyDefaultFieldValues explaining why we need to do it that way

PR #290, opened 2022-01-12
Drewj/custom checkbox

drewj:


src/elements/fields/index.jsx
```diff
@@ -118,6 +118,7 @@ function applyFieldStyles(field, styles) {
       break;
     case 'checkbox':
       styles.applyMargin('fc');
+
```
bo-dun-1, line 121: nit: extra newline

src/elements/fields/CheckboxField.jsx
```diff
@@ -1,6 +1,155 @@
...
+      return {
+        boxShadow: `inset ${width}${widthUnit} ${height}${heightUnit} #${color}`,
+        clipPath: checkmarkClipPath
+      };
+    }
+  );
+};
+
+export function applyCheckboxStyles(element, applyStyles) {
+  applyStyles.addTargets(
```
bo-dun-1, line 45: adding targets and applying styles should ideally happen here: https://github.com/feathery-org/feathery-react/blob/master/src/elements/fields/index.jsx#L30

src/elements/fields/CheckboxGroupField.jsx
```diff
@@ -1,6 +1,15 @@
-import React from 'react';
+import React, { useMemo } from 'react';
 import ReactForm from 'react-bootstrap/Form';
 import { bootstrapStyles } from '../styles';
+import { applyCheckboxStyles, composeCheckboxStyle } from './CheckboxField';
+
+const applyCheckboxGroupStyles = (element, applyStyles) => {
+  applyStyles.addTargets(['checkboxGroup']);
```
bo-dun-1, line 7: nit: should this be called checkbox rather than checkbox group since it's being applied individually to checkboxes

bo-dun-1: also this style logic should also ideally be in https://github.com/feathery-org/feathery-react/blob/master/src/elements/fields/index.jsx#L30

bo-dun-1 (changes requested):

src/elements/fields/CheckboxGroupField.jsx
drewj: It needs a target other than checkbox. Checkbox refers to the literal input HTML element here. Where as checkboxGroup is the checkbox + label grouping. The checkbox input needs to have a separate width and height than the containing div that holds both the checkbox and the label. Does that make sense?

bo-dun-1: cool make ssense

src/elements/fields/CheckboxField.jsx
```diff
@@ -1,6 +1,156 @@
...
+const checkmarkClipPath =
+  'polygon(40% 85%, 89% 23%, 76% 12%, 39% 59%, 23% 44%, 12% 55%)';
+
+const MIN_CHECKBOX_PX = 13;
+
+const scaleCheckboxSize = (fontSize) => {
+  return Math.max(Math.floor(fontSize * 0.5), MIN_CHECKBOX_PX);
+};
+
+const applyHeightAndWidthByFontSize = (applyStyles, target) => {
```
bo-dun-1, line 14: this seems generic enough that we could maybe move this into the ApplyStyles class with the other style helper functions? e.g. radio button groups will probably need to use the same logic

bo-dun-1 (approved):

src/elements/fields/CheckboxField.jsx
drewj: Yes. Let's move it when the time comes.

PR #325, opened 2022-03-01
Elastic Grid

drewj:


src/utils/hydration.js
```diff
@@ -171,12 +198,23 @@ function calculateStepCSS(step) {
...
     gridTemplateRows: desktop.rows.join(' '),
     gridTemplateColumns: desktop.columns.join(' ')
   };
 
-  stepCSS['@media (max-width: 478px)'] = {
-    width: mobile.gridWidth,
+  // 478
+  stepCSS[`@media (max-width: 478px)`] = {
+    width: 'unset',
+    minWidth: 'unset',
```
bo-dun-1, line 213: is unset necessary here? why?

src/form/Form.jsx
```diff
@@ -532,7 +532,10 @@ function Form({
           stepChanged = changeStep(stepKey, newKey, steps, history);
         },
         firstStepLoaded: first,
-        integrationData
+        integrationData,
+        stepStyle: {
```
bo-dun-1, line 536: i think styles: { step: { backgroundColor: ... }  }  would be a little more encompassing, and you can probably include this in commonCallbackProps so it's available to every callback function. also this will need to be added to the sdk documentation too

bo-dun-1 (changes requested):

src/utils/hydration.js
```diff
@@ -142,19 +142,46 @@ function injectRepeatedRows({ step, repeatedRowCount }) {
...
+    gridWidth: formatDimensionValue(gridWidth, 'col'),
+    gridHeight: formatDimensionValue(gridHeight, 'row'),
+    columns: gridColumns.map((v) => formatTrackValue(v, 'col')),
+    rows: gridRows.map((v) => formatTrackValue(v, 'row'))
+  };
 
-  const hasRelativeColumn = gridColumns.some(
-    (column) => column.slice(-2) === 'fr'
-  );
+  if (gridHeight === 'fill') {
```
bo-dun-1, line 176: const height = gridHeight === 'fill' ? '100%' : gridHeight;
dimensions.minHeight = height;
dimensions.maxHeight = height;

bo-dun-1 (changes requested):

src/form/Form.jsx
```diff
@@ -441,20 +441,31 @@ function Form({
...
+        style: {
+          backgroundColor: newStep?.default_background_color
+        }
+      },
+      userId: initInfo().userKey,
+      stepName: newStep?.key ?? ''
+    };
   };
+
+  const commonCallbackProps = getCommonCallbackProps();
```
bo-dun-1, line 467: can we just have every place call `getCommonCallbackProps` to keep the interface the same

src/form/Form.jsx
```diff
@@ -441,20 +441,31 @@ function Form({
...
-      updateFieldValues(values);
-      client.submitCustom(values);
-    },
-    setProgress: (val) => setUserProgress(val),
-    setStep: (stepKey) => {
-      changeStep(stepKey, activeStep.key, steps, history);
-    },
-    userId: initInfo().userKey,
-    stepName: activeStep?.key ?? ''
+  const getCommonCallbackProps = (newStep) => {
```
bo-dun-1, line 444: u can do newStep = activeStep
as the function parameter to specify a default

src/utils/hydration.js
```diff
@@ -142,19 +142,85 @@ function injectRepeatedRows({ step, repeatedRowCount }) {
...
+const formatTrackValue = (value, type = 'col', isEmpty) => {
+  if (isEmpty && value === 'auto') value = '0'; // Default empty value
+  if (type === 'col') return `minmax(0, ${value})`;
+  return value;
+};
+
+const formatDimensionValue = (value, type = 'col') => {
+  switch (value) {
+    case 'fill':
+      return type === 'col' ? '100%' : 'fit-content';
```
bo-dun-1, line 154: can you include a comment here about why fit-content for rows too? and why fit-content for the auto case too?

src/utils/hydration.js
```diff
@@ -142,19 +142,85 @@ function injectRepeatedRows({ step, repeatedRowCount }) {
...
+const getTrackElementsMap = (step) => {
+  const {
+    servar_fields: fields,
+    progress_bars: progressBars,
+    images,
+    texts,
+    buttons
+  } = step;
+
+  const allElements = [
```
bo-dun-1, line 171: i think u should decompose this logic with this logic here https://github.com/feathery-org/feathery-react/blob/master/src/form/Form.jsx#L188

src/utils/hydration.js
```diff
@@ -142,19 +142,85 @@ function injectRepeatedRows({ step, repeatedRowCount }) {
...
+    row[rowIndex].push(e);
+
+    if (!(columnIndex in column)) column[columnIndex] = [];
+    column[columnIndex].push(e);
+  });
+
+  return { row, column };
+};
+
+const mobileWidthHelper = (width) => {
```
bo-dun-1, line 195: why is width defaulting to 100% on mobile?

src/utils/hydration.js
```diff
@@ -142,19 +142,85 @@ function injectRepeatedRows({ step, repeatedRowCount }) {
...
+    ...buttons
+  ];
+
+  const row = {};
+  const column = {};
+
+  allElements.forEach((e) => {
+    const { row_index: rowIndex, column_index: columnIndex } = e;
+
+    if (!(rowIndex in row)) row[rowIndex] = [];
```
bo-dun-1, line 185: could you just store true or false instead of the entire array? then you can just do one line to set the index to true instead of the if check

src/utils/hydration.js
```diff
@@ -142,19 +142,85 @@ function injectRepeatedRows({ step, repeatedRowCount }) {
...
-    rows: gridRows.map((row) => `minmax(${row},min-content)`)
+  const dimensions = {
+    gridWidth: formatDimensionValue(gridWidth, 'col'),
+    gridHeight: formatDimensionValue(gridHeight, 'row'),
+    columns: gridColumns.map((v, i) => {
+      const isEmpty = !elementsMap.column[i] && gridWidth === 'auto';
+      return formatTrackValue(v, 'col', isEmpty);
+    }),
+    rows: gridRows.map((v, i) => {
+      const isEmpty = !elementsMap.row[i] && gridWidth === 'auto';
```
bo-dun-1, line 215: technically `emptyAuto` not just `isEmpty`

src/utils/hydration.js
```diff
@@ -142,19 +142,85 @@ function injectRepeatedRows({ step, repeatedRowCount }) {
   };
 }
 
+const formatTrackValue = (value, type = 'col', isEmpty) => {
+  if (isEmpty && value === 'auto') value = '0'; // Default empty value
+  if (type === 'col') return `minmax(0, ${value})`;
```
bo-dun-1, line 147: can you add a comment here about why we have the minmax?

src/utils/hydration.js
```diff
@@ -142,19 +142,85 @@ function injectRepeatedRows({ step, repeatedRowCount }) {
...
+      const isEmpty = !elementsMap.column[i] && gridWidth === 'auto';
+      return formatTrackValue(v, 'col', isEmpty);
+    }),
+    rows: gridRows.map((v, i) => {
+      const isEmpty = !elementsMap.row[i] && gridWidth === 'auto';
+      return formatTrackValue(v, 'row', isEmpty);
+    })
   };
+
+  dimensions.minHeight = dimensions.maxHeight =
```
bo-dun-1, line 220: can u add a comemnt about why this is needed

bo-dun-1 (changes requested):

src/utils/hydration.js
drewj: That would change the nature of this function, but sure, I guess it isn't needed as such at the moment.

bo-dun-1 (approved):

src/utils/hydration.js
```diff
@@ -1,7 +1,22 @@
...
+const isFill = (v) => {
+  return v === FILL;
+};
+
+const isAuto = (v) => {
+  return v === AUTO;
+};
+
+const isFixed = (v) => {
+  return !isNaN(parseInt(v));
```
bo-dun-1, line 17: we already have a isNum helper function for this

src/utils/hydration.js
```diff
@@ -142,19 +157,75 @@ function injectRepeatedRows({ step, repeatedRowCount }) {
   };
 }
 
+const formatTrackValue = (parentDimension, value, type = 'col', isEmpty) => {
+  if (isAuto(value)) {
+    if (!isAuto(parentDimension)) {
+      value = '1fr';
+    } else {
+      value = isEmpty ? '50px' : 'min-content';
```
bo-dun-1, line 165: weren't empty auto columns going to 0?

src/utils/hydration.js
```diff
@@ -142,19 +157,75 @@ function injectRepeatedRows({ step, repeatedRowCount }) {
   };
 }
 
+const formatTrackValue = (parentDimension, value, type = 'col', isEmpty) => {
+  if (isAuto(value)) {
+    if (!isAuto(parentDimension)) {
```
bo-dun-1, line 162: can we add quick comments explaining what each of these if statement branches are

bo-dun-1 (changes requested):

bo-dun-1 (approved):

PR #338, opened 2022-03-30
on change spinner

bo-dun-1:


src/utils/primitives.js
```diff
@@ -8,4 +8,8 @@ function isNum(candidate) {
   return !isNaN(parseInt(candidate));
 }
 
-export { encodeGetParams, isNum };
+function emptyObject(obj) {
```
obmelvin, line 11: personally I'd call this `isObjectEmpty`. consistent with `isNum` above and functions like `isNaN`

obmelvin (approved):

PR #341, opened 2022-04-01
fix signature width and onChange

obmelvin:
1) field width should now respect the width in style > layout > width 
2) the onChange function passed to <Form /> now properly fires after the end of stroke in signature field
3) when trying to evaluate the signature field value during onChange, the value now contains a file with the signature image. previously was just ''

### DIFF
--- a/src/elements/fields/SignatureField.jsx
+++ b/src/elements/fields/SignatureField.jsx
@@ -7,6 +7,7 @@ function SignatureField({
   applyStyles,
   signatureRef = {},
   elementProps = {},
+  onEnd,
   children
 }) {
   const servar = element.servar;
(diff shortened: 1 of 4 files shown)


src/elements/fields/SignatureField.jsx
```diff
@@ -24,13 +25,16 @@ function SignatureField({
         penColor='black'
         canvasProps={{
           id: servar.key,
-          width: element.styles.width,
           height: element.styles.height,
-          style: applyStyles.getTarget('field', true)
+          style: {
```
bo-dun-1, line 29: why is width being set here and applyWidth is being called on styles?

bo-dun-1 (changes requested):

src/elements/fields/SignatureField.jsx
obmelvin: img 1 - shows the width that gets set with `applyWidth` 
img 2 - shows how the canvas stretches outside the parent when there is no width set directly on the canvas
img 3 - shows how setting the width to 100% makes it occupy the width of its parent. Sorry - I was wrong about setting the width to `${element.styles.width}${element.styles.width_unit}`

image
image
image

obmelvin: oh, sorry I published a new form between images 1 and 3 changing the width to 80%. that's why the canvas doesn't take up the whole column in the 3rd image

this is how it looks at 100%
image
image

bo-dun-1 (approved):

PR #345, opened 2022-04-05
add onView and onViewFields to Form

obmelvin:
Do you think the way we call `fieldOnView` is appropriate? it fires for each field - and it seems each `<VisiblitySensor />` also fires once at the start to report the initial status. should we batch this?

### DIFF
--- a/src/form/Form.jsx
+++ b/src/form/Form.jsx
@@ -76,6 +76,8 @@ function Form({
   onSkip = null,
   onError = null,
   onCustomAction = null,
+  onView = null,
+  onViewElements = [],
   initialValues = {},
   initialStepId = '',
   usePreviousUserData = null,
@@ -1165,6 +1167,19 @@ function Form({
     } else handleRedirect({ metadata });
   };
 
+  const elementOnView =
+    typeof onView === 'function'
+      ? (elementId, isVisible) => {
+          callbackRef.current.addCallback(
+            onView({
+              ...getCommonCallbackProps(),
+              visibilityStatus: { elementId, isVisible }
+            }),
+            loaders
+          );
+        }
+      : undefined;
+
   let fieldCounter = 0;
   return (
     <>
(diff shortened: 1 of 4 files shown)


src/form/Form.jsx
```diff
@@ -75,6 +75,8 @@ function Form({
   onSkip = null,
   onError = null,
   onCustomAction = null,
+  onView = null,
+  onViewFields = [],
```
bo-dun-1, line 79: maybe onViewElements instead? visibility callback could apply to non-field elements too

bo-dun-1 (changes requested):

src/form/Form.jsx
```diff
@@ -141,9 +143,15 @@ function Form({
       ),
     [steps]
   );
+  const onViewFieldsDict = useMemo(
```
bo-dun-1, line 146: kinda seems like you want to use a set rather than an object here

bo-dun-1 (changes requested):

src/form/Form.jsx
```diff
@@ -1258,6 +1277,12 @@ function Form({
             else return 0;
           })
           .map(([el, type]) => {
+            const fieldId = el?.servar?.key ? el?.servar?.key : el.id;
```
bo-dun-1, line 1280: el.servar?.key i think since el will always be an object

bo-dun-1: also const fieldId = el.servar?.key ?? el.id

src/form/Form.jsx
```diff
@@ -1212,6 +1217,20 @@ function Form({
     } else handleRedirect({ metadata });
   };
 
+  const elementOnView = onView
+    ? (elementId, isVisible) => {
+        if (typeof onView === 'function') {
```
bo-dun-1, line 1222: i think this check can be the top-level ternary so you don't need an additional check here

bo-dun-1 (changes requested):

bo-dun-1 (approved):

PR #347, opened 2022-04-05
global field values

bo-dun-1:


src/utils/client.js
```diff
@@ -175,26 +200,59 @@ export default class Client {
...
+    } = initInfo();
     if (this.formKey in sessions)
       return Promise.resolve(sessions[this.formKey]);
 
-    const params = encodeGetParams({
-      form_key: this.formKey,
-      ...(userKey ? { fuser_key: userKey } : {}),
-      ...(authId ? { auth_id: authId } : {})
-    });
+    initState.fieldValuesInitialized = true;
```
obmelvin, line 220: should this function still proceed if `fieldValuesInitialized` is already true?

bo-dun-1: yes it should because there is other session data to initialize

obmelvin: for some reason I wasn't appreciating that `fieldValuesInitialized: noData` is a rename - I think I processed it as a default value. I think I'm now following how you are using this 👍

obmelvin (approved):

PR #348, opened 2022-04-05
change signature value to base64 from File

obmelvin:
* changes the value of signature fields from JS File obj to be a base64 value
* allow `Form`'s `initialValues` and `setValues` to take base64 value for signature
* bug fix - if a signature value exists, the signature field is properly populated with the previous signature


src/utils/client.js
```diff
@@ -215,14 +220,19 @@ export default class Client {
...
+  submitCustom(customKeyValues, signatureRef) {
     const { userKey } = initInfo();
     const url = `${API_URL}panel/custom/submit/v2/`;
 
     const jsonKeyVals = {};
     const formData = new FormData();
     Object.entries(customKeyValues).forEach(([key, val]) => {
-      if (val instanceof Blob) {
+      if (isBase64PNG(val)) {
+        if (signatureRef[key]) signatureRef[key].fromDataURL(val);
```
obmelvin, line 231: I think this line can be folded into the original if block, and the conditional can become `isBase64PNG(val) || val instanceof Blob`?

src/utils/client.js
```diff
@@ -215,14 +220,18 @@ export default class Client {
...
```
bo-dun-1, line 231: there is also a submitCustom in commonCallbackProps that needs to be updated i believe. i think what may be cleaner is to put the signature update logic in updateFieldValues instead, that way it doesn't need to live in both updateSessionValues and submitCustom

bo-dun-1 (changes requested):

src/form/Form.jsx
```diff
@@ -210,6 +211,11 @@ function Form({
   useEffect(() => {
     if (!activeStep) return;
 
+    Object.entries(initialValues).forEach(([key, val]) => {
```
bo-dun-1, line 214: i think this logic should be flipped. we run through activeStep and look for any signature fields. for those fields, see if there exists a value for it. if so, call fromDataURL and initialize it

src/utils/client.js
```diff
@@ -273,14 +278,19 @@ export default class Client {
...
+  submitCustom(customKeyValues, signatureRef) {
     const { userKey } = initInfo();
     const url = `${API_URL}panel/custom/submit/v2/`;
 
     const jsonKeyVals = {};
     const formData = new FormData();
     Object.entries(customKeyValues).forEach(([key, val]) => {
-      if (val instanceof Blob) {
+      if (isBase64PNG(val)) {
+        if (signatureRef && signatureRef[key])
```
bo-dun-1, line 289: i don't think any other fromDataURL logic is necessary other than the one in the hydration function to keep it clean. we can establish a constraint that the value has to be set before the step with the signature field is loaded, toherwise it won't update in real time

bo-dun-1 (changes requested):

src/utils/client.js
obmelvin: this logic is (at least currently) needed to set the canvas & file when the user submits a base64 via `setValues`

src/elements/fields/SignatureField.jsx
```diff
@@ -34,7 +36,12 @@ function SignatureField({
         ref={(ref) => {
           signatureRef[servar.key] = ref;
         }}
-        onEnd={onEnd}
+        onEnd={() => {
```
bo-dun-1, line 39: this logic should live in the onEnd function passed to this component since this is form specific info and the dashboard doesn't need it

src/elements/fields/SignatureField.jsx
```diff
@@ -34,7 +36,12 @@ function SignatureField({
...
+          const base64Img = signatureRef[servar.key].toDataURL('image/png');
+          const newFile = dataURLToFile(base64Img, `${servar.key}.png`);
+          fieldValues[servar.key] = newFile;
```
bo-dun-1, line 42: shouldn't this be wrapped in a Promise.resolve()

src/utils/__test__/formHelperFunctions.spec.js
```diff
@@ -67,6 +67,7 @@ describe('formHelperFunctions', () => {
       };
 
       // Act
+      // even before I changed formatStepFields, it did not take a file for the 3rd param
```
bo-dun-1, line 70: seems like a bug then, feel free to refactor the tests

bo-dun-1 (changes requested):

src/utils/init.js
```diff
@@ -152,7 +153,9 @@ function setValues(userVals) {
   Object.entries(userVals).forEach(([key, value]) => {
     if (Array.isArray(value))
       result[key] = value.map((entry) => _parseUserVal(entry));
-    else result[key] = _parseUserVal(value);
+    else if (isBase64PNG(value)) {
+      result[key] = _parseUserVal(dataURLToFile(value, `${key}.png`));
```
bo-dun-1, line 157: this logic should actually go in _parseUserVal, the array here is to handle repeated fields not an actual field type

src/utils/image.js
```diff
@@ -80,3 +80,31 @@ export function useThumbnailData(files) {
...
+  new Promise((resolve, reject) => {
+    const reader = new FileReader();
+    reader.readAsDataURL(file);
+    reader.onload = () => resolve(reader.result);
+    reader.onerror = (error) => reject(error);
+  });
+
+export const isBase64PNG = (string) => {
+  // eslint-disable-next-line no-useless-escape
+  const base64regex = /(data:image\/png;base64,)([0-9a-zA-Z+\/]{4})*(([0-9a-zA-Z+\/]{2}==)|([0-9a-zA-Z+\/]{3}=))?$/gm;
```
bo-dun-1, line 108: can we make this regex pattern a global const in the file

bo-dun-1 (changes requested):

src/utils/formHelperFunctions.js
```diff
@@ -248,7 +244,7 @@ function getFieldError(value, servar, signatureRef) {
       noVal = !value && servar.metadata?.must_check;
       break;
     case 'signature':
-      noVal = signatureRef[servar.key].isEmpty();
+      noVal = value.isEmpty();
```
bo-dun-1, line 247: isEmpty used to be called on the ref element directly, i'm not sure it can be called on value like this. probably also worth testing

src/utils/formHelperFunctions.js
```diff
@@ -52,14 +48,14 @@ const formatStepFields = (step, fieldValues, signatureRef) => {
...
     let value;
-    if (servar.type === 'signature') {
+    // Only use base64 for signature if these values will be presented to the user
+    if (servar.type === 'signature' && forUser) {
       value =
-        signatureRef && signatureRef[servar.key]
-          ? dataURLToFile(
-              signatureRef[servar.key].toDataURL('image/png'),
-              `${servar.key}.png`
+        fieldValues[servar.key] !== ''
```
bo-dun-1, line 54: this ternary may be cleaner if you first store
const val = fieldValues[servar.key];

not confident about that tho, feel free to disregard

bo-dun-1: also seems like the ternary can go within the Promise.resolve() rather than outside it

src/utils/formHelperFunctions.js
```diff
@@ -52,14 +48,14 @@ const formatStepFields = (step, fieldValues, signatureRef) => {
 
     const servar = field.servar;
     let value;
-    if (servar.type === 'signature') {
+    // Only use base64 for signature if these values will be presented to the user
+    if (servar.type === 'signature' && forUser) {
```
bo-dun-1, line 52: super nit: forUser maybe should go first so type doesn't even need to get evaluated for code paths that aren't for the user

src/utils/client.js
```diff
@@ -273,18 +288,25 @@ export default class Client {
...
 
     const jsonKeyVals = {};
     const formData = new FormData();
-    Object.entries(customKeyValues).forEach(([key, val]) => {
-      if (val instanceof Blob) {
-        formData.append('files', val);
+    const promiseResults = await Promise.all(
+      Object.entries(customKeyValues).map(([key, val]) => Promise.resolve(val))
+    );
+    Object.entries(customKeyValues).forEach(([key, val], index) => {
```
bo-dun-1, line 300: maybe 

const promiseResults = await Promise.all(Object.entries(customKeyValues).map(([key, val]) => [key, Promise.resolve(val)]))
promiseResults.forEach(([key, val]) => {
  if (value instanceof Blob)...
});

src/utils/formHelperFunctions.js
```diff
@@ -24,23 +25,18 @@ const validators = {
...
+ * @param {*} fieldValues
+ * @param {boolean} forUser indicate whether the result of this function is
+ * meant for the user, or Feathery's BE. Presently the only difference is
+ * whether signature field values are base64 or a JS File obj
+ * @returns Formatted fields for the step
+ */
+const formatStepFields = (step, fieldValues, forUser) => {
   const formattedFields = {};
-  step.servar_fields.forEach((field) => {
+  step.servar_fields.forEach(async (field) => {
```
bo-dun-1, line 39: does this function need to be async?

bo-dun-1 (changes requested):

src/utils/client.js
```diff
@@ -177,7 +182,17 @@ export default class Client {
     steps.forEach((step) => {
       step.servar_fields.forEach((field) => {
         const val = getDefaultFieldValue(field);
-        values[field.servar.key] = field.servar.repeated ? [val] : val;
+        // This can't be done when the canvas is hydrated with initial value
```
bo-dun-1, line 185: this comment confused me a little because i thought the canvas hydration logic was fairly detached from this logic. the hydration logic hydrates a canvas, while this logic formats initialValue data URLs to be set to fieldValues

src/utils/client.js
```diff
@@ -177,7 +182,17 @@ export default class Client {
...
+        // because that runs after the Form onSubmit callback is initialized,
+        // which expects a File, not the base64 supplied by the user
+        const { key, repeated, type } = field.servar;
+        if (type === 'signature' && additionalValues[key]) {
```
bo-dun-1, line 189: this check has an edge case if the user tries to set the signature value for a field not present in this form. what if we maintain an invariant that all base 64 strings need to be converted to files instead? and we can typecheck on that directly.

this current logic can also break if the user tries to set a malformed value to a signature field and we attempt to convert that

bo-dun-1 (changes requested):

src/utils/client.js
obmelvin: hm ok yeah, on second thought I think this comment is more confusing than helpful. are you ok with me just removing it?

bo-dun-1: 👍

src/elements/fields/SignatureField.jsx
```diff
@@ -34,7 +36,9 @@ function SignatureField({
         ref={(ref) => {
           signatureRef[servar.key] = ref;
         }}
-        onEnd={onEnd}
+        onEnd={() => {
+          onEnd();
```
obmelvin, line 40: should I revert this to onEnd={onEnd} now, or do you prefer it like this to make sure the event isn't passed up?

bo-dun-1: yeah onEnd={onEnd} would probably be best, it shouldn't matter if the onEnd is ignoring the param anyways

src/utils/client.js
```diff
@@ -273,14 +283,22 @@ export default class Client {
...
+  async submitCustom(customKeyValues) {
     const { userKey } = initInfo();
     const url = `${API_URL}panel/custom/submit/v2/`;
 
     const jsonKeyVals = {};
     const formData = new FormData();
-    Object.entries(customKeyValues).forEach(([key, val]) => {
+    const promiseResults = await Promise.all(
+      Object.entries(customKeyValues).map(([key, val]) =>
+        Promise.all([key, Promise.resolve(val)])
```
bo-dun-1, line 294: whoops, i didn't realize you'd need to wrap promises like this. i no longer fele strongly about using this approach or the previous

src/utils/client.js
```diff
@@ -273,14 +283,22 @@ export default class Client {
...
+      )
+    );
+    promiseResults.forEach(([key, val]) => {
       if (val instanceof Blob) {
+        // If you use val from customKeyValues, instead of value from
```
bo-dun-1, line 299: first comma here is unnecessary i think and threw me off a little. can you explain why you can't use val from customKeyValues?

bo-dun-1 (approved):

src/utils/client.js
obmelvin: I removed the first comma. The second sentence explains why it doesn't work - the files aren't resolved before submission. I confirmed that the original code for this function does *not* actually submit files correctly to the BE

src/elements/fields/SignatureField.jsx
obmelvin: My thinking with this was that maybe it's better to discard the param right away, rather than rely on something further up the stack to do so

src/utils/client.js
bo-dun-1: ah cool makes sense

PR #353, opened 2022-04-07
fix signature conditional causing error with initialValues

obmelvin:


src/utils/client.js
```diff
@@ -178,14 +178,12 @@ export default class Client {
...
-        const { key, repeated, type } = field.servar;
-        if (type === 'signature' && val !== '') {
-          if (isBase64PNG(additionalValues[key])) {
-            additionalValues[key] = dataURLToFile(
-              additionalValues[key],
-              `${key}.png`
-            );
-          } else console.error(`${key} expects a base64 PNG`);
+        const { key, repeated } = field.servar;
+        if (isBase64PNG(additionalValues[key])) {
```
bo-dun-1, line 182: can we add a small comment, like 

// All base64 strings need to be wrapped in a file

bo-dun-1 (approved):

PR #365, opened 2022-04-14
add signature reset icon

obmelvin:


src/elements/fields/SignatureField.jsx
```diff
@@ -10,9 +9,15 @@ function SignatureField({
   signatureRef = {},
   elementProps = {},
   onEnd,
+  onClear,
+  isClearInitiallyVisible,
```
bo-dun-1, line 13: defaultClearVisibility?

bo-dun-1: also should it = false in the parameters

src/form/Form.jsx
```diff
@@ -1401,6 +1401,9 @@ function Form({
                         fieldValues[servar.key] = Promise.resolve(newFile);
                         onChange();
                       }}
+                      onClear={() => (fieldValues[servar.key] = '')}
+                      // We need to pass an initial value to account for possibility of an existing signature value
+                      isClearInitiallyVisible={!!fieldValues[servar.key]}
```
bo-dun-1, line 1406: bool(fieldValues[servar.key])? just an idea, not sure which one is cleaner

bo-dun-1 (changes requested):

src/elements/fields/SignatureField.jsx
obmelvin: I'm not sure if default correctly conveys the purpose of this prop

I wrote a comment in <Form /> (https://github.com/feathery-org/feathery-react/pull/365/files#diff-421a55da4cd6f6a06fadba2ad8f542f24b877f82de1dbdfd9271e482d69b91edR1404) about why this prop exists - so that we know whether the initialState should be true/false based on whether or not there's an existing session value or if it's set with `initialValues` 

It seemed more appropriate to put the comment in Form because this prop is required due to the implementation details of Form, not SignatureField

src/elements/fields/SignatureField.jsx
```diff
@@ -1,18 +1,35 @@
...
-import { fieldValues } from '../../utils/init';
+import { dataURLToFile, toBase64 } from '../../utils/image';
+import { CloseIcon } from '../components/icons';
 
 function SignatureField({
   element,
   fieldLabel,
   applyStyles,
-  signatureRef = {},
+  initialSignature,
```
bo-dun-1, line 10: defaultValue? to be consistent with html inputs?

src/form/Form.jsx
```diff
@@ -1385,22 +1368,16 @@ function Form({
...
-                          'image/png'
-                        );
-                        const newFile = dataURLToFile(
-                          base64Img,
-                          `${servar.key}.png`
-                        );
                         fieldValues[servar.key] = Promise.resolve(newFile);
                         onChange();
                       }}
+                      onClear={() => (fieldValues[servar.key] = '')}
```
bo-dun-1, line 1380: should this be null instead of empty string since files are objects?

bo-dun-1 (changes requested):

src/utils/formHelperFunctions.js
```diff
@@ -96,11 +96,11 @@ function getDefaultFieldValue(field) {
     case 'select':
       return null;
     case 'file_upload':
```
bo-dun-1, line 98: could we put 'select' in the cases here too since they all return null

bo-dun-1 (approved):

PR #367, opened 2022-04-15
fix value race condition

bo-dun-1:


src/form/Form.jsx
```diff
@@ -577,63 +576,48 @@ function Form({
...
-                const newKey =
-                  initialStepId ||
-                  (hashKey && hashKey in steps && hashKey) ||
-                  (saveUserLocation && session.current_step_key) ||
-                  getOrigin(steps).key;
-                setFirstStep(newKey);
-                history.replace(
-                  location.pathname + location.search + `#${newKey}`
-                );
+        .fetchSession(fetchPromise)
```
obmelvin, line 579: maybe rename `fetchPromise` to `formPromise`? would match the param name in the `fetchSession` fn and make it more clear which fetch the promise is for

obmelvin (approved):

src/form/Form.jsx
bo-dun-1: good point

PR #369, opened 2022-04-15
add setValues rerender param

obmelvin:


src/form/Form.jsx
```diff
@@ -516,6 +516,10 @@ function Form({
       return errors;
     };
 
+    initState.setValuesCallbacks[formKey] = () => {
```
bo-dun-1, line 519: renderCallbacks?

src/form/Form.jsx
```diff
@@ -516,6 +516,10 @@ function Form({
...
+      setRender((render) => !render);
```
bo-dun-1, line 520: should we autogenerate a uuid unique to this form in a state variable? and case on that rather than the formKey since potentially they might have the same form rendered in multiple places (e.g. our own job app form)

also this could be moved into a useEffect with an empty dependency array since this function doesn't need to be refreshed every time a new step is loaded

in the cleanup function of that useEffect you could remove both the validate and render callbacks since we don't want to attempt to call the functions of an unmounted component

bo-dun-1 (changes requested):

bo-dun-1 (approved):

PR #373, opened 2022-04-18
editable text support

bo-dun-1:


src/elements/components/TextNodes.jsx
```diff
@@ -31,67 +46,79 @@ function TextNodes({
...
-                trigger: 'click',
-                start: attrs.fullSpan ? undefined : attrs.start,
-                end: attrs.fullSpan ? undefined : attrs.end
+            const attrs = op.attributes || {};
+            let onClick = () => {};
+            let cursor = 'text';
+            if (!editable) {
+              cursor = 'inherit';
+              if (attrs.font_link) {
+                onClick = () => window.open(attrs.font_link, '_blank');
```
obmelvin, line 86: this isn't needed on modern browser versions, but still probably a good idea to do `window.open(attrs.font_link, '_blank', 'noopener noreferrer')` https://mathiasbynens.github.io/rel-noopener

obmelvin (approved):

PR #378, opened 2022-04-20
replace prop formKey with formName for clarity

obmelvin:


bo-dun-1 (approved):
maybe we can have a comment at the top of this useEffect explaining that all mount and unmount logic should be put here

PR #387, opened 2022-04-29
add text align to button and button group

obmelvin:


src/elements/basic/ButtonElement.jsx
```diff
@@ -49,6 +49,7 @@ function applyButtonStyles(element, applyStyles) {
       padding: `${a}px ${b}px ${c}px ${d}px`
     })
   );
+  applyStyles.apply('tc', 'text_align', (a) => ({ justifyContent: a }));
```
bo-dun-1, line 52: where do you add the tc target?

bo-dun-1 (changes requested):

src/elements/basic/ButtonElement.jsx
obmelvin: ah, ty, good catch! I forgot to add it for img as well, both are now added

bo-dun-1 (approved):

PR #394, opened 2022-05-10
unify file upload field

obmelvin:


src/utils/image.js
```diff
@@ -33,6 +33,8 @@ export function getThumbnailType(file) {
 export function useFileData(initialFiles, onSetFiles = () => {}) {
   const [files, setFiles] = useState(toList(initialFiles));
   useEffect(() => {
+    // prevent infinite loop of setting a new empty array as the value
+    if (files.length === 0 && initialFiles.length === 0) return;
```
bo-dun-1, line 37: why does this infinite loop? doesn't look like any of these actions should affect initialFiles

src/elements/fields/FileUploadField.jsx
```diff
@@ -1,35 +1,218 @@
...
+import { justRemove } from '../../utils/array';
 
-import ReactForm from 'react-bootstrap/Form';
+import { FiX } from 'react-icons/fi';
+import { IconContext } from 'react-icons';
+import { Image } from 'react-bootstrap';
+import { FileUploadIcon } from '../components/icons';
 
-function FileUploadField({
+function RichFileUploadField({
```
bo-dun-1, line 10: should this still be FileUploadField, not RichFileUploadField

src/utils/client.js
```diff
@@ -178,15 +174,15 @@ export default class Client {
...
+        const { key, repeated, type } = field.servar;
         if (isBase64PNG(additionalValues[key])) {
           // All base64 strings need to be wrapped in a File
           additionalValues[key] = dataURLToFile(
             additionalValues[key],
             `${key}.png`
           );
         }
-        values[key] = repeated ? [val] : val;
+        values[key] = repeated && type !== 'file_upload' ? [val] : val;
```
bo-dun-1, line 185: can we add a comment for why we're ignoring this for file_upload?

src/form/Form.jsx
```diff
@@ -1431,7 +1397,10 @@ function Form({
                         changeValue(files, el, index);
                         onChange({
                           valueRepeatIndex: fieldIndex,
-                          submitData: false
+                          submitData:
+                            el.properties.submit_trigger === 'auto' &&
+                            el.properties.multiple &&
```
bo-dun-1, line 1402: should submitData be true only if multiple is false?

src/form/Form.jsx
```diff
@@ -676,9 +676,14 @@ function Form({
 
       // Add a repeated row if the value went from unset to set
       // And this is the last field in a set of repeated fields
+      const isPreviousValueDefaultArray =
```
bo-dun-1, line 679: were we just not handling multi file fields properly before?

bo-dun-1: basically any array based field

src/form/Form.jsx
```diff
@@ -700,7 +705,11 @@ function Form({
     updateValues[servar.key] =
       index === null
         ? value
-        : justInsert(fieldValues[servar.key], value, index);
+        : justInsert(
+            fieldValues[servar.key],
+            servar.type === 'file_upload' ? value[0] : value,
```
bo-dun-1, line 710: is this basically making the assumption that if index is not null, it's a repeated row so the file field cannot handle multiple files? if so can we add that as a comment

src/form/Form.jsx
bo-dun-1: also maybe checking is a random value is an empty array should be a helper function in utils

src/form/Form.jsx
```diff
@@ -331,10 +331,10 @@ function Form({
     const fieldKeys = [];
     repeatedServarFields.forEach((field) => {
       const { servar } = field;
-      updatedValues[servar.key] = [
-        ...fieldValues[servar.key],
-        getDefaultFieldValue(field)
-      ];
+      updatedValues[servar.key] =
+        servar.type === 'file_upload'
+          ? fieldValues[servar.key]
```
bo-dun-1, line 336: why aren't we adding an extra null or something to the array of files?

src/elements/fields/FileUploadField.jsx
```diff
@@ -1,35 +1,218 @@
...
+            borderRadius: '4px',
+            overflow: 'hidden',
+            display: 'flex',
+            justifyContent: 'center',
+            alignItems: 'center',
+            boxSizing: 'border-box',
+            ...applyStyles.getTarget('field')
+          }}
+        >
+          {thumbnail && (
```
bo-dun-1, line 109: this might be better as a ternary, thumbnail ? <Image /> : <span />

src/elements/fields/FileUploadField.jsx
```diff
@@ -1,35 +1,218 @@
...
+    minHeight: 0,
+    ...applyStyles.getTarget('img')
+  };
+  const usingDefaultIcon = !element.properties.icon;
+  // Default img width is '', which causes the default icon to have no width
+  // and not appear. Set a default width if the user hasn't set one
+  if (usingDefaultIcon && Number.isNaN(Number.parseInt(imgStyles.width))) {
+    imgStyles.width = '32px';
+  }
+  const icon = !usingDefaultIcon ? (
```
bo-dun-1, line 73: nit: can we use usingDefaultIcon directly and flip the ternary

src/elements/fields/FileUploadField.jsx
```diff
@@ -1,35 +1,218 @@
...
   const servar = element.servar;
+  const showLabel = servar.name !== '';
+  const isMultiple = servar.metadata.multiple;
+  const fileInput = useRef();
+
+  const [rawFiles, setRawFiles] = useFileData(initialFiles);
+  const thumbnailData = useThumbnailData(rawFiles);
+  // verify that thumbnail data still has length when there's no thumbnail for a text file
+  const allowMoreFiles = isMultiple || thumbnailData.length === 0;
+  const fileExists = thumbnailData.length > 0 || rawFiles.length > 0;
```
bo-dun-1, line 30: why check both thumbnailData and rawFiles vs just rawFiles?

src/elements/fields/FileUploadField.jsx
```diff
@@ -1,35 +1,218 @@
...
+  const imgStyles = {
+    maxWidth: '100%',
+    maxHeight: '100%',
+    minHeight: 0,
+    ...applyStyles.getTarget('img')
+  };
+  const usingDefaultIcon = !element.properties.icon;
+  // Default img width is '', which causes the default icon to have no width
+  // and not appear. Set a default width if the user hasn't set one
+  if (usingDefaultIcon && Number.isNaN(Number.parseInt(imgStyles.width))) {
```
bo-dun-1, line 70: instead of setting a width just for the default icon, is there a css prop that basically says "expand to fill container width or height, whichever one is hit first". e.g. max-content, auto, etc.

bo-dun-1 (changes requested):

src/utils/image.js
obmelvin: I changed initialFiles default value to be `[]` from `null` and that triggers an infinite loop with the useEffect unless I add this conditional

src/elements/fields/FileUploadField.jsx
obmelvin: yeah I'll try that

src/elements/fields/FileUploadField.jsx
obmelvin: the conditional used both in the previous multi upload field. I'll double check to see if I can identify a case where that's needed, otherwise theoretically it can be just rawFiles

src/form/Form.jsx
obmelvin: it seemed weird to me to add a null value for the file, but I guess I should add a null to be consistent with the empty string that gets added for a repeated dropdown etc

src/form/Form.jsx
obmelvin: the problem was that `value` is an array of length 1 containing the newly updated file. since a field can't be repeating if it allows multiple files, we know it must only have a single value, so can safely access `[0]`. If the file isn't dereferenced, justInsert inserts an array inside the array, ie `[File, [File]]`. It didn't seem appropriate to modify justInsert given the intention of the function

I can add a comment

src/utils/image.js
obmelvin: make this static

src/form/Form.jsx
```diff
@@ -676,9 +676,14 @@ function Form({
...
+      const isPreviousValueDefaultArray =
+        Array.isArray(previousValue) &&
+        Array.isArray(defaultValue) &&
+        previousValue.length === 0 &&
+        defaultValue.length === 0;
       if (
         index === repeatedRowCount - 1 &&
-        previousValue === defaultValue &&
+        (previousValue === defaultValue || isPreviousValueDefaultArray) &&
         value !== defaultValue
```
bo-dun-1, line 687: similar to how you checked that the previous value was an empty array, you have to check here that the new value is not an empty array or value !== defaultvalue

bo-dun-1: and below for repeated row removal, you also need a similar check that previousValue is not an empty array and the new value is the empty array

bo-dun-1: this logic can probably all be decomposed

src/form/Form.jsx
bo-dun-1: yep ok that's what i thought, a comment would be helpful here!

bo-dun-1 (changes requested):

src/form/Form.jsx
```diff
@@ -331,10 +331,10 @@ function Form({
     const fieldKeys = [];
     repeatedServarFields.forEach((field) => {
       const { servar } = field;
-      updatedValues[servar.key] = [
-        ...fieldValues[servar.key],
-        getDefaultFieldValue(field)
-      ];
+      updatedValues[servar.key] =
+        servar.type === 'file_upload'
+          ? [...fieldValues[servar.key], null]
```
bo-dun-1, line 336: The ternary can just be limited to the value being added

src/form/Form.jsx
```diff
@@ -362,9 +362,12 @@ function Form({
     repeatedServarFields.forEach((field) => {
       const { servar } = field;
       const newRepeatedValues = justRemove(fieldValues[servar.key], index);
-      const defaultValue = getDefaultFieldValue(field);
+      const defaultValue =
+        servar.type === 'file_upload'
+          ? getDefaultFieldValue(field)
```
bo-dun-1, line 367: Getdefaultfieldvalue should be stored in a variable first

bo-dun-1 (changes requested):

bo-dun-1 (approved):

PR #395, opened 2022-05-13
support custom url button

bo-dun-1:


src/form/Form.jsx
```diff
@@ -1111,6 +1112,8 @@ function Form({
           () => clearLoaders()
         );
       }
+    } else if (link === LINK_URL) {
+      window.open(button.properties.link_url, '_blank');
```
obmelvin, line 1116: I'm pretty sure we want to add `'noopener noreferrer'` as a third param

obmelvin (approved):

PR #400, opened 2022-05-20
add support for multiple selections to button group

obmelvin:


src/elements/fields/ButtonGroupField.jsx
```diff
@@ -1,17 +1,25 @@
-import React from 'react';
+import React, { useMemo } from 'react';
 import { imgMaxSizeStyles } from '../styles';
 
 function ButtonGroupField({
   element,
   applyStyles,
   fieldLabel,
-  fieldVal = null,
+  fieldVal = [],
```
bo-dun-1, line 8: it might be good to leave fieldVal by default to be null since a default array param will trigger the useMemo every render

src/utils/formHelperFunctions.js
```diff
@@ -385,17 +392,30 @@ function shouldElementHide({ fields, values, element }) {
...
-    ? matchValues.includes(value)
-    : !matchValues.includes(value);
+  const matchFn = (val) =>
+    hideIf.comparison === 'equal'
+      ? matchValues.includes(val)
+      : !matchValues.includes(val);
+
+  if (Array.isArray(value)) {
+    return value.some((val) => {
+      if (Array.isArray(val)) {
```
bo-dun-1, line 410: can you leave a quick comment explaining when 2D arrays happen?

bo-dun-1 (approved):

src/utils/formHelperFunctions.js
obmelvin: I realized that I don't need to handle 2d array here - because when evaluating a repeated row this function is only exposed to that row's value. So it will just see a normal array, and can use the same logic that button group uses when not repeating

src/elements/fields/ButtonGroupField.jsx
obmelvin: ty for catching

PR #409, opened 2022-05-31
fix tracking default

obmelvin:


obmelvin:
leaving default as is, should be updating docs instead

PR #412, opened 2022-06-02
add stytch integration

obmelvin:


src/form/Form.jsx
```diff
@@ -952,7 +957,13 @@ function Form({
       const fieldVal = fieldValues[servar.key];
       if (servar.type === 'login') {
         setLoader();
-        return await sendLoginCode(fieldVal, servar);
+        // Unless we want to do something more complex here we need to make a
+        // choice and can't just return both. Prioritize stytch
+        if (integrations.stytch) {
+          return await sendMagicLink(fieldVal);
```
bo-dun-1, line 963: could we instead just return these directly and make this function not async

src/utils/init.js
```diff
@@ -152,6 +158,14 @@ function validateStep(formKey, trigger = true) {
   return callback(trigger);
 }
 
+function setAuthClient(client) {
+  initState.authClient = client;
+}
+
+function getAuthClient() {
```
bo-dun-1, line 165: out of curiosity: did you design getAuthClient to be backwards compatible with how upfront currently sets the firebase auth client? as in we aren't using a unique internal global name for the auth client, we're using the default firebase global.firebase? if so i agree with that since it means there's no immediate deprecation work necessary for upfront

src/integrations/utils.js
```diff
@@ -18,16 +19,23 @@ export function dynamicImport(dependency, async = true, index = 0) {
   }
 }
 
-export async function initializeIntegrations(integrations, clientArg, init) {
+export async function initializeIntegrations(integrations, clientArg) {
+  // This fixes the test that was failing, but it's not really clear to me why this started failing
```
bo-dun-1, line 23: which test?

src/integrations/stytch.js
```diff
@@ -0,0 +1,88 @@
...
+  const token = new URLSearchParams(window.location.search).get('token');
+  const client = getAuthClient();
+  // if jwt already exists, but stytch token still in URL params, then don't try
+  // to auth again as it will fail due to the info already being used to login
+  if (jwt || authSent || !client || !type || !token || !config) return;
+
+  authSent = true;
+
+  const opts = {
+    // TODO: remove this fallback after merge - only needed because I checked out older version of dashboard to update redirect URLs, and that version didn't have session duration
```
bo-dun-1, line 56: make sure this is removed

obmelvin:
I haven't set firebase up to use get/setAuthClient. that should probably be a follow up PR

src/integrations/utils.js
obmelvin: hmm, I removed it and the tests all pass

src/integrations/stytch.js
```diff
@@ -0,0 +1,85 @@
...
+          }
+        );
+      }
+    });
+    return stytchPromise;
+  }
+}
+
+export async function sendMagicLink(fieldVal) {
+  getAuthClient().magicLinks.email.loginOrCreate(fieldVal, {
```
bo-dun-1, line 34: should this be returned? even if the return value isn't being used right now

src/utils/init.js
bo-dun-1: yep makes sense

bo-dun-1 (approved):

src/integrations/stytch.js
obmelvin: yeah, good point

PR #439, opened 2022-06-23
Drewj/repeat fix

drewj:


src/form/gig/component.js
```diff
@@ -301,32 +302,76 @@ const repeat = (node, values, repeatIndex) => {
...
+    if (match) textVariables = match;
+  }
+
+  return textVariables.map((variable) => variable.slice(2, -2));
+};
+
+const getAllTextVariables = (node, variables = []) => {
+  const textVariables = getTextVariables(node);
+  if (textVariables)
+    textVariables.forEach((variable) => variables.push(variable));
```
obmelvin, line 320: I've not really been ensuring that all functions are pure, but it stuck out to me here that we are pushing directly into the param

bo-dun-1 (approved):
Probably good to note that we'll eventually need comprehensive unit tests for repeat logic @obmelvin

src/form/gig/component.js
bo-dun-1: Yeah likely better to return an array from the recursive function that you merge instead of doing param side effects

drewj: Yeah, I am pushing to it on purpose. This function is called recursively. The initial array is passed the recursive calls.

PR #448, opened 2022-06-28
Do not remove repeat unless explicit action taken and only adds repeat if last field changed

drewj:


src/form/Form.jsx
```diff
@@ -692,27 +692,27 @@ function Form({
...
-      if (
-        // index === repeatedRowCount - 1 &&
-        (previousValue === defaultValue || isPreviousValueDefaultArray) &&
-        value !== defaultValue
-      ) {
-        repeatRowOperation = 'add';
-      }
+      // And this is the last field in a set of repeated fields
+      const isLastRepeatedField =
+        Array.isArray(valueList) && index === valueList.length - 1;
```
bo-dun-1, line 706: nit: i think if you check if valueList is truthy you can assume it's an array here. if it breaks it's an upstream error anyways

src/form/Form.jsx
```diff
@@ -692,27 +692,27 @@ function Form({
...
 
-      // Remove a repeated row if the value went from set to unset
-      if (
-        (previousValue !== defaultValue && value === defaultValue) ||
-        (!isEmptyArray(previousValue) && isEmptyArray(value))
-      )
-        repeatRowOperation = 'remove';
+      if (isLastRepeatedField) {
+        if (
```
bo-dun-1, line 709: merge these two if statements into one?

bo-dun-1 (approved):

PR #455, opened 2022-07-01
fix order, undefined before defaultOptions and options are merged, if…

obmelvin:
… the option not specified by user


obmelvin:
superseeded by https://github.com/feathery-org/feathery-react/pull/456

PR #473, opened 2022-07-11
fix: Contain checkbox when cell width is "fit"

ondrosh:
## Changes

- Fix checkboxes not being contained properly within a cell with width "fit"


bo-dun-1 (changes requested):
i think drew pulled the the checkbox css composition into the CheckboxField.jsx file

drewj (approved):

PR #479, opened 2022-07-13
feat: Support background images and styling on cells

ondrosh:
## Changes

- Add support for background images on cells
- Add background image styles to cells

### DIFF
--- a/src/elements/styles.js
+++ b/src/elements/styles.js
@@ -358,6 +358,64 @@ class ApplyStyles {
       this.setStyle('placeholderFocus', 'display', 'none');
     }
   }
+
+  applyBackgroundImageStyles(target) {
+    const targetStyles = [
+      'background_image_url',
+      'background_image_display',
+      'background_image_layout',
+      'background_image_vertical_layout',
+      'background_image_size',
+      'background_image_size_x',
+      'background_image_size_y',
+      'background_image_repeat'
+    ];
+
+    this.apply(target, targetStyles, (...styles) => {
+      const [
+        imageUrl,
+        imageDisplay,
+        imageLayout,
+        imageVerticalLayout,
+        imageSize,
+        imageSizeX,
+        imageSizeY,
+        imageRepeat
+      ] = styles;
+
+      if (!imageUrl) {
+        return {};
+      }
+
+      const formattedStyles = {
+        backgroundImage: `url(${imageUrl})`,
+        backgroundRepeat: imageRepeat,
+        backgroundPositionX: imageLayout,
+        backgroundPositionY: imageVerticalLayout
+      };
+
+      switch (imageDisplay) {
+        case 'fill':
+        case 'fit':
+          Object.assign(formattedStyles, {
+            backgroundSize: imageSize
+          });
+          break;
+        case 'tile':
+          Object.assign(formattedStyles, {
+            backgroundSize: `${imageSize}%`
+          });
+          break;
+        case 'set_scale':
+          Object.assign(formattedStyles, {
+            backgroundSize: `${imageSizeX}px ${imageSizeY}px`
+          });
+          break;
+      }
+
+      return formattedStyles;
+    });
+  }
 }
 
 const noTextSelectStyles = {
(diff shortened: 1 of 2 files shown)


src/elements/styles.js
```diff
@@ -358,6 +358,64 @@ class ApplyStyles {
...
+      'background_image_display',
+      'background_image_layout',
+      'background_image_vertical_layout',
+      'background_image_size',
+      'background_image_size_x',
+      'background_image_size_y',
+      'background_image_repeat'
+    ];
+
+    this.apply(target, targetStyles, (...styles) => {
```
bo-dun-1, line 374: these styles have to be broken down into individual apply statements based on style groups. if you look at what apply does under the hood, it doesn't set mobile breakpoint styles if no mobile override exists for any of the styles passed in. but if at least one of the styles do exist, the mobile breakpoint CSS will be applied entirely.

passing all of these styles into the same apply function can create an unnecessary number of mobile overrides

src/elements/styles.js
```diff
@@ -358,6 +358,64 @@ class ApplyStyles {
...
+        backgroundImage: `url(${imageUrl})`,
+        backgroundRepeat: imageRepeat,
+        backgroundPositionX: imageLayout,
+        backgroundPositionY: imageVerticalLayout
+      };
+
+      switch (imageDisplay) {
+        case 'fill':
+        case 'fit':
+          Object.assign(formattedStyles, {
```
bo-dun-1, line 400: why not just formattedStyles.backgroundSize = imageSize

bo-dun-1 (changes requested):

src/elements/styles.js
```diff
@@ -358,6 +358,69 @@ class ApplyStyles {
       this.setStyle('placeholderFocus', 'display', 'none');
     }
   }
+
+  applyBackgroundImageStyles(target) {
+    if (
```
bo-dun-1, line 363: is this initial if statement necessary?

src/elements/styles.js
```diff
@@ -358,6 +358,69 @@ class ApplyStyles {
...
+    // Apply background image display (contain, fit, etc)
+    const displayTargets = [
+      'background_image_display',
+      'background_image_size',
+      'background_image_size_x',
+      'background_image_size_y'
+    ];
+
+    this.apply(target, displayTargets, (...styles) => {
+      const [imageDisplay, imageSize, imageSizeX, imageSizeY] = styles;
```
bo-dun-1, line 386: why not have these be explicit function parameters directly instead of aggregating then unaggregating them

src/elements/styles.js
```diff
@@ -358,6 +358,69 @@ class ApplyStyles {
...
+        case 'fill':
+        case 'fit':
+          displayStyles.backgroundSize = imageSize;
+          break;
+        case 'tile':
+          displayStyles.backgroundSize = `${imageSize}%`;
+          break;
+        case 'set_scale':
+          displayStyles.backgroundSize = `${imageSizeX}px ${imageSizeY}px`;
+          break;
```
bo-dun-1, line 399: don't all switch statements need a default case?

src/elements/styles.js
```diff
@@ -358,6 +358,69 @@ class ApplyStyles {
...
+      (imageLayout, imageVerticalLayout) => {
+        return {
+          backgroundPositionX: imageLayout,
+          backgroundPositionY: imageVerticalLayout
+        };
+      }
+    );
+
+    this.apply(target, ['background_image_repeat'], (imageRepeat) => {
+      if (imageRepeat) {
```
bo-dun-1, line 417: return {backgroundRepeat: imageRepeat || <DEFAULT CSS VALUE>}

src/elements/styles.js
ondrosh: To prevent the `this.apply` from following on multiple lines due to line width constraints. Just cleans it up and makes it easier to read, IMO

src/elements/styles.js
ondrosh: Good catch

src/elements/styles.js
bo-dun-1: i think it's fine to just check the styles dict since if mobileStyles has a background_image_url, then styles will have to have one. and if styles doesn't have one, there's a more serious bug upstream

bo-dun-1 (approved):

PR #482, opened 2022-07-15
fix feathery auth session when returning to embedded feathery form…

obmelvin:
…with existing stytch session

also cleans up the stytch code

### DIFF
--- a/src/integrations/stytch.js
+++ b/src/integrations/stytch.js
@@ -1,4 +1,3 @@
-import { getStytchJwt } from '../utils/browser';
 import { getAuthClient, setAuthClient } from '../utils/init';
 import {
   dynamicImport,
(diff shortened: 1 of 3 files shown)


src/integrations/stytch.js
```diff
@@ -59,45 +59,86 @@ export function sendMagicLink(fieldVal) {
...
-    authFn = () => stytchClient.magicLinks.authenticate(token, opts);
-  } else {
-    return;
+  if (jwt && session) {
+    console.log('session emailLogin', session);
+    // This path to the user email only works for magic link oauth doesn't
+    // contain the email as part of the auth factor. I do have access to a few
+    // stytch IDs (will post in github comments). Not sure if you can use that instead?
+    const authEmail =
+      session.authentication_factors[0]?.email_factor?.email_address ??
```
obmelvin, line 83: This path to the user email only works for magic link oauth doesn't
contain the email as part of the auth factor. I do have access to a few
stytch IDs (will post in github comments). Not sure if you can use that instead?

src/utils/init.js
```diff
@@ -158,8 +159,12 @@ function validateStep(formKey, trigger = true) {
   return callback(trigger);
 }
 
-function setAuthClient(client) {
+function setAuthClient(client, triggerLogin = true) {
   initState.authClient = client;
+  // NEED to handle stytch vs firebase - which emailLogin to call
+  // not sure on the best way to get session.integrations info here
+  // Is it saved somewhere I'm overlooking? should we save it to the client or init?
+  if (triggerLogin) emailLogin(defaultClient);
```
obmelvin, line 167: // NEED to handle stytch vs firebase - which emailLogin to call
  // not sure on the best way to get session.integrations info here
  // Is it saved somewhere I'm overlooking? should we save it to the client or init?

src/integrations/utils.js
```diff
@@ -51,5 +52,6 @@ export function transformQueryParamsToUrl() {
 
   const queryParams = new URLSearchParams(search);
   const redirect = queryParams.get('redirect');
-  return `${origin}${redirect ?? ''}`;
+  if (!redirect) return null;
```
obmelvin, line 55: I think I need to update both these functions to ensure they non-redirect query params & URL hashes

bo-dun-1: make this a ternary

src/integrations/stytch.js
```diff
@@ -59,45 +59,82 @@ export function sendMagicLink(fieldVal) {
...
+      'oliver@feathery.io';
+    removeStytchQueryParams();
+    return featheryClient
+      .submitAuthInfo({
+        authId: session.user_id,
+        authEmail,
+        is_stytch_template_key: config.is_stytch_template_key
+      })
+      .error(() => (authSent = false));
```
bo-dun-1, line 88: isn't it .catch?

src/integrations/stytch.js
```diff
@@ -59,45 +59,82 @@ export function sendMagicLink(fieldVal) {
...
+
+    return authFn()
+      .then((result) =>
+        featheryClient
+          .submitAuthInfo({
+            authId: result.user.user_id,
+            authEmail: result.user.emails[0].email,
+            is_stytch_template_key: config.is_stytch_template_key
+          })
+          .error(() => (authSent = false))
```
bo-dun-1, line 114: .catch?

bo-dun-1 (changes requested):

src/integrations/stytch.js
obmelvin: lol, yes

bo-dun-1 (approved):

PR #486, opened 2022-07-17
date selector support

bo-dun-1:
i had to regenerate the yarn lock and node_modules files to get the css file import working with webpack

### DIFF
--- a/src/elements/fields/DateSelectorField/index.jsx
+++ b/src/elements/fields/DateSelectorField/index.jsx
@@ -1,0 +1,102 @@
+import React, { memo, useState } from 'react';
+
+import Placeholder from '../../components/Placeholder';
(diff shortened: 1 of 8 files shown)


src/elements/fields/DateSelectorField/index.jsx
```diff
@@ -0,0 +1,100 @@
...
+  fieldLabel,
+  elementProps = {},
+  required = false,
+  editable = false,
+  onChange = () => {},
+  value = '',
+  inlineError,
+  children
+}) {
+  const [startDate, setStartDate] = useState(value ? new Date(value) : '');
```
obmelvin, line 33: Should this be called startDate if there's no endDate?

obmelvin (approved):
Looks good to me, though I'm mobile so haven't tested out the new field

PR #497, opened 2022-07-22
css in js for date selector

bo-dun-1:


obmelvin:
I think this makes sense to do. My concern was that I thought we'd have to transform the upstream CSS file anytime we needed to update - didn't realize you could just copy paste into a string literal. Sorry I should've still proposed this to you @ericpias

ericpias (approved):

PR #500, opened 2022-07-29
add typescript to externally facing files

obmelvin:


obmelvin:
Do we want to upgrade eslint to major version 7 to support the rule that is causing the circleci test to fail?

.eslintrc
```diff
@@ -1,11 +1,16 @@
 {
-  "parser": "babel-eslint",
+  // not 100% sure if we want to stay with the babel parser
```
obmelvin, line 2: see comments regarding ts parser vs babel parser

package.json
```diff
@@ -24,15 +25,18 @@
...
     "@babel/core": "^7.12.10",
+    "@babel/eslint-parser": "^7.18.2",
     "@babel/plugin-transform-runtime": "^7.14.3",
     "@babel/preset-env": "^7.12.11",
     "@babel/preset-react": "^7.12.10",
     "@testing-library/react": "^12.1.0",
     "@testing-library/user-event": "^13.2.1",
-    "babel-eslint": "^10.0.3",
+    "@typescript-eslint/eslint-plugin": "^5.28.0",
+    "@typescript-eslint/parser": "^5.28.0",
```
obmelvin, line 35: we should only keep this or @babel/eslint-parser in the end

src/form/Form.d.ts
```diff
@@ -0,0 +1,118 @@
...
+  lastStep: boolean;
+  setStep: SetStep;
+  firstStepLoaded: boolean;
+  integrationData: IntegrationData;
+}
+
+interface ContextOnSubmit extends Context {
+  // I haven't used the plaid integration so not quite sure how to type submitFields
+  // does including Plaid field values change the type or is it still fine as FormattedValues
+  submitFields: FormattedValues;
```
obmelvin, line 65: not sure about this type

src/utils/init.d.ts
```diff
@@ -0,0 +1,31 @@
...
+  apiKey: string,
+  options?: {
+    userKey?: null | string;
+    formKeys?: string[];
+    tracking?: 'cookie' | 'fingerprint';
+    authId?: string;
+    authEmail?: string;
+    authPhoneNumber?: string;
+  }
+  // not totally sure about this return type - <void> is the current implicit type
```
obmelvin, line 11: another question

tsconfig.json
```diff
@@ -0,0 +1,69 @@
+{
```
obmelvin, line 1: I left all these options in w/ the provided explanations so that people could review them. Will cut down to only those in use before merging

package.json
```diff
@@ -4,7 +4,8 @@
   "description": "React library for Feathery",
   "author": "Boyang Dun",
   "license": "MIT",
-  "main": "dist/index.js",
+  "repository": "https://github.com/feathery-org/feathery-react",
+  "main": "dist/index",
```
obmelvin, line 8: dropping the .js extension here was crucial for this all to work

bo-dun-1:
> Do we want to upgrade eslint to major version 7 to support the rule that is causing the circleci test to fail?

upgrading seems reasonable to me. will that require a refactor first?

src/form/Form.d.ts
```diff
@@ -0,0 +1,118 @@
...
+
+export interface ElementProps {
+  [fieldId: string]: {
+    [propName: string]: string;
+  };
+}
+
+type SetStep = (stepKey: string) => void;
+// am I missing trigger actions?
+type TriggerAction =
```
bo-dun-1, line 12: the onChange trigger can only be field or addressSelect currently while the onSubmit and onSkip can be click or change. this is all in our documentation https://docs.feathery.io/develop/react/onsubmit

src/form/Form.d.ts
bo-dun-1: yeah plaid fields are a slightly different structure since they're hidden fields. normal formatted field payloads are { displayText, value, type} while hidden field payload is just value. wonder if we should maybe structure it so hidden fields are more consistent with form fields? or shoudl we just update the typings

tsconfig.json
bo-dun-1: will defer to @drewj @ericpias @ondrosh here

tsconfig.json
```diff
@@ -0,0 +1,69 @@
...
+  "compilerOptions": {
+    /* Visit https://aka.ms/tsconfig.json to read more about this file */
+
+    /* Basic Options */
+    // "incremental": true,                   /* Enable incremental compilation */
+    "target": "es5" /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019', 'ES2020', or 'ESNEXT'. */,
+    "module": "esnext" /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', 'es2020', or 'ESNext'. */,
+    // "lib": [],                             /* Specify library files to be included in the compilation. */
+    // "allowJs": true /* Allow javascript files to be compiled. */,
+    // "checkJs": true,                       /* Report errors in .js files. */
```
ericpias, line 11: We don't need it now I guess, but options like allowJs and checkJs will likely be needed later if we begin introducing typescript into the lib.

tsconfig.json
ericpias: The enabled options seem reasonable to me when just adding typings.

src/utils/init.d.ts
```diff
@@ -0,0 +1,31 @@
...
+): Promise<void>;
+
+export function updateUserKey(newUserKey: string, merge?: boolean): void;
+export function setValues(userVals: FieldValues, rerender?: boolean): void;
+export function validateStep(formKey: string, trigger?: boolean): void;
+export function setAuthClient(authClient: any): void;
+export function getAuthClient(): any;
+
+export type FieldValues = {
+  [fieldName: string]:
```
ericpias, line 21: This may be slightly pedantic, but the docs use the term 'fieldKey' instead of 'fieldName'.  We should probably keep in consistent.

src/utils/init.d.ts
```diff
@@ -0,0 +1,31 @@
...
+    authId?: string;
+    authEmail?: string;
+    authPhoneNumber?: string;
+  }
+  // not totally sure about this return type - <void> is the current implicit type
+): Promise<void>;
+
+export function updateUserKey(newUserKey: string, merge?: boolean): void;
+export function setValues(userVals: FieldValues, rerender?: boolean): void;
+export function validateStep(formKey: string, trigger?: boolean): void;
```
ericpias, line 16: validateStep has return type {<fieldKey>: <errorMessage>}

src/elements/index.d.ts
```diff
@@ -0,0 +1,3 @@
+const Elements: JSX.Element[];
```
ericpias, line 1: I guess just using an untyped intrinsic here?  Seems fine for now but we will want to type deeper if the dashboard starts converting to typescript.

src/utils/init.d.ts
ericpias: Void seems right to me.  The docs make no "promises" on nor mention of the return type.

src/form/Form.d.ts
```diff
@@ -0,0 +1,118 @@
+/// <reference types="react" />
+import { FieldValues, setValues } from '../utils/init';
+
+export interface ElementProps {
+  [fieldId: string]: {
+    [propName: string]: string;
```
ericpias, line 6: The property can also be an Object type (specifically documented for a style but possibly other properties).  Not sure how deep this could go.

src/form/Form.d.ts
```diff
@@ -0,0 +1,118 @@
...
+  | 'click'
+  | 'change'
+  | 'addressSelect'
+  | 'load'
+  | 'blur';
+type Trigger = {
+  action: TriggerAction;
+  id: string;
+  text: string;
+  type: string; // we should maybe make a type for the element types
```
ericpias, line 23: Yes.  I think a type enum makes sense.

src/form/Form.d.ts
```diff
@@ -0,0 +1,118 @@
...
+};
+type IntegrationData = { firebaseAuthId?: string }; // this typing is incomplete
```
ericpias, line 25: Should we just punt on this and use any here instead?

ericpias (commented):
I would also suggest that we test with a react TS app embedding a form if you haven't already.  Also, it might be good to test it in a NextJS typescript app as well since we probably know that the first customer use case is there.

src/form/Form.d.ts
```diff
@@ -0,0 +1,118 @@
...
+type FormattedValues = {
```
ericpias, line 26: Should this be named 'FieldData' to better match the docs?

obmelvin:
> I would also suggest that we test with a react TS app embedding a form if you haven't already. Also, it might be good to test it in a NextJS typescript app as well since we probably know that the first customer use case is there.

I have tested it with local hosted forms for a couple of forms, but I haven't tried with a nextJS app. That would definitely be worthwile

src/utils/init.d.ts
bo-dun-1: if you look at the init definition, it returns an awaitable promise that returns void as you specified here. if the docs don't mention it could you go in and add it?

bo-dun-1: specifying the awaitability here IMO is important

src/index.d.ts
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Utility function which renders a form with the provided props in the DOM element with the provided ID.
+ * @param {string} elementId The ID of the DOM element to hold the form
+ * @param {Object} props The props defined on the *Form* component
+ */
+declare function renderAt(elementId: string, props: FormProps): void;
+declare const Feathery: {
+  Form: typeof Form;
+  Elements: typeof Elements;
```
bo-dun-1, line 24: have you tried linking this to the dashboard to test form elements?

bo-dun-1 (commented):
i tested some of the callback functions, the children component, and getAllValues in hosted forms. all seemed to work well for me!

src/utils/init.d.ts
obmelvin: We do say that it is an async function function in the docs

src/elements/index.d.ts
obmelvin: @ericpias in the new src/elements/index.ts file you can see that the Elements type is inferred from all the elements spread into the object

.eslintrc
```diff
@@ -28,6 +30,17 @@
     "react/jsx-handler-names": 0,
     "react/jsx-fragments": 0,
     "react/no-unused-prop-types": 0,
-    "import/export": 0
+    "import/export": 0,
+    "@typescript-eslint/no-explicit-any": 0, // Turn this back on after typing complete
```
bo-dun-1, line 34: make this a TODO

bo-dun-1 (approved):

PR #517, opened 2022-08-09
add override param to submitCustom so initialValues doesn't overwrite…

obmelvin:
… existing values


src/utils/client.js
```diff
@@ -316,6 +316,7 @@ export default class Client {
     formData.set('custom_key_values', JSON.stringify(jsonKeyVals));
     if (this.formKey) formData.set('form_key', this.formKey);
     if (userKey) formData.set('fuser_key', userKey);
+    if (!override) formData.set('override', override); // defaults to true on BE
```
bo-dun-1, line 319: might be better to always explicitly set override, would be slightly cleaner too

bo-dun-1 (approved):

PR #518, opened 2022-08-09
don't include query params in login redirect

obmelvin:
Don't include query parameters when logging in. They require whitelisting from stytch

### DIFF
--- a/src/integrations/stytch.js
+++ b/src/integrations/stytch.js
@@ -39,13 +39,25 @@ export function installStytch(stytchConfig) {
   }
 }
 
+export function googleOauthRedirect() {
+  const stytchClient = getAuthClient();
+  if (!stytchClient) return;
+
+  const redirectUrl = getRedirectUrl();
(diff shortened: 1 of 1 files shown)


src/integrations/stytch.js
```diff
@@ -129,12 +141,7 @@ function removeStytchQueryParams() {
...
-  const stytchClient = getAuthClient();
-  if (!stytchClient) return;
-
-  stytchClient.oauth.google.start({
-    login_redirect_url: window.location.href,
-    signup_redirect_url: window.location.href
-  });
+function getRedirectUrl() {
+  const { origin, pathname, hash } = window.location;
+  return `${origin}${pathname}${hash}`;
```
obmelvin, line 146: it would be fair to ask why I construct this URL instead of using URLSearchParams. It's just simpler to construct this rather than needing to iterate over URLSearchParams keys, delete the keys, and then get the URL

bo-dun-1 (approved):

PR #520, opened 2022-08-09
don't send magic link on button action if no email

obmelvin:


src/elements/fields/index.jsx
```diff
@@ -191,7 +191,8 @@ function applyFieldStyles(field, styles) {
       break;
     default:
       styles.addTargets('sub-fc', 'tooltipIcon');
-      if (type !== 'checkbox') styles.applyWidth('fc'); // Avoid applying width to checkbox to ensure the checkbox width is properly set by the component
+      // Avoid applying width to checkbox to ensure the checkbox width is properly set by the component
```
obmelvin, line 194: the autoformat kept moving this comment from the end of the line to be above the `default:` line, so I moved it to be in front of the appropriate line

src/utils/formHelperFunctions.js
```diff
@@ -367,7 +368,7 @@ async function setFormElementError({
  * @param inlineErrors
  */
 function getInlineError(field, inlineErrors) {
-  const data = inlineErrors[field.servar.key];
+  const data = inlineErrors[field?.servar?.key ?? field?.id];
```
obmelvin, line 371: took me a little bit to figure out that this was the last missing piece

src/utils/formHelperFunctions.js
```diff
@@ -154,8 +154,9 @@ const nextStepKey = (nextConditions, metadata, fieldValues) => {
         } else {
           if (rule.comparison === 'is_type') {
             ruleMet =
-              (ruleVal === 'email' && emailPattern.test(userVal)) ||
-              (ruleVal === 'phone' && phonePattern.test(userVal));
+              (ruleVal === 'email' && validators.email(userVal)) ||
+              // Not sure if this phone validator can be changed - not the same exact codepath (unlike the email changes)
+              (ruleVal === 'phone' && validators.phone(userVal));
```
obmelvin, line 159: as called out in the comment above, not sure if this line should be changed. that said, it's not quite clear looking at the code why we'd want to use the phonePattern instead of using `libphonenumber`

src/elements/fields/index.jsx
bo-dun-1: can you move it right above the line? instead of right below it

src/utils/formHelperFunctions.js
bo-dun-1: yeah because here we're just looking to see that it's formatted as a phone number, not that it's an actual valid phone number. that is only needed for explicit error-checking validation logic

src/utils/formHelperFunctions.js
bo-dun-1: oops yeah we kinda overload the definition of "id". imo maybe field.servar? field.servar.key : field.id

so we can get rid of the really broad elvis operators

src/utils/formHelperFunctions.js
obmelvin: I kinda feel that it is easier to maintain/simpler if there is just one way

bo-dun-1 (approved):

PR #526, opened 2022-08-15
lottie refactor

bo-dun-1:


src/elements/components/Lottie.jsx
```diff
@@ -0,0 +1,29 @@
...
+    LOTTIE_PROMISE.then(() =>
+      global.lottie.loadAnimation({
+        container: lottieRef.current,
+        renderer: 'svg',
+        loop: true,
+        autoplay: true,
+        animationData
+      })
+    );
+  }, [lottieRef.current]);
```
obmelvin, line 26: I'm pretty sure ref's don't work properly in a dep array because mutating the ref doesn't cause a re-render

That said, I'm assuming you've tested this so curious to get input from @drewj or @ondrosh

bo-dun-1: hmm good point, it seems like recommended approach is to use state + callback instead. let me reafactor

src/utils/client.js
```diff
@@ -194,6 +195,35 @@ export default class Client {
...
+          .then((font) => document.fonts.add(font))
+      );
+    });
+    // Load Lottie if form needs animations
+    let needLottie = false;
+    res.steps.forEach((step) =>
+      step.buttons.forEach((button) => {
+        if (needLottie) return; // Already loaded
+
+        const bp = button.properties;
```
obmelvin, line 219: you can do `const {loading_icon, loading_icon_type} = button.properties`

src/elements/components/Lottie.jsx
bo-dun-1: actually apparently you can just assume it's mounted in the useEffect, which makes sense

obmelvin (approved):

PR #531, opened 2022-08-16
Stripe Integration

ericpias:
This is V1 of stripe integration and supports payment method(s) setup only.
See functional requirements: https://docs.google.com/document/d/1etw2-A-zjceqvqJmNJYBKNSX6unxCf3Gnd5jVbyqS8I/edit?usp=sharing
See Tech Integration: https://docs.google.com/document/d/1vqrKkI-iF4se0JcBQjxyLl8vb_-yxpC3lIqIe0we3-k/edit?usp=sharing

Test Outline:

- Optional PM field and no card entered - does NOT setup
- Optional PM field and card entered - does setup
- Required PM field and no card entered - required field error
- Required PM field and card entered - does setup
- Card entry incomplete - cannot submit - error shown
- Test live form and review and make sure the proper stripe key is used.  Simulate using two stripe accounts.  Use hosted forms with diff feathery api keys.  Using 2 stripe accounts ([eric.pias@comcast.net](mailto:eric.pias@comcast.net) (PW saved in Safari) and [ericdpias@gmail.com](mailto:ericdpias@gmail.com) (FF)).
- Multi-step form
- Setup card and then go back, must setup PM again if required or card entered.
- Verify card not setup on steps where card does not appear.
- Test multiple cards  (must both go to same customer)
- Two cards one step
- Two steps, each with a PM field
- Two steps, first 2 cards, 2nd 1 card.
- Test a form with no PM field
- Test DB master against stripe Lib and BE.


src/elements/fields/PaymentMethodField.jsx
```diff
@@ -0,0 +1,191 @@
...
+  'lineHeight',
+  'letterSpacing',
+  'textAlign',
+  'padding',
+  'textDecoration',
+  'textShadow',
+  'textTransform'
+];
+
+//
```
obmelvin, line 37: empty comment :)

src/elements/fields/PaymentMethodField.jsx
```diff
@@ -0,0 +1,191 @@
...
+  useStripe
+} from '@stripe/react-stripe-js';
+
+import { getStripePromise } from '../../integrations/stripe';
+import { filterStyleValues } from '../../utils/styles';
+import { filterKeys } from '../../utils/utils';
+
+import InlineTooltip from '../components/Tooltip';
+
+const stripeKey = 'any-key-does-not-matter-what';
```
obmelvin, line 15: not sure if this was needed during dev, or still needed. would benefit from a comment explaining why you need a dummy key

src/elements/fields/PaymentMethodField.jsx
```diff
@@ -0,0 +1,191 @@
...
+      }
+    },
+    classes: {
+      base: element.servar.key
+    }
+  };
+  const handleCardChange = async (event) => {
+    // Listen for changes in the CardElement
+    // and display any errors as the customer types their card details
+    setFieldError(event.error ? event.error.message : '');
```
obmelvin, line 98: you could do `event.error?.message ?? ''`

src/elements/fields/index.jsx
```diff
@@ -189,6 +191,37 @@ function applyFieldStyles(field, styles) {
       styles.applyCorners('field');
       styles.applyBorders('field');
       break;
+    case 'payment_method':
+      styles.addTargets('sub-fc', 'tooltipIcon', 'completed');
+      styles.applyHeight('sub-fc');
+      // styles.applyWidth('sub-fc');
```
obmelvin, line 197: double check if you need this style or not

src/integrations/stripe.js
```diff
@@ -0,0 +1,101 @@
+import { loadStripe } from '@stripe/stripe-js';
```
obmelvin, line 1: the other integrations are all loaded async rather than being installed in package.json. can we not do that here?

webpack.development.js
```diff
@@ -1,19 +1,19 @@
...
-module.exports = config;
+  config.devtool = 'eval-cheap-module-source-map';
+  config.externals = ['react'];
+  config.plugins.push(
+    new webpack.optimize.LimitChunkCountPlugin({
+      maxChunks: 1
+    })
+  );
+  return config;
+};
```
obmelvin, line 19: not too important, but apparently missing new line at end of file

src/elements/fields/PaymentMethodField.jsx
ericpias: Yes it is still used in the dashboard only.  Adding a comment.

src/elements/fields/index.jsx
ericpias: Not needed, width is in 'fc' only.

src/integrations/stripe.js
ericpias: Stripe-js is a very light weight wrapper lib that does really one thing and that is load stripe.js from a secure, PSI-compliant location.  Yes, we could duplicate that and I looked into doing that, but it is not trivial and why re-invent the wheel here so this is the way I went.  The uncompressed size of stripe-js is 3.4k.  Quite small.

ericpias: BTW, the actual loading of stripe.js (the larger and secure one) is done async and only if the integration is active.

obmelvin: ok cool, I saw that it was importing loadStripe but wasn't 100% sure if that was happening. I think I'm ok with this if @bo-dun-1 is

package.json
```diff
@@ -15,6 +15,7 @@
     "build:node": "webpack --mode production --config webpack.node.js",
     "build:umd": "webpack --mode production --config webpack.umd.js",
     "dev": "webpack --mode development --config webpack.development.js --watch",
+    "bundle-analysis": "webpack --mode development --config webpack.development.js --env analyze",
```
bo-dun-1, line 18: analyze-bundle?

src/elements/fields/PaymentMethodField.jsx
```diff
@@ -0,0 +1,193 @@
...
+      return;
+    }
+    setCardElement(element.servar.key, elements.getElement(CardElement));
+  }, [stripe, element.servar.key, elements]);
+
+  const cardElementOptions = {
+    style: {
+      base: {
+        textDecoration: 'none', // Bug in card element - force the reset to none
+        ...filterKeys(applyStyles.getTarget('field'), supportedStripeCssProps),
```
bo-dun-1, line 68: does the stripe card element support mobile styles? in this current usage of applystyles we are not fetching the mobile styles, which are stored as an object in a media query

src/elements/fields/PaymentMethodField.jsx
```diff
@@ -0,0 +1,193 @@
...
+        '::placeholder': {
+          textDecoration: 'none', // Bug in card element - force the reset to none
+          ...filterKeys(
+            applyStyles.getTarget('placeholder'),
+            supportedStripeCssProps
+          )
+        },
+        ':hover': filterStyleValues(
```
bo-dun-1, line 76: instead of retroactively filtering out teh !important suffix, it might be cleaner to pass in a flag to omit the !important flag when generating the hover/selected styles for applyStyles

src/elements/fields/PaymentMethodField.jsx
```diff
@@ -0,0 +1,193 @@
...
+          filterKeys(applyStyles.getTarget('hover'), supportedStripeCssProps),
+          '!important'
+        ),
+        ':focus': filterStyleValues(
+          filterKeys(applyStyles.getTarget('active'), supportedStripeCssProps),
+          '!important'
+        )
+      },
+      complete: {
+        ...applyStyles.getTarget('completed')
```
bo-dun-1, line 86: no need to be spread

src/elements/fields/PaymentMethodField.jsx
```diff
@@ -0,0 +1,193 @@
...
+      },
+      invalid: {
+        color: '#d6504a',
```
bo-dun-1, line 89: the error color we're using in the react lib right now is #F42525, e.g. with inline errors. we might want to make that a const in utils and use it everywhere in teh lib, including here

src/elements/fields/PaymentMethodField.jsx
```diff
@@ -0,0 +1,193 @@
...
+          ...applyStyles.getTarget('sub-fc'),
+          '&:focus': applyStyles.getTarget('active'),
+          '&:hover': applyStyles.getTarget('hover'),
+          ...(inlineError ? { borderColor: '#F42525' } : {})
+        }}
+      >
+        <div css={{ width: '100%', position: 'relative' }}>
+          {/* position an input field under the card element to support html5 errors */}
+          {runtime && errorDisplayMode === 'html5' && (
+            <input
```
bo-dun-1, line 139: we do something similar with a hdiden input field for hte file _upload element. we might want to unify those components

src/elements/fields/PaymentMethodField.jsx
```diff
@@ -0,0 +1,193 @@
...
+    if (!runtime)
+      document.dispatchEvent(
+        new CustomEvent('stripe_key_loaded', {
+          detail: { key: stripeKey }
+        })
+      );
+  }, [runtime]);
+
+  return (
+    <>
```
bo-dun-1, line 185: fragment is unnecessary here

src/integrations/stripe.js
```diff
@@ -0,0 +1,101 @@
...
+    promise.then((stripe) => {
+      resolve(stripe);
+    });
+  });
+});
+
+export function getStripePromise() {
+  return stripePromise;
+}
+export async function getStripe() {
```
bo-dun-1, line 15: the async and await can be removed here, and you can just call await getStripe()

src/integrations/stripe.js
```diff
@@ -0,0 +1,101 @@
...
+const stripePromise = new Promise((resolve) => {
+  document.addEventListener('stripe_key_loaded', (e) => {
+    const promise = loadStripe(e.detail.key);
+    promise.then((stripe) => {
+      resolve(stripe);
+    });
+  });
+});
+
+export function getStripePromise() {
```
bo-dun-1, line 12: const getStripePromise = () => stripePromise;

src/integrations/stripe.js
bo-dun-1: export const getStripe = () => getStripePromise()

but this also seems redundant with the above function

src/integrations/stripe.js
```diff
@@ -0,0 +1,101 @@
...
+
+export function getStripeCustomerFieldValues(stripeConfig, fieldValues) {
+  if (!stripeConfig?.metadata?.customer_field_mappings) return {};
+  const customer = getObjectMappingValues(
+    stripeConfig.metadata.customer_field_mappings,
+    fieldValues
+  );
+  // the shipping object in customer must have a name property (stripe requires this),
+  //  but we don't want to make the mapping required.  So default it here.
+  customer.shipping &&
```
bo-dun-1, line 38: i would make this more explicit.

if (!customer.shipping?.name) customer.shipping.name = 'Name not supplied'

src/form/Form.tsx
```diff
@@ -953,6 +962,63 @@ function Form({
     });
     if (invalid) return;
 
+    // Perform any Stripe payment method setup here if and only if the
+    // payment method field has a complete value
+    const paymentMethodFields = getPaymentMethodFieldsOnStep(activeStep);
+    const haveCompletePaymentMethod = paymentMethodFields.some((pmf: {servar: {key: string}}) =>
+      // @ts-expect-error
```
bo-dun-1, line 969: can we add a comment to each expect-error so we know why we're leaving an error in?

src/form/Form.tsx
```diff
@@ -953,6 +962,63 @@ function Form({
...
+      Boolean(formattedFields[pmf.servar.key].value?.complete)
```
bo-dun-1, line 970: Boolean() isn't necessary here

src/form/Form.tsx
```diff
@@ -953,6 +962,63 @@ function Form({
...
+    );
+    // @ts-expect-error
+    if (integrations.stripe && haveCompletePaymentMethod) {
```
bo-dun-1, line 973: all of this logic should live in handleActions below, which was designed to be the entrypoint for integration submissino logic

bo-dun-1: in general we should minimize the amount of stripe-specific logic in Form.tsx, which should be for pure form logic and is already quite hefty.

src/form/grid/Cell.jsx
```diff
@@ -390,6 +394,31 @@ const Cell = ({ node: el, form }) => {
             }}
           />
         );
+      case 'payment_method':
+        return (
+          <Elements.PaymentMethodField
+            {...fieldProps}
+            runtime
+            setCardElement={setCardElement}
+            setFieldError={async (message) => {
```
bo-dun-1, line 403: async/await can be removed, and return setFormElementError from the function directly

src/integrations/stripe.js
```diff
@@ -0,0 +1,101 @@
...
+  client,
+  formFieldValues,
+  paymentMethodFieldId,
+  stripeConfig,
+  stripe,
+  cardElement
+) {
+  if (!stripeConfig || Object.keys(stripeConfig).length === 0) return null;
+
+  const { intent_secret: intentSecret } = await client.setupPaymentIntent(
```
bo-dun-1, line 82: can any of this payment method setup logic be moved to the backend? in general whatever we can handle on the backend, we should. it'll also keep the react lib more lightweight

webpack.development.js
```diff
@@ -1,19 +1,19 @@
...
-module.exports = config;
+  config.devtool = 'eval-cheap-module-source-map';
+  config.externals = ['react'];
+  config.plugins.push(
+    new webpack.optimize.LimitChunkCountPlugin({
+      maxChunks: 1
+    })
+  );
+  return config;
+}
```
bo-dun-1, line 19: think we need a newline here

webpack.development.js
```diff
@@ -1,19 +1,19 @@
...
-config.plugins.push(
-  new webpack.optimize.LimitChunkCountPlugin({
-    maxChunks: 1
-  })
-);
+module.exports = (env) => {
+  if (env.analyze) {
+    const BundleAnalyzerPlugin =
+      require('webpack-bundle-analyzer').BundleAnalyzerPlugin;
+    config.plugins.push(new BundleAnalyzerPlugin());
```
bo-dun-1, line 8: can we include the flags? defaultSizes: 'stat', openAnalyzer: true

bo-dun-1 (changes requested):

src/utils/utils.js
```diff
@@ -0,0 +1,7 @@
+export function filterKeys(obj, allowedKeys) {
```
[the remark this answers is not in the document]
ericpias: Moved to primitives

src/elements/fields/PaymentMethodField.jsx
ericpias: It made sense to put in src/elements/styles.

src/integrations/stripe.js
ericpias: Yes, simplification is in order here.  Combining them into one.

src/integrations/stripe.js
ericpias: There is a subtle but important difference between the code and what is suggested.  customer.shipping is not required, but if present, name must be there.  So changed it to be:
if (customer.shipping && !customer.shipping.name) customer.shipping.name = 'Name not supplied'

src/integrations/stripe.js
ericpias: To the best of my knowledge we have moved everything possible to the backend.  The stripe.confirmCardSetup must be done from the client api in order to send the cardElement with all the card info like number, exp. cvc, etc.

src/elements/fields/PaymentMethodField.jsx
ericpias: Changed ApplyStyles.applySelectorStyles just a bit (with a unit test as well) so that the !important is parameterized.  Removed filterStyleValues() completely.

src/form/grid/Cell.jsx
ericpias: Yep, I refactored how I was doing this and left this in.  Fixed.

src/elements/fields/PaymentMethodField.jsx
ericpias: @bo-dun-1 I looked at the input in FileUploadField.  I don't see a lot of commonality other than they both serve as 'anchors' for an html5 error.  The one in FileUploadField actually serves a functional purpose beyond that.  The one in PaymentMethod only serves as the anchor.  So, I just don't see the value in commonizing them.  If you disagree, maybe you can offer some specifics?

src/form/Form.tsx
ericpias: Ok, good to know regarding the purpose of handleActions.  I missed that.  I moved the stripe card setup to there which is cleaner and makes use of the existing design to set the field error, etc.

src/elements/fields/PaymentMethodField.jsx
ericpias: @bo-dun-1 So I looked into this and understand mobile styles and how they work a bit more.  In general, mobile styles are supported for the payment method field.  However, we do this using media queries.  The Stripe card element does not support this so the mobile styles that are set in media queries directly in the card element's style object are (safely) ignored.  This means that mobile styles like color and the various font styles do not work (mobile settings will not be applied).  However, other mobile styles for things like background color and borders are set in the parent of the card element and DO WORK.

I am sure we could make all the mobile styles work using javascript matchMedia, etc. instead of media queries.  However, I am inclined to ignore this issue for now and I have put a comment similar to the above in the payment method component code.  If you think this must instead be addressed/fixed now, let me know.

src/form/Form.tsx
ericpias: Regarding the second comment, I moved the bulk of the code into stripe.js in Integrations in a new method: setupPaymentMethods.

src/form/Form.tsx
ericpias: With other changes I made, much of this code moved out of Form.tsx.  I commented the remaining expected TS error.

src/elements/fields/PaymentMethodField.jsx
bo-dun-1: makes sense! let's ignore for now

src/elements/fields/PaymentMethodField.jsx
```diff
@@ -0,0 +1,194 @@
...
+      invalid: {
+        color: ERROR_COLOR,
+        iconColor: ERROR_COLOR
+      }
+    },
+    classes: {
+      base: element.servar.key
+    }
+  };
+  const handleCardChange = async (event) => {
```
bo-dun-1, line 100: this doesn't seem like it needs to be async since no await is being used

src/elements/fields/PaymentMethodField.jsx
bo-dun-1: maybe at least their styles? currently file uses opacity and this one uses background

src/form/Form.tsx
```diff
@@ -1053,7 +1059,31 @@ function Form({
...
+    setCardElementMappings((cardElementMappings) => ({
+      ...cardElementMappings,
+      [key]: cardElement
+    }));
+  }
+
+  async function handleActions(setLoader: any, formattedFields: any) {
+    // Perform any Stripe payment methods setup here
+    try {
+      await setupPaymentMethods(
```
bo-dun-1, line 1072: you can't try/catch errors from async functions like this, you'd have to do 

const handleActions = (...) => setupPaymentMethods(...).catch((p) => p);

src/integrations/stripe.js
```diff
@@ -0,0 +1,149 @@
...
+          updateFieldValues({
+            [paymentMethodField.servar.key]: paymentMethodData
+          });
+          formattedFields[paymentMethodField.servar.key] = {
+            value: paymentMethodData,
+            type: paymentMethodField.servar.type,
+            displayText: paymentMethodField.servar.name
+          };
+        } catch (e) {
+          throw Promise.resolve({
```
bo-dun-1, line 141: you should be able to throw the error directly

src/form/Form.tsx
```diff
@@ -166,6 +168,9 @@ function Form({
 
   const [repeats, setRepeats] = useState(0);
 
+  // Stripe - the card elements on the active step, if any
+  const [cardElementMappings, setCardElementMappings] = useState({});
```
bo-dun-1, line 172: this state management, the setCardElement function below, and the setupPAymentMethod function initialization imo can all be condensed into a single usePayments custom hook

src/integrations/stripe.js
```diff
@@ -0,0 +1,149 @@
...
+  fieldValues,
+  formattedFields,
+  stripeConfig,
+  cardElementMappings,
+  updateFieldValues,
+  setLoader
+) {
+  // Perform any Stripe payment method setup here if and only if the
+  // payment method field has a complete value
+  const paymentMethodFields = getPaymentMethodFieldsOnStep(activeStep);
```
bo-dun-1, line 109: instead of doing this in the setupPaymentMethods function, handleActions already iterates through all of the step fields to see which ones an action needs to be performed on. i think we should leverage the handleActions logic instead of recreating it here

webpack.development.js
```diff
@@ -1,19 +1,21 @@
...
 const webpack = require('webpack');
 
-config.devtool = 'eval-cheap-module-source-map';
-config.externals = ['react'];
-config.plugins.push(
-  new webpack.optimize.LimitChunkCountPlugin({
-    maxChunks: 1
-  })
-);
+module.exports = (env) => {
```
bo-dun-1, line 4: can we pull this bundle analyzer refactor into a separate PR that we can merge first?

src/form/Form.tsx
obmelvin: I might be missing a bit of nuance about what exactly this code is trying to do, but you can use try/catch blocks within an async fn - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await#promise_rejection

src/integrations/stripe.js
ericpias: @bo-dun-1 I purposely put this before the login and Pin actions because it seems logical to me that card setup needs to succeed first before and login/pin input attempt.  I think it matters, but maybe not?

src/form/Form.tsx
ericpias: I simplified here to remove the throw and catch.  A simple error return was sufficient.

src/integrations/stripe.js
ericpias: I simplified here to remove the throw and catch. A simple error return was sufficient.

src/form/Form.tsx
ericpias: Created the hook function usePayments.

src/integrations/stripe.js
ericpias: @bo-dun-1 The setup payment method "action" needs to happen for each payment method field on the step and like I said before it seems that it should be done before the login/PIN input actions.  Also, for the login/pin actions, the logic is purposely written to only do one of them if there are multiple.  It just seems these are separate and different purposes here, so I left the setupPaymentMethods() as it was (before the other actions).  If there is a problem setting up a payment method, then the login/pin won't even be attempted and that seems correct to me.

If you really disagree, of course, push back.  @obmelvin and I discussed and he basically agreed.

src/elements/fields/PaymentMethodField.jsx
```diff
@@ -0,0 +1,194 @@
...
+          alignItems: 'center',
+          ...applyStyles.getTarget('sub-fc'),
+          '&:focus': applyStyles.getTarget('active'),
+          '&:hover': applyStyles.getTarget('hover'),
+          ...(inlineError ? { borderColor: ERROR_COLOR } : {})
+        }}
+      >
+        <div css={{ width: '100%', position: 'relative' }}>
+          {/* position an input field under the card element to support html5 errors */}
+          {runtime && errorDisplayMode === 'html5' && (
```
bo-dun-1, line 141: editable is set to true in dashboard and falsey in react lib, so you don't need an additional runtime boolean here

ericpias: Ah, I missed editable.  Good to know.  Fixed.

src/integrations/stripe.js
ericpias: As we discussed, I implemented an array of action configurations that groups actions by servar type, orders them, configures continue behavior, etc.  The handleActions() method now has a generic loop to runs through the action configs by servar type and executes what if finds matching fields for.

src/elements/fields/PaymentMethodField.jsx
```diff
@@ -0,0 +1,193 @@
...
+        </div>
+        <InlineTooltip element={element} applyStyles={applyStyles} />
+      </div>
+    </div>
+  );
+};
+
+function PaymentMethodField({ editable = false, ...props }) {
+  useEffect(() => {
+    // if not runtime, i.e. running in dashboard, then just load stripe with key so cardElement shows
```
bo-dun-1, line 177: if editable

src/form/Form.tsx
```diff
@@ -1053,29 +1056,88 @@ function Form({
...
-        setLoader();
-        // Unless we want to do something more complex here we need to make a
-        // choice and can't just return both. Prioritize stytch
-        // @ts-expect-error
-        if (integrations.stytch) {
-          return sendMagicLink(fieldVal);
-        } else {
-          return sendLoginCode(fieldVal, servar);
+  // usePayments /Stripe
+  const [setCardElement, setupPaymentMethod] = usePayments();
```
bo-dun-1, line 1060: i think it would be cleaner to pull setupPaymentMethod out of the hook, then you can pull the entire actionConfigurations out into a global const too.

src/form/Form.tsx
```diff
@@ -1053,29 +1056,88 @@ function Form({
...
+        integrationKey: 'firebase',
+        actionFn: ({ fieldVal, servar }: ActionData) =>
+          sendLoginCode(fieldVal, servar),
+        continue: false
+      },
+      {
+        servarType: 'pin_input',
+        integrationKey: 'firebase',
+        isMatch: ({ servar }: ActionData) => servar.metadata.verify_sms_code,
+        actionFn: ({ fieldVal, servar, client }: ActionData) =>
```
bo-dun-1, line 1097: you can just modify the helper functions directly to take in ActionData instead of having an intermediate actionFn parse it

src/form/Form.tsx
bo-dun-1: i might put it in integrations/utils

src/form/Form.tsx
ericpias: These functions are called from elsewhere as well.  I was trying to avoid the regression test effort by not changing the signatures, but I guess I will just do it.

src/form/Form.tsx
ericpias: One alternative that I just that thought of is to move targetElementFn out of actionConfigurations and then in handleActions always just set a function called getCardElement into the ActionData (but only stripe would ever use it).  Not perfect, but I think actionConfigurations could then be a static global.  I favor leaving it as is, but if you like this idea  or some other idea better let me know.

bo-dun-1 (approved):

PR #539, opened 2022-08-27
check if stytch or fb are loaded before inferring email login

bo-dun-1:


src/integrations/utils.ts
```diff
@@ -64,7 +64,7 @@ export async function initializeIntegrations(
   ]);
 
   if (gtm) initializeTagManager(gtm);
-  inferEmailLoginFromURL(clientArg);
+  if (fb || stytch) inferEmailLoginFromURL(clientArg);
```
obmelvin, line 67: I think the login code checks to see if the client exists, but I don't see anything wrong with this check

obmelvin (approved):

PR #552, opened 2022-09-09
Stripe Integration V2: Product selection and payment

ericpias:
This pull request is for V2 of the stripe integration and covers:

1. Support for selecting and deselecting configured stripe products and the totaling of the cost into a configured hidden field.  This cost can be shown with a text var (optional)
2. Support for group actions on any sub-grid/group, including the new "Select stripe product" action.  
3. Support for the new "Select stripe product" action on buttons (when Stripe integration configured).
4. Support for payments/charges in stripe if there are products selected and/or a non-zero cost.
5. The setting of a paid indicator to '1' in order to facilitate optional navigation and hide rules to prevent payment twice (if the designer chooses this).

The detailed functional requirements are here: https://docs.google.com/document/d/1etw2-A-zjceqvqJmNJYBKNSX6unxCf3Gnd5jVbyqS8I/edit#heading=h.jzois51nxngw

Here is my test outline:

- Group Actions
  - Test a group inside a group, both with click actions - only one action works
  - Test that Navigate to url works
  - Test that Firebase/SMS works
  - Test Firebase/magic link works
  - Test that Stytch magic link is called properly
  - Test that Stytch OAuth is called properly

- Product selection

  - Test additional cost - added to total
  - Test multiple products
    - Single-select - groups are many to one with hidden field
    - Multi select - group is one-to-one with hidden field
  - Test one-time product price
  - Test that recurring stripe product cost is added into total 
  - Test that package, graduated, volume and customer sets price product pricing schemes, if used do NOT add in cost when selected.
  - Test that if a non-USD price is in stripe we treat it as USD regardless
  - Test that summed cost field is set if mapped
    - Test that total is displayed with text var {{}}
  - Test that if total summed field is not mapped, then product selection still works (total just not shown)
  - Test that if product id not configured, product button cannot be selected
  - Test that if product selected hidden field not configured, product button cannot be selected
  - Test that total matches the selected products after screen refresh

- Payment
  - Test product selection and payment method on same step
  - Test product selection on first step and payment method on second step
  - Test that if no cost (no products selected), no payment/charge is done
  - Test if non-zero additional cost, a payment/charge is done
  - Test that receipt email is set on payment (if field mapped) and that receipt email sent
  - Payment indicator
    - If mapped make sure set
    - If mapped, test it in a navigation rule
    - If not mapped, payment still works without error
  - Test that on form re-entry or on a back navigation after product selection and payment that all product selections and total are cleared
  - Test that if total summed field is not mapped, then payment still works (total just not shown)
  - Test that recurring stripe product is charged for
- Test against NextJS


src/elements/basic/ButtonElement.tsx
```diff
@@ -82,9 +70,13 @@ function applyButtonStyles(element: any, applyStyles: any) {
   }
   applyStyles.applyBorders('buttonActive', 'selected_');
   if (element.styles.selected_background_color) {
-    applyStyles.apply('buttonHover', 'selected_background_color', (a: any) => ({
-      backgroundColor: `#${a} !important`
-    }));
+    applyStyles.apply(
+      'buttonActive',
+      'selected_background_color',
+      (a: any) => ({
```
bo-dun-1, line 76: : string

src/form/grid/index.tsx
```diff
@@ -74,14 +85,39 @@ const Subgrid = ({
...
   applyStyles.apply('cell', 'background_color', (c: any) => ({
     backgroundColor: `#${c}`
   }));
-  return applyStyles.getTarget('cell');
+  applyStyles.apply('cellHover', 'background_color', (a: any) => {
+    const color = `${adjustColor(a || 'ffffffff', -20)}!important`;
+    return {
+      backgroundColor: color,
+      borderColor: color,
+      transition: 'background 0.3s !important'
```
bo-dun-1, line 104: maybe this transition style can be applied statically to the cell? applyStyles is only necessary for dynamic styles that may vary between mobile and desktop

src/form/grid/index.tsx
```diff
@@ -118,20 +154,61 @@ const getCellContainerStyle = (axis: string, layout: string) => {
...
+
+  if (cellData) {
+    const [cellStyle, cellHoverStyle, cellActiveStyle] = getCellStyle(cellData);
+    const {
+      link = LINK_NONE,
+      product_id: productId,
+      selected_product_id_field: selectedProductIdField
+    } = properties;
+
+    const hasSubGridLink = link && link !== LINK_NONE && link !== LINK_STRIPE;
```
bo-dun-1, line 184: this coudl just be `![LINK_NONE, LINK_STRIPE].includes(link)`

src/integrations/stripe.ts
```diff
@@ -15,82 +17,91 @@ const stripePromise = new Promise((resolve) => {
...
+  if (metadata) {
+    // eslint-disable-next-line camelcase
+    if (metadata.client_key) {
+      document.dispatchEvent(
+        new CustomEvent('stripe_key_loaded', {
+          detail: { key: stripeConfig.metadata.client_key }
+        })
+      );
+    }
+    // If a total hidden field is configured, then initialize it to any addition cost
```
bo-dun-1, line 30: this seems like it could be done on the backend instead

src/integrations/stripe.ts
```diff
@@ -127,6 +139,51 @@ export async function setupPaymentMethod(
...
+                [totalFieldKey]: '0.00'
+              });
+            }
+
+            // set any mapped paid indicator field
+            const paidIndicatorFieldKey =
+              integrationData.metadata.payment_field_mappings?.paid_indicator;
+            if (paidIndicatorFieldKey)
+              updateFieldValues({
+                [paidIndicatorFieldKey]: '1'
```
bo-dun-1, line 183: why is this a string and not a boolean?

bo-dun-1: or even an int

src/integrations/stripe.ts
```diff
@@ -127,6 +139,51 @@ export async function setupPaymentMethod(
...
+              updateFieldValues({
+                [key]: ''
+              })
+            );
+            // clear the total in the total hidden field if mapped (leaving the currency symbol off)
+            const totalFieldKey =
+              integrationData.metadata?.payment_field_mappings?.total;
+            if (totalFieldKey) {
+              updateFieldValues({
+                [totalFieldKey]: '0.00'
```
bo-dun-1, line 174: shoudl this be a float instead of string?

src/integrations/stripe.ts
```diff
@@ -127,6 +139,51 @@ export async function setupPaymentMethod(
...
+              errorField: servar,
+              errorMessage: result.error.message
+            };
+          else {
+            // Complete the payment (clears fields, sets paid indicator)
+            const { product_fields_to_clear: productFieldKeys } =
+              await client.paymentComplete();
+
+            // clear the selected product hidden fields
+            productFieldKeys.forEach((key: string) =>
```
bo-dun-1, line 164: instead of doing all this logic below to manually figure out which fields to clear, could we just have the backend return updated field value mappings we can apply directly to updateFieldValues?? i think whatever business logic we can move into the backend, we should be momving

src/integrations/stripe.ts
astrozzc: At the beginning, the hidden field total does not have a value yet (until a product is selected and total calculated). Should I find stripe integration, check the total, create a hiddenfieldvalue for it here? https://github.com/feathery-org/feathery-backend/blob/master/apps/api/views.py#L280

src/integrations/stripe.ts
```diff
@@ -138,6 +195,92 @@ export async function setupPaymentMethod(
...
+  if (
+    productId &&
+    selectedProductIdFieldKey &&
+    selectedProductIdFieldId &&
+    integrations.stripe
+  ) {
+    // get the value from the hidden field
+    const fieldVal = fieldValues[selectedProductIdFieldKey] || {};
+    const quantity = fieldVal[productId];
+    let newQuantity = quantity ^ 1;
```
bo-dun-1, line 242: haha this is very clever, but maybe add a comment explaining the XOR?

src/integrations/stripe.ts
```diff
@@ -138,6 +195,92 @@ export async function setupPaymentMethod(
...
+
+    // set the RETURNED value into the hidden field so that BE is source of truth
+    updateFieldValues({
+      [fieldUpdate.field_key]: {
+        [fieldUpdate.value.stripe_product_id]: fieldUpdate.value.quantity
+      }
+    });
+
+    // set the total into the total hidden field if mapped (leaving the currency symbol off)
+    const totalFieldKey =
```
bo-dun-1, line 271: this logic seems like it could be part of the client's return value for updateProductSelection. ideally couldn't we have updateProductSelection returna. dict of field values to update directly?

src/integrations/stripe.ts
ericpias: This gets presented on the ui using a text var.  AFAIK, there is no formatting supported with text var, so need to format it nicely in the hidden field as a monetary format.  Not ideal, I agree.  Let me know if there is another way.

ericpias: All of this logic is moving to the BE so not ex-licit change is needed here.  However, there is a reason to format this as a string.  The designer will be displaying this total using a text var which has not formatting functionality.  This needs to display as a monetary amount.

src/form/grid/index.tsx
ericpias: Just removing it

src/integrations/stripe.ts
ericpias: Moved to BE.

src/integrations/stripe.ts
ericpias: Now a boolean on the BE.

bo-dun-1 (approved):

bo-dun-1 (approved):

PR #560, opened 2022-09-21
handle back button action

bo-dun-1:


ericpias (commented):
I tried to do a bit of testing but I am getting "invalid link value" error when saving a test form.  I did not see a BE branch for this.

obmelvin (approved):

obmelvin:
code looks good to merge once the BE supports saving the action

src/utils/formHelperFunctions.ts
```diff
@@ -495,6 +495,15 @@ function getNewStepUrl(stepKey: any) {
...
 
+function getPrevStepUrl(curStep: any, stepMap: Record<string, string>) {
+  let newStepKey = stepMap[curStep.key];
+  if (!newStepKey) {
+    const prevCondition = curStep.previous_conditions[0];
+    if (prevCondition) newStepKey = prevCondition.previous_step_key;
+  }
+  return newStepKey ? getNewStepUrl(newStepKey) : '';
+}
+
```
ericpias, line 506: There was a bit of behavior related to this that I wasn't expecting.  I set up a form with 3 steps, each with a back and next button.  The last step had an additional button with a NEXT action to the first step.  So basically you could navigate in circles, but the last step also had a submit to end the flow.  Then I tested it and without doing anything else, clicked the back button on the first step and it went to the last step.  I was thinking that this feature worked only with history and I was not expecting this (because I had not been to the last step yet).  Not sure if this boundary case behavior is intended or not.

ericpias (changes requested):
Left a comment/question.

ericpias (changes requested):
"Go to previous step" is not working with a text.

PR #561, opened 2022-10-06
Custom Validations

ericpias:
Custom Validations - Part 1 - Supporting all logic comparisons (field type-specific) but the right side only supports literal values. No field to field comparisons yet.

See PRD : https://docs.google.com/document/d/1i1N8rIHYC2whL3MGq-AP9V-_Npmbo53yk0SM90avlqs/edit#
See: https://www.figma.com/file/rnaRNQXZMQTpYHeAACiSJA/Design-Handoff---Archived-8%2F31%2F22?node-id=3648%3A68636

Test Plan:

- Standard validations
  - Std validations take priority over custom validations
  - Required
  - email
  - phone
- Test custom validation operators
  - Test is_filled, is_empty on each type TextField - > the advanced fields
  - Test is_true/false, on checkbox
  - Test is_numerical, is_text for textfield/textarea
  - Test is_numerical, is_text for other option on checkbox and radio groups
  - Test is_numerical, is_text for pin
  - Test > < >= <= for number
  - Test (not) contains with all applicable control types
  - Test (not) starts-with with all applicable control types
  - Test (not) ends-with with all applicable control types
  - Test is and isn’t (equals and not equals) with all applicable control types
  - Test multiple values
  - Test repeating fields
- Test rule on a button
  - If triggered, cannot submit the step
- Test that a newly set value is properly used in the validation (i.e. does the value get into fieldValues or is it only in formattedFields)
- Test that validation shows on each field if the field loses focus but only after first submit attempt.
- Test that a newly set value triggers the custom validation 
- Test that a value from a previous step is used the validation rule on later step
- Test removing a field that is involved in a rule - what happens?
- Test that auto-validate is only after first submit
  - Test auto-validate is reset on next page
  - Test that auto-validate is reset on going back
  - Test that auto validate works with repeats - change a repeat, add another, etc.
- Test an SDK client calling validateStep
- SDK client - onError
- Test with NextJS

### DIFF
--- a/src/form/Form.tsx
+++ b/src/form/Form.tsx
@@ -1,8 +1,16 @@
 import { BrowserRouter, Route, useHistory } from 'react-router-dom';
-import React, { useEffect, useMemo, useRef, useState } from 'react';
+import React, {
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+  useCallback
+} from 'react';
 
 import ReactForm from 'react-bootstrap/Form';
 import { useHotkeys } from 'react-hotkeys-hook';
+// @ts-expect-error TS(7016): Could not find a declaration file for module 'loda... Remove this comment to see the full error message
+import debounce from 'lodash.debounce';
 
 import { calculateStepCSS, isFill } from '../utils/hydration';
 import {
@@ -11,7 +19,6 @@ import {
   formatStepFields,
   getAllElements,
   getDefaultFieldValue,
-  getFieldError,
   getFieldValue,
   getNewStepUrl,
   getOrigin,
@@ -20,9 +27,9 @@ import {
   nextStepKey,
   recurseProgressDepth,
   setFormElementError,
-  shouldElementHide,
-  validators
+  shouldElementHide
 } from '../utils/formHelperFunctions';
+import { validators, validateElements } from '../utils/validation';
 import {
   initInfo,
   initState,
@@ -142,6 +149,7 @@ function Form({
   const [client, setClient] = useState(null);
   const history = useHistory();
 
+  const [autoValidate, setAutoValidate] = useState(false);
   const [first, setFirst] = useState(true);
   const [firstStep, setFirstStep] = useState(true);
   // If true, will automatically redirect to firstStep if logged back in
@@ -338,6 +346,8 @@ function Form({
   useEffect(() => {
     if (!activeStep) return;
 
+    setAutoValidate(false); // Each step to initially not auto validate
+
     if (formSettings.autofocus && focusRef.current) {
       focusRef.current.focus({
         preventScroll: true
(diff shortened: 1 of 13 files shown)


src/utils/validation.ts
```diff
@@ -0,0 +1,221 @@
...
+      required: boolean;
+    };
+    validations?: ResolvedCustomValidation[];
+  },
+  fieldValues: { [key: string]: any }
+): string {
+  // First priority is standard validations for servar fields
+  const { servar, validations } = element;
+  if (servar) {
+    const errorMsg = getFieldError(fieldValues[servar.key], servar);
```
bo-dun-1, line 106: getStandardFieldError?

src/form/Form.tsx
```diff
@@ -587,28 +608,30 @@ function Form({
 
     initState.validateCallbacks[formKey] = (trigger: any) => {
       const inlineErrors = {};
-      const errors = newStep.servar_fields
+      const errors = [...newStep.servar_fields, ...newStep.buttons]
```
bo-dun-1, line 611: could this instead call validateElements directly? stuff like filtering out hidden elements is already handled by that function.

src/form/Form.tsx
```diff
@@ -472,6 +479,20 @@ function Form({
     Object.assign(fieldValues, newFieldValues);
     // Always rerender from empty state for display purposes
     if (rerender || empty) setRender((render) => !render);
+
+    if (autoValidate) {
+      // validate all step fields and buttons
+      validateElements({
```
bo-dun-1, line 485: what does performance look like here? esp for text fields where validation would now be triggered for each entry

src/form/Form.tsx
ericpias: Refactored validateElements and used it here as well.

src/form/Form.tsx
```diff
@@ -141,6 +144,8 @@ function Form({
   const [client, setClient] = useState(null);
   const history = useHistory();
 
+  // Should the system validate on each field change?
```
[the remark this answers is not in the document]
ericpias: Moved the comment to a more appropriate place.

src/form/Form.tsx
ericpias: Good point.  We do need to revalidate all steps field and buttons on field value updates in order to get the behavior we want.  So, in order to make sure this is as fast as possible, I have implemented a 750ms debounce on the validateElements call in updateFieldValues.

src/form/Form.tsx
```diff
@@ -464,6 +474,29 @@ function Form({
...
+          fieldValues,
+          triggerErrors: true,
+          errorType: formSettings.errorType,
+          formRef,
+          setInlineErrors
+        });
+    }, 750),
+    [activeStep?.id, formRef, validateElements]
+  );
+  React.useEffect(() => {
```
bo-dun-1, line 494: just useEffect instead of React.useEffect

src/form/Form.tsx
```diff
@@ -473,6 +506,12 @@ function Form({
     Object.assign(fieldValues, newFieldValues);
     // Always rerender from empty state for display purposes
     if (rerender || empty) setRender((render) => !render);
+
+    // Should the system validate on each field change?
```
bo-dun-1, line 510: is this question addressed now that we're debouncing validations?

bo-dun-1 (approved):

src/form/Form.tsx
ericpias: This question is just my unorthodox way of commenting on the meaning of the if statement below it.  Since causing a concern/question by it presence, I will change it.

PR #569, opened 2022-10-17
Custom Validations - field to field comparison support

ericpias:
This completes custom validations by adding support for field to field comparisons. This means that the right side of rule comparison expression can now have a mix of free-form values and fields.

See:

https://www.figma.com/file/rnaRNQXZMQTpYHeAACiSJA/Design-Handoff---Archived-8%2F31%2F22?node-id=5282%3A97229

- Test Plan:
- Test field to field comparison on field
- Test field to field comparison on button
- Test single valued field comparison with a multi-valued field (checkbox group)
- Test multi-valued field comparison with a multi-valued field (checkbox groups)
- Test single valued field comparison with a multi-valued field in a REPEAT

### DIFF
--- a/src/utils/logic.ts
+++ b/src/utils/logic.ts
@@ -18,12 +18,19 @@ type OPERATOR_CODE =
   | 'is_numerical'
   | 'is_text';
 
+export type FieldValueType = {
+  field_type: 'servar' | 'hidden';
+  field_id: string;
+  field_key: string;
+};
+export type ValueType = string | FieldValueType;
+
 export interface ComparisonRule {
   field_type?: '' | 'servar' | 'hidden';
   hidden_field?: string | null;
   servar?: string | null;
   comparison?: OPERATOR_CODE; // always present after the initial state
-  values?: any[];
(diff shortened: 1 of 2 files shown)


src/utils/logic.ts
```diff
@@ -50,25 +62,28 @@ const evalComparisonRule = (
...
-  return (
-    Array.isArray(fieldValue)
-      ? fieldValue.length
-        ? fieldValue
-        : [undefined]
-      : [fieldValue]
-  )
+  // flatten the right side values/fields into flat list of values
+  const flatValues = rule.values.flatMap((value) => {
+    if (value !== null && valueTypeIsField(value))
```
bo-dun-1, line 67: if (value?.field_key)

maybe woudl be simpler?

bo-dun-1 (approved):

src/utils/logic.ts
ericpias: Unfortunately, no.  Typescript needs narrowing here and the best way is the type predicate with the typeof guard.

PR #571, opened 2022-10-17
international phone number support

bo-dun-1:


src/utils/validation.ts
```diff
@@ -132,14 +133,22 @@ const LIB_PHONE_NUMBER_URL =
...
+  email: (a: string) => emailPattern.test(a),
+  phone: (a: string) => {
     try {
       return global.libphonenumber.isValidPhoneNumber(a, 'US');
     } catch (e) {
       // Invalid phone number
       return false;
     }
+  },
+  internationalPhone: (a: string) => {
```
ericpias, line 145: Could we combine phone and international phone into one validator and just try both formats?  It seems 'phone' is only now used if set in 'login' metadata as a method (?) but it should work there for a common validation.

src/utils/validation.ts
```diff
@@ -124,6 +124,7 @@ function validateElement(
 const emailPatternStr =
   "^[a-zA-Z0-9.!#$%&'*+\\/=?^_`{|}~-]+@[a-zA-Z0-9-]+(?:\\.[a-zA-Z0-9-]+)+$";
 const emailPattern = new RegExp(emailPatternStr);
+// TODO: deprecate and support international format
 const phonePattern = /^\d{10}$/;
```
ericpias, line 128: Can we just get rid of this now and replace with the validation method below.  It seems like firebase and the is_type comparison would behave roughly the same using it instead?

src/elements/fields/PhoneField/index.tsx
```diff
@@ -0,0 +1,278 @@
...
+          target={triggerRef.current}
+          show={show}
+          onHide={() => setShow(false)}
+          placement='bottom-start'
+        >
+          {(props) => {
+            ['placement', 'arrowProps', 'show', 'popper'].forEach(
+              (prop) => delete props[prop]
+            );
+            return (
```
ericpias, line 183: It would be really nice if this list had a search/filter on it.  It was slightly tedious to scroll looking for a country.  The filter could accept partial names or partial country codes.

ericpias (changes requested):

src/utils/formHelperFunctions.ts
```diff
@@ -379,8 +379,6 @@ function textFieldShouldSubmit(servar: any, value: any) {
       methods = servar.metadata.login_methods;
       onlyPhone = methods.length === 1 && methods[0] === 'phone';
       return onlyPhone && value.length === 10;
-    case 'phone_number':
-      return value.length === 10;
```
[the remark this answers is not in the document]
bo-dun-1: autosubmit gets triggered once the field gets completely filled out, but since phone numbers can have vairable lengths now, i'm not sure we can make any assumptions about completion

src/utils/validation.ts
bo-dun-1: think @obmelvin discussed this with you offline, we'll do the login refactor in a separate change

src/utils/validation.ts
bo-dun-1: will be refactored later

src/elements/fields/PhoneField/index.tsx
```diff
@@ -0,0 +1,324 @@
...
+                fontSize: '14px',
+                '&:hover': {
+                  backgroundColor: '#e6e6e6'
+                }
+              }}
+              onClick={() => {
+                setCurCountryCode(countryCode);
+                setRawNumber('');
+                setFormattedNumber('');
+                setShow(false);
```
ericpias, line 152: After changing countries, it would be good to put the focus on the input field for immediate entry of a number.

src/elements/fields/PhoneField/index.tsx
```diff
@@ -0,0 +1,324 @@
...
+          placeholder=''
+          value={formattedNumber}
+          ref={(ref) => {
+            inputRef.current = ref;
+            setRef(ref);
+          }}
+          type='tel'
+          onBlur={() => setTriggerOnChange(!triggerOnChange)}
+          onKeyDown={(e) =>
+            e.key === 'Enter' && setTriggerOnChange(!triggerOnChange)
```
ericpias, line 271: I think it would be good UX to have it work this way: if focus is the phone number input, then if the user types + or any letter, the overlay opens and the input is in the search/filter  area.  The user uses up/down arrow to make a country choice, hits enter and then focus is on the input field to accept numeric input.  Having to go between mouse and keyboard is clunky.

ericpias (changes requested):

src/utils/formHelperFunctions.ts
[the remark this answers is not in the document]
ericpias: Ok, if autosubmit is not supported for phone, then we should remove it from the dashboard.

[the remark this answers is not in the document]
bo-dun-1: removed

ericpias (approved):

PR #574, opened 2022-10-18
fix nav rules after auth change

obmelvin:


src/form/Form.tsx
```diff
@@ -770,8 +774,9 @@ function Form({
   }, [steps]);
 
   useEffect(() => {
+    // Is it ok to run this again when stepKey is truthy and authId is updated?
```
obmelvin, line 777: question in comment

src/form/Form.tsx
```diff
@@ -256,6 +257,9 @@ function Form({
     initState.renderCallbacks[formKey] = () => {
       setRender((render) => !render);
     };
+    // Need to initialize setAuthId somehow
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const test = new Client(null, null, setAuthId);
```
obmelvin, line 262: thinking i should probably export an object from Client that contains setAuthId and then I can set it here, so that I don't need to have a contrived constructor call

doesn't seem appropriate to put into initState semantically, though it would fit the global access pattern

src/form/Form.tsx
```diff
@@ -149,6 +149,7 @@ function Form({
 }: Props) {
   const formKey = _formName ?? _formKey;
   const [client, setClient] = useState(null);
+  const [authId, setAuthId] = useState<typeof initState.authId>();
```
obmelvin, line 152: I need to create state because initState.authId is always undefined in the `getNewStep` while loop. Need to force re-evaluation of the rules once authId has been set

src/utils/client.ts
```diff
@@ -328,6 +332,7 @@ export default class Client {
   submitAuthInfo({ authId, authPhone = '', authEmail = '' }: any) {
     const { userKey } = initInfo();
     initState.authId = authId;
+    if (setAuthId) setAuthId(authId);
```
obmelvin, line 335: we have to wait for the stytch or firebase call to finish and authId be returned - so there is still a noticeable lag between the page loading and navigating to the correct step

obmelvin: I don't think we can optimistically make an update - because the pure presence of stytch tokens in URl etc doesn't mean they are correct & that the auth call will succeed. So, we'd run the risk of letting a user into an auth gate section when their auth fails

I think I may need to detect if auth is going to happen, and delay rendering the entire form so we can go straight to the correct step

src/elements/components/Spinner.tsx
```diff
@@ -6,7 +6,7 @@ function FeatherySpinner() {
     <Spinner
       animation='border'
       style={{
-        color: 'white',
+        color: '#bdc2ce', // gray-50
```
obmelvin, line 9: the <form> html native element hasn't rendered by the time this has rendered for the auth loading...I tried a few things but ended up just making it grey so it is visible in most situations

src/form/Form.tsx
```diff
@@ -779,8 +805,9 @@ function Form({
   }, [steps]);
 
   useEffect(() => {
-    if (stepKey) getNewStep(stepKey);
-  }, [stepKey]);
+    if (stepKey && (prevStepKey !== stepKey || prevAuthId !== initState.authId))
```
bo-dun-1, line 808: why do you need to do these checks? doesn't this function being triggered imply one of these is different?

src/elements/components/Spinner.tsx
bo-dun-1: how does that look visually? if you need UI suggestions gabriela could probably help here too

obmelvin: image

I talked to Eric a bit about it, and doesn't seem like there's a reliable way to get the background color under the spinner without needing to traverse up the Cells to find which cell (if any) is responsible for the visible background color

Also the `<form>`* html native element hasn't rendered before the spinner is used while waiting on auth, so that would also require some additional work because the spinner is set in the mount/unmount useEffect

* what I was trying to say in my first message - didn't realize github parsed the form tag I had written out

bo-dun-1: another way is you could do like a black border and white background so it's visible everywhere, but not sure if that looks good

PR #588, opened 2022-11-02
deprecate login field, use email and phone instead

obmelvin:


src/utils/validation.ts
```diff
@@ -125,14 +123,7 @@ const loadPhoneValidator = () =>
...
-  phone: (a: string) => {
-    try {
-      return global.libphonenumber.isValidPhoneNumber(a, 'US');
-    } catch (e) {
-      // Invalid phone number
-      return false;
-    }
-  },
+  // should this be changed to just phone, now that it is the only phone validator?
   internationalPhone: (a: string) => {
```
obmelvin, line 127: change this to `phone`  since it is now the only used phone validator?

src/utils/client.ts
```diff
@@ -31,7 +31,12 @@ const CDN_URL_OPTIONS = {
   production: 'https://cdn.feathery.io/api/'
 };
 
-const environment = 'production';
+// ************************************************************
+// ************************************************************
+// const environment = 'production';
+const environment = 'local';
```
obmelvin, line 37: need to change this back

src/form/Form.tsx
```diff
@@ -346,23 +343,17 @@ function Form({
...
       (b: any) =>
-        b.properties.link === LINK_NEXT && !b.properties.link_no_submit
+        b.properties.link === LINK_SEND_SMS && !b.properties.link_no_submit
+    );
+    const smsButton = renderedButtons.find(
+      (b: any) => b.properties.link === LINK_SEND_SMS
     );
-    if (hasLoginField && submitButton) {
+
+    if ((hasPhoneField && submitButton) || smsButton) {
```
obmelvin, line 352: I tried to keep this conditionals equivalent to the old code, but I think I can just check for `smsButton`?

src/elements/fields/PhoneField/countryData.ts
```diff
@@ -1470,3 +1470,64 @@ export default [
     countryName: '\u00c5land Islands'
   }
 ];
+
+// This list was generated by assessing the differences between two sets
+// Set A consists of countries that Firebase indicates should receive SMS verification without issue (https://firebase.google.com/support/faq/#phone-auth-countries)
+// Set B consists of countries listed above, that are used for libphonenumber-js
+// The list below is the set difference B - A, so those for which libphonenumber-js can parse, but for which Firebase isn't confident in the ability to deliver an SMS
+// The CountryDropdown component also has a safeguard to filter out any countries that it does not support
+export const countryCodesNotWellSupportedByFirebase = [
```
bo-dun-1, line 1479: would it be better to make this a set and maybe just the country codes supported by firebase? rather than the converse? no need to do an intersection with libphonenumber here since the set can be directly checked in the phone number filtering logic.

also maybe firebaseSMSCountries for brevity?

src/form/Form.tsx
bo-dun-1: yes you can just check for smsButton

src/form/Form.tsx
```diff
@@ -1345,7 +1336,15 @@ function Form({
       const email = fieldValues[fieldKey] as string;
       if (validators.email(email)) {
         clickPromise = setButtonLoader(button)
-          .then(() => sendMagicLink({ fieldVal: email }))
+          .then(() => {
+            if (getStytchJwt()) return sendMagicLink({ fieldVal: email });
```
bo-dun-1, line 1340: can we make a method that calls maybe !!getStytchJwt() and the method name is 

isAuthStytch? or something of that nature to be more clear

src/form/Form.tsx
```diff
@@ -1345,7 +1336,15 @@ function Form({
...
+            else
+              return sendLoginCode({
```
bo-dun-1, line 1342: maybe rename to sendFirebaseCode to distinguish from stytch?

src/integrations/firebase.ts
```diff
@@ -58,9 +58,23 @@ export function emailLogin(featheryClient: any) {
...
+  method: 'phone' | 'email';
+}) {
+  const errorPayload = {
+    errorMessage: 'Invalid login',
+    errorField: servar
+  };
+
+  if (typeof fieldVal !== 'string') {
+    return errorPayload;
+  } else if (method === 'phone' && validators.internationalPhone(fieldVal)) {
```
bo-dun-1, line 77: should we surface an error message if the user attempts to trigger login and it fails the validator

src/form/Form.tsx
```diff
@@ -1336,7 +1327,7 @@ function Form({
           sendLoginCode({
```
bo-dun-1, line 1327: don't we also need a toggle to allow the user to optionally configure the ability to also submit the step and navigate to the next one in addition to sending the sms code or magic link?

src/integrations/firebase.ts
bo-dun-1: same for the email validator below

src/utils/validation.ts
bo-dun-1: yep agreed

src/form/Form.tsx
obmelvin: yes ty, I have that in the dashboard but missed it here - updated

src/integrations/firebase.ts
obmelvin: We were already checking if the email was valid in Form.tsx and displaying an error if it wasn't, so I went ahead and did the same for phone as well. Then I just removed the validation checks from here - we didn't seem to be doing anything in the original code with the error object returned from here anyway

let me know what you think of the change

src/form/Form.tsx
```diff
@@ -1078,6 +1065,12 @@ function Form({
             (!actionConfig.isMatch ||
               (actionConfig.isMatch && actionConfig.isMatch(actionData)))
           ) {
+            // We only want to run firebase submit logic if it is happening via field auto-submit
+            if (
```
bo-dun-1, line 1069: can we fully remove auth-related logic from the handleActions path? this function is only called as part of the submission path. We've already been pulling auth logic out into a pre-submission path, and IMO we should double down on that

src/form/Form.tsx
```diff
@@ -1278,7 +1271,21 @@ function Form({
     buttonClicks[button.id] = true;
     let clickPromise = Promise.resolve();
 
-    const link = button.properties.link;
+    const determineNavAndSubmit = () => {
```
bo-dun-1, line 1274: authNavAndSubmit?

src/form/Form.tsx
```diff
@@ -1443,16 +1437,25 @@ function Form({
       };
       if (submitData) {
         const submitButton = activeStep.buttons.find(findSubmitButton);
+        const authButton = activeStep.buttons.find(findAuthButton);
```
bo-dun-1, line 1440: let target = activeStep.buttons.find(findSubmitButton);
if (!target) target = activeStep.buttons.find(findAuthButton);

if (target && ...) buttonOnClick(target);

src/utils/formHelperFunctions.ts
```diff
@@ -86,7 +86,11 @@ const getAllElements = (step: any) => {
 const lookUpTrigger = (step: any, elementID: any, elementType: any) => {
   if (elementType === 'button') {
     const element = step.buttons.find((button: any) => button.id === elementID);
-    return { id: elementID, text: element.properties.text };
+    return {
+      id: elementID,
+      text: element.properties.text,
+      link: element.properties.link
```
bo-dun-1, line 92: why is link being returned here?

src/form/Form.tsx
bo-dun-1: yeah actually findAuthButton logic can be put into findSubmitButton

src/form/Form.tsx
```diff
@@ -1191,16 +1157,31 @@ function Form({
...
       const metadata = {
         elementType: 'button',
         elementIDs: [button.id]
       };
       if (submitData) {
+        // ************************************************
+        // ************************************************
+        // ************************************************
+        // ************************************************
+        // what about the case where we are going to nav, but not submit?
```
obmelvin, line 1175: if the button is configured like this
image

should we execute buttonAction? This code does *not* execute the btn action in this case. I think it probably should, which is overall an easy change

in this case, since validation is turned off - it should just attempt to execute buttonAction without setting any errors because it the expected bx is to transition to the next step no matter what?

bo-dun-1 (approved):

src/utils/formHelperFunctions.ts
obmelvin: needed to know the link type for something before, but its no longer needed - removed

PR #591, opened 2022-11-08
Fixes sentry error - repeatIndex issue

ericpias:
This is a sentry reported issue: https://sentry.io/organizations/feathery-forms/issues/3726710077/?project=6078176&referrer=slack

I can replicate using /to/gjK7ps#Listing%20Information on prod.

It seems to be just bad code here, but not 100% sure.  Drew should definitely look this over.


ericpias:
Fixed by https://github.com/feathery-org/feathery-react/pull/592

PR #601, opened 2022-11-11
make stytch check more robust

obmelvin:


astrozzc (approved):
Oliver tried looking at the stytch client object to see how to identify the object from firebase. it has this JS symbol on it that is called stytch__internal, which is a clear way to tell it apart

bo-dun-1 (approved):

PR #604, opened 2022-11-14
Hide/Nav Rule Refactor to use the new custom validation core logic engine

ericpias:
Hide rule and nav rules now support all of the new operators that custom validation supports.

Test Plan

- Hide If
  - Test multi valued rule
  - Test field-to-field comparison rule
  - Test hide rule on repeat that the indexing works as it did before
- Nav
  - Test various conditional nav rules
- Do some regression testing on validations including the sdk method validateStep()

### DIFF
--- a/src/form/Form.tsx
+++ b/src/form/Form.tsx
@@ -26,9 +26,9 @@ import {
   lookUpTrigger,
   nextStepKey,
   recurseProgressDepth,
-  setFormElementError,
-  shouldElementHide
+  setFormElementError
 } from '../utils/formHelperFunctions';
+import { shouldElementHide, getHideIfReferences } from '../utils/hideIfs';
 import { validators, validateElements } from '../utils/validation';
 import {
   initInfo,
@@ -336,7 +336,6 @@ function Form({
     const renderedButtons = activeStep.buttons.filter(
       (element: any) =>
         !shouldElementHide({
-          fields: activeStep.servar_fields,
           element: element
         })
     );
@@ -376,6 +375,13 @@ function Form({
         (f: any) => f.servar.type === 'gmap_line_1' && fieldValues[f.servar.key]
       )
     );
+  }, [activeStep?.id]);
+
+  // Figure out which fields are used in hide rules so that observed changes can be used
+  // to trigger rerenders
+  const hideIfFieldReferences = useMemo(() => {
+    if (activeStep) return getHideIfReferences(getAllElements(activeStep));
+    return new Set<string>();
   }, [activeStep?.id]);
 
   const scrollToRef = (ref: any) =>
@@ -474,7 +480,6 @@ function Form({
       activeStep &&
         validateElements({
           elements: [...activeStep.servar_fields, ...activeStep.buttons],
-          servars: activeStep.servar_fields,
           triggerErrors: true,
           errorType: formSettings.errorType,
           formRef,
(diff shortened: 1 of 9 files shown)


src/utils/hide.ts
```diff
@@ -0,0 +1,65 @@
+import { evalComparisonRule, ResolvedComparisonRule } from './logic';
```
bo-dun-1, line 1: should this file be hideIfs.ts?

src/utils/formHelperFunctions.ts
```diff
@@ -98,7 +99,7 @@ const lookUpTrigger = (step: any, elementID: any, elementType: any) => {
   } else return {};
 };
 
-const nextStepKey = (nextConditions: any, metadata: any) => {
+const nextStepKey = (nextConditions: any, metadata: any, fieldValues: any) => {
```
bo-dun-1, line 102: fieldValues is globally available so it doens't have to be passed in, just import it into the file

src/form/Form.tsx
```diff
@@ -484,22 +498,40 @@ function Form({
...
+      debouncedHideRuleCheck.cancel();
     };
-  }, [debouncedValidate]);
+  }, [debouncedValidate, debouncedHideRuleCheck]);
 
   const updateFieldValues = (newFieldValues: any, rerender = true) => {
     const entries = Object.entries(newFieldValues);
-    const noChange = entries.every(([key, val]) => fieldValues[key] === val);
-    if (noChange) return false;
+    const changedEntries = entries.filter(
```
bo-dun-1, line 522: use some instead of filter, it'll return a boolean directly

src/form/Form.tsx
```diff
@@ -484,22 +498,40 @@ function Form({
...
+    );
+    if (!changedEntries.length) return false;
 
     const empty = entries.some(([key, val]) => !val || !fieldValues[key]);
     Object.assign(fieldValues, newFieldValues);
+
     // Always rerender from empty state for display purposes
     if (rerender || empty) setRender((render) => !render);
 
+    debouncedHideRuleCheck(changedEntries);
```
bo-dun-1, line 533: how does this affect performance? especially while typing in text fields on large steps

bo-dun-1: also wouldn't this debouncing potentially prevent renreders that need to happen?

src/form/Form.tsx
```diff
@@ -379,6 +379,21 @@ function Form({
     );
   }, [activeStep?.id]);
 
+  // Figure out which fields are used in hide rules so that observed changes can be used
+  // to trigger rerenders
+  const hideIfFieldReferences = useMemo(() => {
+    if (activeStep)
+      return getHideIfReferences([
```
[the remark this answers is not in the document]
ericpias: I did not find getStepElements or applyToStepElements which seem to be dashboard only.  I did find getAllElements in formHelperFunctions and used that instead.

src/form/Form.tsx
```diff
@@ -484,22 +490,44 @@ function Form({
...
 
     const empty = entries.some(([key, val]) => !val || !fieldValues[key]);
     Object.assign(fieldValues, newFieldValues);
+
     // Always rerender from empty state for display purposes
     if (rerender || empty) setRender((render) => !render);
 
+    // If any fields involved in a hideIf have changed, then trigger a re-render to re-evaluate the hide rule
+    if (
+      !entries.some(
```
bo-dun-1, line 524: doing !entries.some is equivalent to doing entries.every on the opposite check, which may read clearer

src/form/grid/Cell.tsx
```diff
@@ -63,8 +63,8 @@ const Cell = ({ node: el, form }: any) => {
 
   if (
     shouldElementHide({
-      fields: activeStep.servar_fields,
-      element: el
+      element: el,
+      fieldValues
```
bo-dun-1, line 67: fieldValues is global and does not need to be passed as a parameter into functions

src/form/Form.tsx
```diff
@@ -484,22 +490,44 @@ function Form({
     }, 750),
     [activeStep?.id, formRef, validateElements]
   );
+
+  // Debouncing the re-render due to changing values of referenced fields in hide if rules.
+  // Need to rate limit re-renders here for performance reasons.
+  const debouncedHideRuleRender = useCallback(
```
bo-dun-1, line 496: should we just use this debounced function for all rerenders? then this doesn't have to be hide rule specific and we can unify some logic in updateFieldValues

src/form/grid/Cell.tsx
obmelvin: I get your point yesterday @ericpias regarding it being easy to pass in for testing, but we already need to handle it being global for all the other code so it is easiest to be consistent - rather than need to pass in the value or mock depending on what is being tested

src/form/Form.tsx
```diff
@@ -483,21 +488,43 @@ function Form({
...
+      rerender ||
+      empty ||
+      entries.some(
+        ([key, val]) =>
+          fieldValues[key] !== val && hideIfFieldReferences.has(key)
+      )
+    )
+      debouncedRerender();
+
+    Object.assign(fieldValues, newFieldValues);
```
bo-dun-1, line 527: i think we probably need to assign field values and *then* rerender

bo-dun-1 (approved):

PR #616, opened 2022-11-23
add save hidden field click action

obmelvin:


bo-dun-1 (changes requested):
we also need the button to appear selected when the value matches the value it's trying to store if the link property is STORE_FIELD_VALUE

src/form/Form.tsx
```diff
@@ -1348,6 +1349,10 @@ function Form({
         client,
         integrations
       });
+    } else if (link === LINK_SAVE_HIDDEN_FIELD) {
+      const { hidden_field_id: id, hidden_field_value: value } =
+        button.properties;
+      updateFieldValues({ [id]: value });
```
bo-dun-1, line 1355: in the submit flow, you'll need to also look through buttons for fields like these whose values might have been updated that we want to send back to the backend. you can send hidden field values via submitCustom

src/form/Form.tsx
```diff
@@ -999,16 +1019,18 @@ function Form({
...
         metadata,
         formattedFields
       });
     } else {
-      return submitAndNavigate({
-        metadata,
-        formattedFields
-      });
+      submitHidden();
+      // return submitAndNavigate({
```
bo-dun-1, line 1030: please remove commented out code or uncomment

bo-dun-1: this kinda seems like it should maybe just be int he submitAndNavigate function if both functions need to be c alled in both places

src/form/Form.tsx
```diff
@@ -1348,6 +1372,12 @@ function Form({
         client,
         integrations
       });
+    } else if (link === LINK_STORE_FIELD) {
+      const { hidden_field_id: id, hidden_field_value: value } =
```
bo-dun-1, line 1376: these property keys dont seem accurate?

src/form/Form.tsx
```diff
@@ -937,6 +938,25 @@ function Form({
       return;
     }
 
+    const submitHidden = () => {
+      const checkLink =
```
bo-dun-1, line 942: instead of checkLInk maybe call it `isStoreFieldValueAction`?

src/form/Form.tsx
```diff
@@ -937,6 +938,25 @@ function Form({
       return;
     }
 
+    const submitHidden = () => {
```
bo-dun-1, line 941: submitStepHiddenFields?

bo-dun-1 (approved):

PR #638, opened 2022-12-07
change how auth nav rules are configured

obmelvin:


src/Form/index.tsx
```diff
@@ -593,27 +593,34 @@ function Form({
...
+        !window.firebaseConfirmationResult &&
+        nextStepIsProtected
+      )
+        nextStep = findStep(metadata.logout_step);
+
+      if (
+        nextStep !== undefined &&
+        changeStep(nextStep, newKey, steps, history)
+      )
+        clearLoaders(true);
```
bo-dun-1, line 623: does there need to be a return statement after clearLOaders?

src/Form/index.tsx
```diff
@@ -593,27 +593,34 @@ function Form({
...
+      let nextStep: undefined | string;
+
+      if (userAuthed && !nextStepIsProtected)
+        nextStep = findStep(metadata.login_step);
+      else if (
+        !userAuthed &&
+        // Re: firebaseConfirmationResult, the user hasn't authenticated yet
+        // but they're in the process of doing so and we don't want to
+        // consider them "unauthenticated" for the purposes of redirecting
+        !window.firebaseConfirmationResult &&
```
bo-dun-1, line 614: i think we can remove this tbh, it's a weird case that probably doesn't need to be supported

src/Form/index.tsx
```diff
@@ -183,7 +183,7 @@ function Form({
   const [inlineErrors, setInlineErrors] = useState({});
   const [, setRepeatChanged] = useState(false);
 
-  const [integrations, setIntegrations] = useState({});
+  const [integrations, setIntegrations] = useState<any>({});
```
bo-dun-1, line 186: Record<string, any>?

src/Form/index.tsx
obmelvin: yes, ty good catch, added that back in

src/Form/index.tsx
```diff
@@ -574,27 +574,30 @@ function Form({
...
+    if (authSteps.length > 0) {
+      const userAuthed = Boolean(initInfo().authId);
+      const nextStepIsProtected = authSteps.includes(newStep.id);
+      const findStep = (stepId: string) =>
+        // @ts-expect-error TS(2531): Object is possibly 'null'.
+        Object.values(steps).find((step) => step.id === stepId).key;
+      let nextStep: undefined | string;
+
+      if (userAuthed && !nextStepIsProtected)
+        nextStep = findStep(metadata.login_step);
```
bo-dun-1, line 590: if login_step or logout_step happent o be invalid, empty string, etc. then the form will hard break here because 

Object.values(steps).find((step) => step.id === stepId).key;

will attempt to access the key prop on null. think we should check here to be robust, shoudl also fix the typescript error

bo-dun-1: also i'm not sure we need this branch of the if statement. the user can be logged in and access a logged out step, the inverse is the only thing that we need to prevent.

src/Form/index.tsx
```diff
@@ -576,27 +576,34 @@ function Form({
...
+      const nextStepIsProtected = authSteps.includes(newStep.id);
+      const findStepName = (stepId: string) => {
+        const step: undefined | { key: string } = Object.values(steps).find(
+          (step: any) => step.id === stepId
+        ) as undefined | { key: string };
+        return step?.key;
+      };
+      let nextStep: undefined | string;
+
+      if (userAuthed && initInfo().loginRedirectFlag) {
```
obmelvin, line 594: we seem to use initInfo().x to read and initState.x = y to write
is this the desired distinction? if so, should initInfo() return a spread initState so values can't be set?

bo-dun-1: that is the desired distinction, but spreading isn't a perfect solve since it's a shallow copy

src/utils/client.ts
```diff
@@ -347,6 +347,7 @@ export default class Client {
   submitAuthInfo({ authId, authPhone = '', authEmail = '' }: any) {
     const { userId } = initInfo();
     initState.authId = authId;
+    initState.loginRedirectFlag = true;
```
bo-dun-1, line 350: are there any race conditions here

obmelvin: the renderCb below handles the race condition between authentication and form navigation

bo-dun-1 (approved):

PR #640, opened 2022-12-08
fix PIN raw value

obmelvin:
creates helper function so rawValue is set correctly in all locations
also fixes the check in handleMultipleValues that would only evaluate the first item in the array

### DIFF
--- a/src/elements/fields/PinInputField.tsx
+++ b/src/elements/fields/PinInputField.tsx
@@ -106,6 +106,8 @@ function SingleOtpInput({
   );
 }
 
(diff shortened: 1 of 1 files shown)


bo-dun-1 (approved):

src/elements/fields/PinInputField.tsx
```diff
@@ -114,12 +114,14 @@ function OtpInput({
   value,
   inlineError
 }: any) {
+  const convertValueToRaw = (value: any) => value.toString().split('');
```
bo-dun-1, line 117: this can probably be pulled out into a function outside of the component

PR #645, opened 2022-12-11
decouple option value and label

bo-dun-1:


src/utils/formHelperFunctions.ts
```diff
@@ -76,6 +76,17 @@ function getDefaultFieldValue(field: any) {
   }
 }
 
+function updateStepFieldOptions(step: any, newFieldOptions: any) {
```
ericpias, line 79: newFieldOptions: Record<string, string[]> ?

src/utils/formHelperFunctions.ts
```diff
@@ -76,6 +76,17 @@ function getDefaultFieldValue(field: any) {
...
+  step.servar_fields.forEach((field: any) => {
+    const servar = field.servar;
+    if (servar.key in newFieldOptions) {
+      const options = newFieldOptions[servar.key];
+      servar.metadata.options = options;
+      servar.metadata.option_labels = options;
```
ericpias, line 85: Should we support setting option labels from the SDK as well?

ericpias (changes requested):
Also, I was expecting that option label would be supported by the dashboard.  I suppose another PR is coming for that?

src/utils/formHelperFunctions.ts
bo-dun-1: added

PR #666, opened 2022-12-26
move stytch code to ui lib

obmelvin:


src/hooks/useAuth.ts
```diff
@@ -0,0 +1,90 @@
...
+      if (hasAuthed) return;
+      // This onChange is just meant to set the token on initial login. It is
+      // not to cache the token. The token is retrieved directly every time it
+      // is needed by the function getStytchJwt
+      const newToken =
+        newSession?.stytch_session?.session_jwt ?? getStytchJwt();
+      if (!newToken) return;
+      setHasAuthed(true);
+      // cb to fetch resources
+      onAuth();
```
obmelvin, line 41: I should try to listen here to check if the client revokes the session themselves. if so, I need to call onLogout

src/hooks/useAuth.ts
```diff
@@ -0,0 +1,90 @@
...
+import { getAuthClient } from '../utils/init';
+/** These next 2 should maybe be dynamically imported, but having trouble with that */
+import { useIdleTimer } from 'react-idle-timer';
+import throttle from 'lodash.throttle';
+
+const TEN_SECONDS_IN_MILLISECONDS = 1000 * 10;
+const FIVE_MINUTES_IN_MILLISECONDS = 1000 * 60 * 5;
+
+export const useAuth = ({
+  onAuth = () => {},
```
obmelvin, line 13: onLogin for consistency

src/hooks/useAuth.ts
```diff
@@ -0,0 +1,90 @@
...
+  onLogout = () => {
```
obmelvin, line 14: this default fn needs to call `setHasAuthed`

src/integrations/utils.ts
```diff
@@ -47,20 +47,46 @@ export async function initializeIntegrations(
   integs: any,
   featheryClient: Client
 ) {
+  // const isAuthIntegActive = Boolean(integs.stytch || integs.firebase);
   await Promise.all([
     installPlaid(!!integs.plaid),
     installFirebase(integs.fb),
     installStytch(integs.stytch),
```
obmelvin, line 54: remove this from the general install. these auth integs should only be installed via the useAuth hook's preload

src/hooks/useAuth.ts
```diff
@@ -0,0 +1,90 @@
...
+import { getStytchJwt } from '../utils/browser';
+import { getAuthClient } from '../utils/init';
+/** These next 2 should maybe be dynamically imported, but having trouble with that */
+import { useIdleTimer } from 'react-idle-timer';
+import throttle from 'lodash.throttle';
+
+const TEN_SECONDS_IN_MILLISECONDS = 1000 * 10;
+const FIVE_MINUTES_IN_MILLISECONDS = 1000 * 60 * 5;
+
+export const useAuth = ({
```
obmelvin, line 12: pass in a single formName to preload (use existing preload call)

if useAuth called with multiple formNames we should raise an exception

src/utils/init.ts
```diff
@@ -35,13 +35,11 @@ type InitOptions = {
...
   preloadForms?: string[];
   userTracking?: 'cookie' | 'fingerprint';
-  authId?: string;
-  authEmail?: string;
-  authPhoneNumber?: string;
   language?: string;
 } & DeprecatedOptions;
 
 type InitState = {
+  authId?: string;
```
obmelvin, line 42: I moved authId out of the init() fn call and into the auth gate like we talked about. but otherwise left the state itself here

since doing that I've made the authState in the auth gate, so lmk if you think it should stay or be moved there

src/utils/client.ts
```diff
@@ -353,18 +354,16 @@ export default class Client {
     return [trueSession, formData];
   }
 
-  submitAuthInfo({ authId, authPhone = '', authEmail = '' }: any) {
+  async submitAuthInfo({ authId, isStytchTemplateKey }: any) {
```
obmelvin, line 357: I realized that we were passing in `isStytchTemplateKey`, but it wasn't actually being used by this fn..

src/elements/components/FeatheryAuthGate.tsx
```diff
@@ -0,0 +1,190 @@
...
+    onActive,
+    timeout: TEN_SECONDS_IN_MILLISECONDS
+  });
+
+  const form = (
+    <>
+      {showLoader && (
+        <div
+          style={{
+            // TODO: what should the background color be?
```
obmelvin, line 161: do you think we should provide a component prop for the loader background color? or just leave it as white until there is a demand for customizing the loader completely?

obmelvin: will delete commented line below. left so you can see the old value since the code was moved

src/Form/index.tsx
```diff
@@ -1065,13 +1066,27 @@ function Form({
...
+        const isTerminalStepAuth =
+          authState.sentAuth &&
+          authIntegration?.auth_gate_steps.length &&
+          !authIntegration?.auth_gate_steps.includes(steps[stepKey].id);
+        // Need to guard against setting completed back to false actually this
+        // doesn't work properly, because if you create multiple accounts with
+        // the same fuser ID, the activation storm still thinks it's complete
+        // after logging out. so probably just need this to stay as true like we
+        // discussed (tested more and it actually can be set back to false)
+        // eventData.completed = formCompleted || !isTerminalStepAuth;
```
obmelvin, line 1087: Sorry, I realized I was wrong and actually I confirmed that I can overwrite the completed flag from true to false. 

I tried to guard against setting completed back to false, but actually it doesn't work properly. If you create multiple accounts with the same fuser ID, the activation form still thinks it's complete after logging out

so probably just need this to stay true like we discussed

src/elements/components/FeatheryAuthGate.tsx
```diff
@@ -0,0 +1,178 @@
...
+      setShowLoader(true);
+    }
+
+    const { location, history } = window;
+    if (whitelistPath && location.pathname !== whitelistPath) {
+      // If user is not at the URL whitelisted for auth, take them there for login
+      history.replaceState(null, '', whitelistPath);
+    }
+
+    authState.clearLoader = () => setShowLoader(false);
```
bo-dun-1, line 82: do we need this callback function? can't we just hide loader when authState.onLogin is called? aka

authState.onLogin = () => {setShowLoader(false); onLogin()}

src/elements/components/FeatheryAuthGate.tsx
```diff
@@ -0,0 +1,178 @@
...
+    // Register onLogin cb so it can be called by Client.submitAuthInfo
+    authState.onLogin = onLogin;
+    authState.onLogout = onLogout;
+
+    // If user passes authId as a prop, we need to submit it
+    if (authIdProp) {
+      const defaultClient = new Client();
```
bo-dun-1, line 89: can we just use the defaultClient from initState?

src/elements/components/FeatheryAuthGate.tsx
bo-dun-1: i think it's fine to omit the component prop until it's needed. If this is the same loader component code as in the form, could we unify into a separate loader component? That maybe takes in a show boolean prop

src/elements/components/FeatheryAuthGate.tsx
```diff
@@ -0,0 +1,178 @@
...
+          <div style={{ height: '10vh', width: '10vh' }}>
+            <Spinner />
+          </div>
+        </div>
+      )}
+      <Form {...formProps} />
+    </>
+  );
+  if (!initInfo().authId || !formCompleted) {
+    return !fullPageLogin ? (
```
bo-dun-1, line 170: IMO it might be safe for now to always assume fullPageLogin since i assume an auth gate will need to gate entire pages rather than sections of a page

src/Form/index.tsx
bo-dun-1: oh hmm whoops. i could pretty easily update the backend logic so completed can't be unset?

src/elements/components/FeatheryAuthGate.tsx
```diff
@@ -0,0 +1,178 @@
...
+}: {
+  authId?: string;
+  formProps: FormProps;
+  whitelistPath?: string;
+  fullPageLogin?: boolean;
+  onLogin?: () => void;
+  onLogout?: () => void;
+  children: JSX.Element;
+}) => {
+  const formCompleted = initInfo().sessions[formProps.formName]?.form_completed;
```
bo-dun-1, line 47: instead of needing to store sessions globally, can't you leverage the onFormComplete callback handler exposed by the Form component?

bo-dun-1: and if that callback handler event is buggy (which it might be), it'd probably be better to robustify the handler rather than store session globally

src/elements/components/FeatheryAuthGate.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+  // Need to use this flag because when doing magic link login the onChange
+  // event doesn't seem to be added early enough to catch the first event which
+  // is the one containing the token. subsequent events do not contain the token
+  const hasAuthedRef = useRef(false);
+  // Use this render state to force re-evaluation of initState, since initState isn't reactive as-is
```
bo-dun-1, line 53: what initState use case are you trying to refresh for?

src/elements/components/FeatheryAuthGate.tsx
```diff
@@ -0,0 +1,178 @@
...
+  // This is a flag so we only redirect to the login start step immediately
+  // after auth, not during other form navigation
+  redirectAfterLogin: false,
+  sentAuth: false,
+  clearLoader: () => {},
+  onLogin: () => {},
+  onLogout: () => {}
+};
+
+const FeatheryAuthGate = ({
```
bo-dun-1, line 30: is this the best prop name? other ideas include
* LoginFormWrapper
* LoginFormGate
* AuthFormGate
* AuthFormWrapper
* FormLogin
* FormAuth

could be a good idea to poll the team on some ideas to see what reads clearer for them

src/utils/init.ts
bo-dun-1: yeah IMO authId and authClient should probably both be moved there

src/utils/client.ts
bo-dun-1: nice catch

src/utils/init.ts
```diff
@@ -35,13 +35,11 @@ type InitOptions = {
   userId?: string;
   preloadForms?: string[];
   userTracking?: 'cookie' | 'fingerprint';
-  authId?: string;
-  authEmail?: string;
```
bo-dun-1, line 39: should we remove authEmail and authPhoneNumber from the backend / db too if we're deprecating them?

src/Form/index.tsx
obmelvin: that's at least the easiest solve from my end, so if you are fine with that change it sounds good to me. thinking about it more, I do think it makes sense to freeze completed once it's true - even if a user goes back and changes things they shouldn't be able to unset required fields

src/elements/components/FeatheryAuthGate.tsx
obmelvin: fair point about using the onFormComplete callback, that is probably a more direct way to handle this

src/utils/init.ts
obmelvin: probably. can you see when the last time either field was set to confirm that no one is using them?

src/elements/components/FeatheryAuthGate.tsx
obmelvin: oh, yeah that's a fair point. was trying to not generalize too much from our case, but it does seem like a more safe default than I was thinking

src/elements/components/FeatheryAuthGate.tsx
obmelvin: yeah we could export it and use it

src/elements/components/FeatheryAuthGate.tsx
obmelvin: will double check if there's a reason that doesn't work, or if I was just trying to clear the loader in the same spot as the Form. agreed that would be better

src/elements/components/FeatheryAuthGate.tsx
obmelvin: oh, the reason I didn't do that is because this logic isn't just needed for when the onboarding questions are completed, but also determining if the onboarding form was already completed in the past

there was already an object to store sessions in initState so I leveraged that for this info

src/elements/components/FeatheryAuthGate.tsx
obmelvin: updated comment to reflect the updated code. don't think I could get rid of this without further changes because authId is still not react/redux state

src/Form/index.tsx
```diff
@@ -683,14 +684,15 @@ function Form({
         // @ts-expect-error TS(2345): Argument of type 'Promise<any[]>' is not assignabl... Remove this comment to see the full error message
         .fetchSession(formPromise, true)
         .then(([session, steps]) => {
+          initState.sessions[formName] = session;
```
obmelvin, line 687: remove this. should be able to store completed form boolean in the context. rename sessions to cachedSessions

src/elements/components/FeatheryAuthGate.tsx
```diff
@@ -0,0 +1,167 @@
...
+import throttle from 'lodash.throttle';
+import LoaderContainer from './LoaderContainer';
+
+const TEN_SECONDS_IN_MILLISECONDS = 1000 * 10;
+const FIVE_MINUTES_IN_MILLISECONDS = 1000 * 60 * 5;
+
+export const authState = {
+  authClient: null,
+  authEmail: '',
+  authId: '',
```
obmelvin, line 23: we should move this into react state in a new commit

src/hooks/useRefState.ts
```diff
@@ -0,0 +1,33 @@
...
+    return () => {
+      state.current = null;
+    };
+  }, []);
+
+  const getState = () => {
+    return state.current;
+  };
+
+  const setState = (fn: any, dontCopy = false) => {
```
obmelvin, line 26: ended up adding this 2nd param to handle not copying the fns & authClient. I didn't want to change the default behavior since it would be confusing to have the fn behave differently in feathery-react and feathery-frontend

I guess ideally this would be exported and used in feathery-frontend

src/elements/components/FeatheryAuthGate.tsx
```diff
@@ -0,0 +1,187 @@
...
+const defaultValues = {
+  authClient: null as any,
+  authEmail: '',
+  authId: '',
+  authPhoneNumber: '',
+  // This is a flag so we only redirect to the login start step immediately
+  // after auth, not during other form navigation
+  redirectAfterLogin: false,
+  sentAuth: false,
+  onLogin: () => {},
```
obmelvin: what do you think of afterLogin and afterLogout?

src/elements/components/LoginForm.tsx
```diff
@@ -30,76 +31,93 @@ export const authState = {
   onLogin: () => {},
   onLogout: () => {}
 };
+export let getAuthState: () => typeof defaultValues = () => defaultValues;
+export let setAuthState: (updatedProps: any) => void = () => {};
 
-const FeatheryAuthGate = ({
+const LoginForm = ({
```
bo-dun-1, line 37: LoginProvider? i know we've done a bunch of renames but maybe that will feel more "gatey"

src/elements/components/LoginForm.tsx
```diff
@@ -30,76 +31,93 @@ export const authState = {
...
   authClient: authClientProp,
   authId: authIdProp,
   formProps,
   loginPath,
   onLogin = () => {},
   onLogout = () => {},
-  children
+  children = <></>
```
bo-dun-1, line 44: why does children need this default value?

src/utils/init.ts
```diff
@@ -31,51 +32,39 @@ type DeprecatedOptions = {
...
-  authPhoneNumber?: string;
   language?: string;
 } & DeprecatedOptions;
 
 type InitState = {
   initialized: boolean;
   sdkKey: string;
   preloadForms: { [formName: string]: any };
-  sessions: { [formName: string]: any };
+  cachedSessions: { [formName: string]: any };
```
bo-dun-1, line 45: maybe rename this back to sessions

bo-dun-1: or formSessions?

src/utils/client.ts
```diff
@@ -373,6 +365,7 @@ export default class Client {
       auth_phone: authPhone,
       auth_email: authEmail,
       is_stytch_template_key: isStytchTemplateKey,
+      active_sessions: Object.keys(initInfo().sessions),
```
obmelvin, line 368: need to update key based on BE

src/utils/client.ts
```diff
@@ -381,9 +374,18 @@ export default class Client {
...
-    return this._fetch(url, options).then((response) => {
-      return response ? response.json() : Promise.resolve();
-    });
+    return this._fetch(url, options)
+      .then((response) => {
+        return response ? response.json() : Promise.resolve();
+      })
+      .then((data: any) => {
+        if (data === undefined) return Promise.resolve();
+        Object.entries<boolean>(data.form_completed).forEach(
```
obmelvin, line 383: need to update based on BE

src/Form/index.tsx
```diff
@@ -683,16 +684,22 @@ function Form({
...
```
bo-dun-1, line 687: i think you should move this line into the fetchSession function. then you can also remove this line https://github.com/feathery-org/feathery-react/blob/master/src/utils/init.ts#L167

src/Form/index.tsx
```diff
@@ -1058,7 +1065,9 @@ function Form({
       submitData || ['button', 'text'].includes(metadata.elementType);
     if (!redirectKey) {
       if (explicitNav) {
+        setFormCompleted(true);
```
bo-dun-1, line 1068: let's remove this boolean state and instead directly use the session boolean variable?

src/utils/client.ts
```diff
@@ -373,9 +374,18 @@ export default class Client {
...
+          ([formKey, formCompleted]) =>
+            authState.setFormCompleted(formKey, formCompleted)
```
bo-dun-1, line 385: instead of needing a custom authState.setFormCompleted function, could we update the session form_copmleted flag directly then call rerenderAllForms?

src/Form/index.tsx
```diff
@@ -1058,7 +1065,9 @@ function Form({
...
         eventData.completed = true;
+        authState.setFormCompleted(formName, true);
```
bo-dun-1, line 1070: can just update the session flag directly here

src/Form/index.tsx
```diff
@@ -1070,7 +1079,16 @@ function Form({
         b.properties.actions.some((action: any) => action.type === ACTION_NEXT)
       );
       const terminalStep = !hasNext && nextStep.next_conditions.length === 0;
-      if (terminalStep) eventData.completed = true;
+      if (terminalStep) {
+        const authIntegration = getAuthIntegrationMetadata(integrations);
+        // Completed should be false if it is a 'terminal step' in the middle of
+        // a login flow with steps configured for after login
+        const isTerminalStepAuth =
```
bo-dun-1, line 1086: wonder if all this logic should be a auth helper function in a utils file in the  auth directory

src/elements/components/LoginProvider.tsx
```diff
@@ -0,0 +1,192 @@
...
+  useEffect(() => {
+    if (
+      // We should set loader for new auth sessions
+      window.location.search.includes('stytch_token_type') ||
+      isHrefFirebaseMagicLink() ||
+      // and existing ones
+      getStytchJwt()
+    ) {
+      authState.redirectAfterLogin = true;
+      // We always need to set the redirect flag, but on local hosted forms we don't want to set the loader
```
bo-dun-1, line 87: why don't we want to set the loader on localhost?

src/Form/index.tsx
bo-dun-1: realized we may not have an auth directory now, but could be nice to make a top level folder called auth, then put the LoginProvider, useAuth (maybe renamed to useFormAuth), and any utility functions in it

src/hooks/useAuth.tsx
```diff
@@ -1,71 +1,40 @@
...
-          loader: (
-            <div style={{ height: '10vh', width: '10vh' }}>
-              <Spinner />
-            </div>
-          )
-        }
-      }));
-    }
-  }, []);
+  const authId = authState.authId;
```
bo-dun-1, line 21: is this necessary? could just use authState.authId directly

src/utils/init.ts
```diff
@@ -212,14 +198,14 @@ function setValues(userVals: FieldValues, rerender = true): void {
 }
 
 function setAuthClient(client: any): void {
```
bo-dun-1, line 200: maybe move these functions to a utils file at /auth/utils.ts

(along with /auth/LoginProvider.tsx, /auth/useFormAuth.ts, etc.)

src/utils/init.ts
```diff
@@ -106,6 +95,12 @@ function init(sdkKey: string, options: InitOptions = {}): Promise<string> {
 
   initState.sdkKey = sdkKey;
   optionsAsInitState.forEach((key) => {
+    // ------------------------------------
+    // ------------------------------------
+    // I don't really get why removing authId and authClient caused this?
```
bo-dun-1, line 100: what is the error?

src/elements/components/LoginProvider.tsx
obmelvin: I should expand the comment. The loader isn't set on localhost so that the loader doesn't appear on initial load for every form if you are also logged into the dashboard on localhost

src/utils/init.ts
obmelvin: image

src/Form/index.tsx
```diff
@@ -305,9 +305,11 @@ function Form({
 
   // All mount and unmount logic should live here
   useEffect(() => {
-    initState.renderCallbacks[_internalId] = () => {
-      setRender((render) => !render);
-    };
+    registerRenderCallback(_internalId, {
```
bo-dun-1, line 308: can this function take three params - ID, key (form or loginProvider), and render callback?

src/Form/index.tsx
```diff
@@ -1079,11 +1065,11 @@ function Form({
       submitData || ['button', 'text'].includes(metadata.elementType);
     if (!redirectKey) {
       if (explicitNav) {
-        setFormCompleted(true);
         eventData.completed = true;
-        if (internalState[_internalId])
-          internalState[_internalId].formCompleted = true;
-        authHookCb.cb();
+        initState.formSessions[formName].form_completed = true;
```
bo-dun-1, line 1069: maybe get the session at the top of the component so it can be used directly where needed

const session = initState.formSessions[formName];
...
session.form_completed = true;
...
session.form_completed && ...

src/utils/init.ts
bo-dun-1: i can't figure it out either rip, maybe leave in the expect-error without this comment?

bo-dun-1 (approved):

PR #668, opened 2022-12-30
Stripe Phase 2 - collect payment action and stripe checkout support

ericpias:
See: https://docs.google.com/document/d/1KI1sSi54yGQKMVj7PBgDz0v-gy7M-qVSdYQbZdecI1s/edit# for details.

### DIFF
--- a/src/Form/index.tsx
+++ b/src/Form/index.tsx
@@ -49,7 +49,8 @@ import {
   usePayments,
   toggleProductSelection,
   isProductSelected,
-  setupPaymentMethodAndPay
+  setupPaymentMethod,
+  collectPayment
 } from '../integrations/stripe';
 import {
   ActionData,
@@ -94,6 +95,7 @@ import {
   ACTION_LOGOUT,
   ACTION_NEXT,
   ACTION_REMOVE_REPEATED_ROW,
+  ACTION_COLLECT_PAYMENT,
   ACTION_SELECT_PRODUCT,
   ACTION_SEND_MAGIC_LINK,
   ACTION_SEND_SMS,
@@ -839,7 +841,7 @@ function Form({
         ({ servar: { type } }: any) => type === 'payment_method'
       )
     ) {
-      const errors: any = await handleStepSubmissionPayment(formattedFields);
+      const errors: any = await setupStepPaymentMethod(formattedFields);
       const { errorMessage, errorField } = errors;
       if (errorMessage && errorField) {
         await setFormElementError({
@@ -959,27 +961,71 @@ function Form({
   // usePayments (Stripe)
   const [getCardElement, setCardElement] = usePayments();
 
-  async function handleStepSubmissionPayment(formattedFields: any) {
+  async function setupStepPaymentMethod(formattedFields: any) {
     for (let i = 0; i < activeStep.servar_fields.length; i++) {
       const servar = activeStep.servar_fields[i].servar;
       if (servar.type === 'payment_method') {
         const integrationData = integrations?.stripe;
         const actionData: ActionData = {
-          fieldVal: fieldValues[servar.key],
           servar,
           client,
           formattedFields,
           updateFieldValues,
-          step: activeStep,
           integrationData,
           targetElement: getCardElement(servar.key)
         };
-        const actionResult = await setupPaymentMethodAndPay(actionData);
+        const actionResult = await setupPaymentMethod(actionData);
         return actionResult ?? {};
       }
     }
-
     return {};
+  }
+
+  async function collectPaymentAction(triggerElement: any) {
+    const errorCallback = getErrorCallback({
+      // could be container or button but model as button for the time being...
+      trigger: lookUpTrigger(activeStep, triggerElement.id, 'container')
+    });
+    // validate all step fields and buttons.  Must be valid before payment.
+    const { invalid, inlineErrors: newInlineErrors } = validateElements({
+      elements: [...activeStep.servar_fields, ...activeStep.buttons],
+      triggerErrors: true,
+      errorType: formSettings.errorType,
+      formRef,
+      errorCallback,
+      setInlineErrors
+    });
+    if (invalid) return false;
+
+    // payment/checkout
+    const pm = activeStep.servar_fields.find(
+      ({ servar: { type } }: any) => type === 'payment_method'
+    );
+    const errors = await collectPayment({
+      triggerElement,
+      servar: pm ? pm.servar : null,
+      client,
+      formattedFields: formatStepFields(activeStep, false),
+      updateFieldValues,
+      integrationData: integrations?.stripe,
+      targetElement: pm ? getCardElement(pm.servar.key) : null
+    });
+    if (errors) {
+      const { errorMessage, errorField } = errors;
+      await setFormElementError({
+        formRef,
+        errorCallback,
+        fieldKey: errorField.servar ? errorField.servar.key : errorField.id,
+        message: errorMessage,
+        servarType: errorField.servar ? errorField.servar.type : '',
+        errorType: formSettings.errorType,
+        inlineErrors: newInlineErrors,
+        setInlineErrors: setInlineErrors,
+        triggerErrors: true
+      });
+      return false;
+    }
+    return true;
   }
 
   async function goToNewStep({
@@ -1274,7 +1320,9 @@ function Form({
           }
         }
       } else if (type === ACTION_BACK) await goToPreviousStep();
-      else if (type === ACTION_SELECT_PRODUCT) {
+      else if (type === ACTION_COLLECT_PAYMENT) {
+        if (!(await collectPaymentAction(element))) break;
+      } else if (type === ACTION_SELECT_PRODUCT) {
         await toggleProductSelection({
           productId: action.product_id,
           selectedProductIdFieldId: action.selected_product_id_field,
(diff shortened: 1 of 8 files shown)


src/Form/index.tsx
```diff
@@ -1028,7 +1033,7 @@ function Form({
   // usePayments (Stripe)
   const [getCardElement, setCardElement] = usePayments();
 
-  async function handleStepSubmissionPayment(
+  async function handleStepSubmissionPaymentMethodSetup(
```
bo-dun-1, line 1036: setupStepPaymentMethod for brevity?

src/utils/client.ts
```diff
@@ -574,11 +574,31 @@ export default class Client {
     );
   }
 
-  createPayment(paymentMethodFieldKey: string) {
-    return this.payment('POST', { field_id: paymentMethodFieldKey });
+  createPayment() {
+    return this.payment('POST');
```
bo-dun-1, line 578: maybe _payment since this is an internal method.

bo-dun-1: also i can't coment directly on the payment method for some reason, but you could do

const data = { form_key: this.formKey, ...(userId ? ....), ...extraBodyParams};
const options = {body: data}

which is slightly simpler. also you do {method: method} but that can just be {method}

bo-dun-1: also i think you may be able to assume userId always exists, i don't think we'd support payment actions without knowing who the user is.

src/utils/client.ts
```diff
@@ -574,11 +574,31 @@ export default class Client {
...
     return this.payment('PUT');
   }
+
+  async createCheckoutSession(successUrl: string, cancelUrl?: string) {
+    await initFormsPromise;
+    const { userId } = initInfo();
+    const url = `${API_URL}stripe/checkout/`;
+    const data = {
+      form_key: this.formKey,
+      ...(userId ? { user_id: userId } : {}),
```
bo-dun-1, line 591: i think you may be able to assume userId always exists, i don't think we'd support payment actions without knowing who the user is.

src/integrations/stripe.ts
```diff
@@ -162,6 +143,86 @@ export async function setupPaymentMethodAndPay(
   return null;
 }
 
+/**
+ * Used to either collect payment directly or launch stripe checkout to collect payment.
+ */
+export async function collectPayment(
+  actionData: ActionData,
+  stripePromise = getStripe()
```
bo-dun-1, line 151: not sure it's best practice to have a function call be the default. maybe default should be null or something and call getStripe() explicitly in the function?

src/integrations/stripe.ts
```diff
@@ -162,6 +143,86 @@ export async function setupPaymentMethodAndPay(
...
+  const {
+    client,
+    triggerElement, // action button that triggered this
+    formattedFields,
+    updateFieldValues,
+    integrationData: stripeConfig
+  } = actionData;
+
+  // sync fields to BE
+  await syncFieldChanges(client, formattedFields, stripeConfig);
```
bo-dun-1, line 162: this is fine for now i think but eventually these backend calls would ideally be unified so the data update is atomic and we save an extra network call

src/integrations/stripe.ts
```diff
@@ -162,6 +143,86 @@ export async function setupPaymentMethodAndPay(
...
+      stripePromise
+    );
+    if (payMethodResult) {
+      // error
+      return payMethodResult;
+    }
+  }
+
+  // Now collect payment or do stripe checkout
+
```
bo-dun-1, line 178: extra newline

src/integrations/stripe.ts
```diff
@@ -162,6 +143,86 @@ export async function setupPaymentMethodAndPay(
...
+    updateFieldValues,
+    integrationData: stripeConfig
+  } = actionData;
+
+  // sync fields to BE
+  await syncFieldChanges(client, formattedFields, stripeConfig);
+
+  // setup any payment method on this step
+  if (actionData.targetElement) {
+    const payMethodResult = await setupPaymentMethod(
```
bo-dun-1, line 166: what if payment method has already been set up? is that handled well?

src/integrations/stripe.ts
```diff
@@ -162,6 +143,86 @@ export async function setupPaymentMethodAndPay(
...
+      const {
+        success_url: successUrl = window.location.href,
+        cancel_url: cancelUrl = null
+      } = triggerElement?.properties || {};
+      const { checkout_url: checkoutUrl, error = null } =
+        await client.createCheckoutSession(successUrl, cancelUrl);
+      if (error) return { errorField: triggerElement, errorMessage: error };
+      // preserve browser back button function to current page/step
+      window.location.href = checkoutUrl;
+    } else {
```
bo-dun-1, line 192: instead of a blanket else, IMO it's best practice to be explicit w/ an else-if even if you believe there's only one other else branch for now. that may be true now but may lead to issues in future development. also helps witgh understanding the code

src/Form/index.tsx
```diff
@@ -1353,6 +1403,13 @@ function Form({
       });
     } else if (link === LINK_BACK) {
       await goToPreviousStep();
+    } else if (link === LINK_COLLECT_PAYMENT) {
+      await collectPaymentAction(button);
+      // For custom payment, also submit and navigate
+      await buttonOnSubmit({
```
bo-dun-1, line 1409: does submission / navigation always need to happen? couldn't collectPayment just open up a new tab with the stripe checkout session and leave the current step as is?

src/integrations/stripe.ts
ericpias: This is specifically called out in the spec and is a manual test case that I will test for.  See: https://docs.google.com/document/d/1ADF3b3u-u008Jyhws6g9DT9VEdYzLkQFLNltiLQAhAs/edit#heading=h.fneum7xxs0q9 under the runtime behavior bullet points.

src/utils/client.ts
ericpias: Eliminated extraBodyParams.

src/integrations/stripe.ts
ericpias: No changes made rn.

src/integrations/stripe.ts
ericpias: Actually, this was intentionally coded this way because legacy forms will not necessarily have have checkout_type (default is custom).  I will put the explicit check in along with the check of undefined, but it does seem messy to me.

src/Form/index.tsx
ericpias: There are two issues to consider in your comment.  

The first is the part about opening the stripe checkout in a new tab.  We certainly could do this but should we?  Also, would we always open the checkout page in a new tab or would it be a form designer's choice?  If it is a choice, then we need another property for it.  In my opinion, opening in another tab is unlikely to be what most people will want and I suggest that we leave this feature out until someone needs it.

With the above in mind, it was my thinking that with stripe checkout there is always a synchronous redirect happening in collectPaymentAction before the buttonOnSubmit call here.  Basically, the buttonOnSubmit call only ever happens with a custom payment.  This is why I coded it this way.

ericpias: Per our discussion, leaving this as is and not supporting checkout in another tab/window for rn.

src/Form/index.tsx
```diff
@@ -1028,33 +1030,85 @@ function Form({
...
     }
-
     return {};
   }
 
+  async function collectPaymentAction(
+    triggerElement: any,
+    setLoader = () => {}
+  ) {
+    const errorCallback = triggerElement.type
```
bo-dun-1, line 1058: i don't think elements have a type attribute

src/integrations/stripe.ts
```diff
@@ -60,37 +61,44 @@ function getObjectMappingValues(mappingObj: any, fieldValues: any) {
   }, {} as { [key: string]: any });
 }
 
+async function syncFieldChanges(
```
bo-dun-1, line 64: syncStripeFieldChanges?

src/Form/index.tsx
```diff
@@ -959,29 +961,78 @@ function Form({
...
+      errorCallback,
+      setInlineErrors
+    });
+    if (invalid) return false;
+
+    // payment/checkout
+    const pm = activeStep.servar_fields.find(
+      ({ servar: { type } }: any) => type === 'payment_method'
+    );
+    setLoader();
```
bo-dun-1, line 1007: loader logic doesn't need to exist here anymore. loaders are set if configured before all actions are run and cleared after they run

src/utils/formHelperFunctions.ts
```diff
@@ -115,7 +115,7 @@ export const lookUpTrigger = (step: any, elementID: any, elementType: any) => {
   let payload = {};
   if (elementType === 'button') {
     const element = step.buttons.find((button: any) => button.id === elementID);
-    payload = { text: element.properties.text };
+    payload = { text: element?.properties?.text ?? '' };
```
bo-dun-1, line 118: in what case do we not have the element?

ericpias: This is called from collectPaymentAction in Form/index.tsx.  CollectPaymentAction does a validation first which potentially calls back on the sdk errorCallback which needs a trigger element.  However, this action could be triggered by either a button or a container.  After thinking about it quite a bit, I decided to sort of leave it as is with the trigger looking like a button but with no text nor id property.

bo-dun-1: you should pass it in with elementType = 'container' then. We actually handle that case well since it just falls back to the default payload returned

bo-dun-1 (approved):

PR #669, opened 2023-01-03
multiple click actions support

bo-dun-1:


src/utils/callbackQueue.tsx
```diff
@@ -1,7 +1,7 @@
 import { isObjectEmpty } from './primitives';
 import Spinner from '../elements/components/Spinner';
-import { LINK_NEXT } from '../elements/basic/ButtonElement';
 import React from 'react';
+import { ACTION_NEXT } from './elementActions';
```
obmelvin, line 4: not sure if this export should be deleted or if we still want it in the filter (think the former)

src/elements/basic/ButtonElement.tsx
```diff
@@ -5,30 +5,7 @@ import TextNodes from '../components/TextNodes';
...
-const LINK_VERIFY_SMS = 'verify_sms';
-const SUBMITTABLE_LINKS = [
-  LINK_NEXT,
-  LINK_TRIGGER_PLAID,
-  LINK_LOGOUT,
-  LINK_SEND_MAGIC_LINK,
-  LINK_SEND_SMS,
-  LINK_VERIFY_SMS
-];
+import { ACTION_NEXT, SUBMITTABLE_ACTIONS } from '../../utils/elementActions';
```
ericpias, line 8: ACTION_NEXT not used

ericpias (commented):
I went through a lot here and spot tested a number of things including: action migration, repeat (add/remove) actions, back, next, stripe actions, url, actions on text, group/cell and buttons.  However, this change is critical and really could use another pair of eyes on it.  I especially thing that validating the the changes around the oauth, logout, magic link and sms actions would be important (Oliver).

PR #670, opened 2023-01-03
add form context via ref

obmelvin:


src/Form/index.tsx
```diff
@@ -121,6 +156,9 @@ export interface Props {
   initialStepId?: string;
   display?: 'inline' | 'modal';
   elementProps?: ElementProps;
+  formContextRef?: React.MutableRefObject<null | ReturnType<
```
obmelvin, line 159: lmk what you think about the name
alternatives
`ref`
`formRef`
`contextRef`

src/Form/index.tsx
```diff
@@ -341,6 +392,13 @@ function Form({
 
     setAutoValidate(false); // Each step to initially not auto validate
 
+    if (
+      formContextRef &&
+      Object.prototype.hasOwnProperty.call(formContextRef, 'current')
+    )
+      // Need to update the context because it has activeStep closure'd in
+      formContextRef.current = getFormContextHelper();
```
obmelvin, line 400: I tried just setting .current once but I was getting a runtime error that `key` didn't exist on `newStep.key` when running 

```
setStep: (stepKey: any) => {
    changeStep(stepKey, newStep.key, props.steps, props.history);
  },
```

obmelvin: another alt for naming would be to make this `ref.current.context = ....`

src/Form/index.tsx
```diff
@@ -104,6 +104,43 @@ import usePrevious from '../hooks/usePrevious';
...
+  setProgress: (val: any) => props.setUserProgress(val),
+  setStep: (stepKey: any) => {
+    changeStep(stepKey, newStep.key, props.steps, props.history);
+  },
+  step: {
+    style: {
+      backgroundColor: newStep?.default_background_color
+    }
+  },
+  userId: initInfo().userId,
```
bo-dun-1, line 140: IMO userId should actually be pulled into like the return value of the init call rather than in per-for context

src/Form/index.tsx
```diff
@@ -254,6 +295,17 @@ function Form({
...
   const stepCSS = useMemo(() => calculateStepCSS(activeStep), [activeStep]);
 
+  const getFormContextHelper = (newStep = activeStep) => {
+    return getFormContext(newStep, {
+      history,
+      client,
+      updateFieldOptions,
+      setUserProgress,
+      steps,
+      activeStep
```
bo-dun-1, line 305: is activeStep even needed if we're using newStep already?

src/Form/index.tsx
```diff
@@ -104,6 +104,43 @@ import usePrevious from '../hooks/usePrevious';
...
+  setStep: (stepKey: any) => {
+    changeStep(stepKey, newStep.key, props.steps, props.history);
+  },
+  step: {
+    style: {
+      backgroundColor: newStep?.default_background_color
+    }
+  },
+  userId: initInfo().userId,
+  stepName: newStep?.key ?? ''
```
bo-dun-1, line 141: curStepName?

src/Form/index.tsx
```diff
@@ -104,6 +104,43 @@ import usePrevious from '../hooks/usePrevious';
...
+      step_key: props.activeStep.key,
+      event: 'load',
+      completed: flag
+    }),
+  setOptions: props.updateFieldOptions(props.steps),
+  setProgress: (val: any) => props.setUserProgress(val),
+  setStep: (stepKey: any) => {
+    changeStep(stepKey, newStep.key, props.steps, props.history);
+  },
+  step: {
```
bo-dun-1, line 135: attributes: { stepBackgroundColor: newStep?.default_background_color } maybe is cleaner?

src/Form/index.tsx
```diff
@@ -104,6 +104,43 @@ import usePrevious from '../hooks/usePrevious';
...
+) => ({
+  setValues,
+  setFormCompletion: (flag: boolean) =>
+    props.client.registerEvent({
+      step_key: props.activeStep.key,
+      event: 'load',
+      completed: flag
+    }),
+  setOptions: props.updateFieldOptions(props.steps),
+  setProgress: (val: any) => props.setUserProgress(val),
```
bo-dun-1, line 131: setFormProgress?

src/Form/index.tsx
```diff
@@ -104,6 +104,43 @@ import usePrevious from '../hooks/usePrevious';
...
+  }
+) => ({
+  setValues,
+  setFormCompletion: (flag: boolean) =>
+    props.client.registerEvent({
+      step_key: props.activeStep.key,
+      event: 'load',
+      completed: flag
+    }),
+  setOptions: props.updateFieldOptions(props.steps),
```
bo-dun-1, line 130: setFieldOptions?

src/Form/index.tsx
bo-dun-1: I'm a fan of contextRef, IMO it prevents the user from thinking its going to be a ref pointing to the actual form element & provides consistency with the callback functions, where we call its parameter "context" in docs. I also think we should pull the validate and render functions into this context instead of leaving them as global entities

src/Form/index.tsx
obmelvin: oh, yeah think I can change the activeStep.key to newStep.key and clean up the code a bit
``` setFormCompletion: (flag: boolean) =>
          client.registerEvent({
            step_key: activeStep.key,
            event: 'load',
            completed: flag
          }),```

src/Form/index.tsx
obmelvin: added to asana task

PR #671, opened 2023-01-04
V1.0.573

obmelvin:


src/Form/index.tsx
```diff
@@ -341,6 +390,13 @@ function Form({
 
     setAutoValidate(false); // Each step to initially not auto validate
 
+    if (
+      contextRef &&
+      Object.prototype.hasOwnProperty.call(contextRef, 'current')
+    )
+      // Need to update the context because it has activeStep closure'd in
+      contextRef.current = getFormContextHelper();
```
bo-dun-1, line 398: should this actually be called in the getNewStep fucntion before the onLoad() so the user has these functions available ASAP

bo-dun-1 (approved):

src/Form/index.tsx
```diff
@@ -104,6 +104,43 @@ import usePrevious from '../hooks/usePrevious';
 import ReactPortal from './components/ReactPortal';
 import { replaceTextVariables } from '../elements/components/TextNodes';
 
+// getFormContext needs to live outside of the component so that we can export
+// the types for the end user
+const getFormContext = (
```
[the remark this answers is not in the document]
obmelvin: yeah I should've already done that

src/Form/index.tsx
obmelvin: oh yeah I can just call it there and pass in newStep instead of needing activeStep to be set

src/Form/index.tsx
```diff
@@ -185,7 +182,7 @@ function Form({
     brandPosition: undefined,
     allowEdit: 'yes'
   });
-  const [inlineErrors, setInlineErrors] = useState({});
+  const [inlineErrors, setInlineErrors] = useState<Record<string, unknown>>({});
```
bo-dun-1, line 185: should be Record<string, {message: string; index: number}>

PR #675, opened 2023-01-05
handle context properly for vanillajs

obmelvin:


src/utils/formContext.ts
```diff
@@ -2,55 +2,83 @@ import { changeStep, FieldOptions } from './formHelperFunctions';
...
+    const { contextProps, currentStep } = initInfo().formState[formUuid];
+    if (!contextProps) return;
+    changeStep(
+      stepKey,
+      currentStep.key,
+      contextProps.steps,
+      contextProps.history
+    );
   },
   step: {
```
obmelvin, line 59: step, userId, and stepName all face the same problem - they are directly returned and so get baked in as undefined. even if we adjusted how this is set a bit so that they weren't undefined, they would still be locked into the initial values

not really sure how to address this without turning them into fns or making the user call getContext() instead of directly accessing .context

package.json
```diff
@@ -101,6 +101,7 @@
     "sass": "^1.49.11",
     "scriptjs": "^2.5.9",
     "uuid": "^8.3.2",
+    "uuidv4": "^6.2.13",
```
obmelvin, line 104: remove this package

src/Form/index.tsx
```diff
@@ -114,6 +115,7 @@ export interface Props {
   onViewElements?: string[];
   initialValues?: FieldValues;
   initialStepId?: string;
+  uuid?: string;
```
obmelvin, line 118: need to edit Props so this isn't exposed to the user as a prop that could be passed in

src/Form/index.tsx
```diff
@@ -147,6 +153,7 @@ function Form({
   onViewElements = [],
   initialValues = {},
   initialStepId = '',
+  uuid,
```
obmelvin, line 156: pass this in at the bottom, don't need to set in the useEffect

src/Form/index.tsx
```diff
@@ -608,12 +608,22 @@ function Form({
...
     }
     newStep = JSON.parse(JSON.stringify(newStep));
-    if (
-      contextRef &&
-      Object.prototype.hasOwnProperty.call(contextRef, 'current')
-    )
-      // Need to update the context each time because it has the step closure'd in
-      contextRef.current = getFormContextHelper(newStep);
+
+    initState.formState[formUuid] = {
```
obmelvin, line 612: make a new internalState global for these

obmelvin: add comment in internalState and here about what these properties are for
explain need for context & having up to date values during runtime

src/Form/index.tsx
```diff
@@ -622,38 +632,34 @@ function Form({
...
-              completed: flag
-            }),
-          setOptions: updateFieldOptions(steps, newStep),
-          firstStepLoaded: first,
-          integrationData
-        };
-      },
-      newStep
-    );
+      return {
```
obmelvin, line 642: should be able to clean up the redundant props - verify that they are updated properly after step change due to context changes

src/index.tsx
```diff
@@ -27,7 +28,12 @@ function getAllValues() {
 function renderAt(elementId: any, props: FormProps) {
   const container = featheryDoc().getElementById(elementId);
 
-  ReactDOM.render(<Form {...props} />, container);
+  const uuid = props.uuid ?? uuidv4();
+  console.log('renderAt uuid', uuid);
+
+  ReactDOM.render(<Form {...props} uuid={uuid} />, container);
+
+  return { context: getFormContext(uuid) };
```
obmelvin, line 36: can just return gFC

src/utils/formContext.ts
```diff
@@ -2,55 +2,83 @@ import { changeStep, FieldOptions } from './formHelperFunctions';
...
     style: {
-      backgroundColor: newStep?.default_background_color
+      backgroundColor:
+        // This gets baked in.. and isn't present in the context as a result
+        initInfo().formState[formUuid]?.currentStep?.default_background_color
     }
   },
+  // This gets baked in..
   userId: initInfo().userId,
```
obmelvin, line 67: return userId when you call init
need to update docs & communicate change

src/utils/formContext.ts
```diff
@@ -2,55 +2,83 @@ import { changeStep, FieldOptions } from './formHelperFunctions';
...
-  stepName: newStep?.key ?? '',
+  // This gets baked in..
+  stepName: initInfo().formState[formUuid]?.currentStep?.key ?? '',
```
obmelvin, line 69: add getAttributes fn to context. keep step & stepName for now - mark to deprecation 

update docs for getAttributes

src/utils/formContext.ts
obmelvin: issue noted in comment

src/utils/internalState.ts
```diff
@@ -0,0 +1,18 @@
...
+    [formUuid: string]: { contextProps: null | ContextProps; currentStep: any };
+  };
+};
+
+const internalState: InternalState = {
+  formState: {}
+};
+
+// do we want this fn? mostly did it to be semi-consistent with the initState
+function getInternalState() {
```
obmelvin, line 14: do we want this fn? mostly did it to be semi-consistent with the initState

src/utils/internalState.ts
```diff
@@ -0,0 +1,17 @@
+import { ContextProps } from './formContext';
+
+type InternalState = {
+  formState: {
```
obmelvin, line 4: probably should drop the formState prop? Since init is already global, I guess things here are likely to be tied to a specific form so maybe makes since for the UUIDs to be the primary key for this obj

src/Form/index.tsx
```diff
@@ -124,6 +126,10 @@ export interface Props {
   children?: JSX.Element;
 }
 
+interface InternalProps {
+  uuid: string; // Used to uniquely identify forms, even when the same form is rendered multiple times
```
bo-dun-1, line 130: formId?

src/Form/index.tsx
```diff
@@ -1570,7 +1572,7 @@ export default function FormWithRouter({
         {/* eslint-disable-next-line @typescript-eslint/ban-ts-comment */}
         {/* @ts-ignore */}
         <Route path='/'>
-          <Form {...props} formName={formName} key={formName} />
+          <Form {...props} formName={formName} key={formName} uuid={uuidv4()} />
```
bo-dun-1, line 1575: what if FormWithRouter gets rerendered? wouldn't this generate a new uuid for the same form

src/Form/index.tsx
bo-dun-1, line 156: _internalId?

src/Form/index.tsx
```diff
@@ -608,12 +604,22 @@ function Form({
...
     newStep = JSON.parse(JSON.stringify(newStep));
-    if (
-      contextRef &&
-      Object.prototype.hasOwnProperty.call(contextRef, 'current')
-    )
-      // Need to update the context each time because it has the step closure'd in
-      contextRef.current = getFormContextHelper(newStep);
+
+    internalState.formState[uuid] = {
+      currentStep: newStep,
```
bo-dun-1, line 609: should this just be a flat object with currentStep as just one of the props

src/utils/internalState.ts
bo-dun-1: i don't think we need this function. initState function is used because it also contains validation logic that runs

src/utils/internalState.ts
bo-dun-1: yeah agreed

src/utils/formContext.ts
```diff
@@ -1,59 +1,101 @@
...
+    if (!contextProps) return;
+    changeStep(
+      stepKey,
+      currentStep.key,
+      contextProps.steps,
+      contextProps.history
+    );
   },
-  step: {
+  getAttributes: () => ({
```
bo-dun-1, line 67: stepProperties()?

src/utils/formContext.ts
```diff
@@ -1,59 +1,101 @@
...
+    stepName: getInternalState().formState[formUuid]?.currentStep?.key ?? '',
```
bo-dun-1, line 68: just name? if we name the function stepProperties

src/utils/formContext.ts
```diff
@@ -1,59 +1,101 @@
...
     style: {
-      backgroundColor: newStep?.default_background_color
+      backgroundColor:
```
bo-dun-1, line 70: should we just flatten this for now? to backgroundColor instead of style: {backgroundColor}. IMO this will be refactored down the line anyways since default_background_color is being deprecated in favor of the grid styles

src/utils/formContext.ts
obmelvin: re: moving this out to be returned by init - that would only handle existing sessions when redirecting from magic link/oauth the integration will call submitAuthInfo after init has ran

src/Form/index.tsx
```diff
@@ -1557,10 +1550,18 @@ function Form({
   );
 }
 
+// normal <Form /> component is exported with just `props: Props`, so need to
+// export this for use in renderAt. Not needed for FormWithRouter since this is
+// all in the same file
+export function InternalForm(props: Props & InternalProps) {
```
bo-dun-1, line 1556: the renderAt form needs router functionality no? for browser nav and stuff

obmelvin: true, adjusted InternalForm and FormWithRouter so that internal form uses the router without us exposing the InternalProps to external users

src/Form/index.tsx
```diff
@@ -1557,10 +1550,14 @@ function Form({
   );
 }
 
-export default function FormWithRouter({
+// normal <Form /> (aka FormWithRouter) component is exported with just `props:
+// Props`, so need this component to support exposing _internalId for use in
+// renderAt without exposing InternalProps to SDK users
+export function InternalForm({
```
bo-dun-1, line 1556: JSForm?

src/Form/index.tsx
```diff
@@ -1570,9 +1567,19 @@ export default function FormWithRouter({
...
+            key={formName}
+            _internalId={_internalId}
+          />
         </Route>
       </BrowserRouter>
     );
   else return null;
 }
+
+export default function FormWithRouter(props: Props): JSX.Element | null {
```
bo-dun-1, line 1582: ReactForm?

bo-dun-1 (approved):

PR #677, opened 2023-01-10
rework how initial step is set when authenticating

obmelvin:


src/Form/index.tsx
```diff
@@ -1484,35 +1529,40 @@ function Form({
     runUserCallback(onFormComplete).then(redirectForm);
   }, [anyFinished]);
 
+  const wholePageLoader = stepLoader && (
+    <div
+      style={{
+        // FFF works for us, but not necessarily for others. thoughts?
+        // activeStep isn't set in the case of waiting on auth handshake
+        backgroundColor: `#${activeStep?.default_background_color ?? 'FFF'}`,
```
obmelvin, line 1537: activeStep isn't set in the case of waiting on auth handshake

FFF works for us, but not necessarily for others. thoughts?

src/Form/index.tsx
```diff
@@ -691,7 +717,21 @@ function Form({
...
+                _integrations: session.integrations,
+                _steps: steps
+              })) ||
+            // ------------------------------------------
+            // ------------------------------------------
+            // should getNextAuthStep or initialStepId be first?
+            // ------------------------------------------
+            // ------------------------------------------
+            // ------------------------------------------
             initialStepId ||
```
obmelvin, line 735: should getNextAuthStep or initialStepId be first?

src/Form/index.tsx
```diff
@@ -577,27 +583,9 @@ function Form({
       integrations.stytch?.metadata ?? integrations.firebase?.metadata;
     const authSteps = metadata?.auth_gate_steps ?? [];
     if (authSteps.length > 0) {
```
bo-dun-1, line 585: this authSteps check can probably also be in getNextAuthStep, and just return '' if there is no next authStep

src/Form/index.tsx
```diff
@@ -1462,6 +1496,11 @@ function Form({
       }))
   };
 
+  // If we delayed setting stepKey after loading the session, we want to set the
```
obmelvin, line 1499: do this as a cb fn to the client

src/utils/client.ts
```diff
@@ -33,7 +34,7 @@ const CDN_URL_OPTIONS = {
   production: 'https://cdn.feathery.io/api/'
 };
 
-const environment = 'production';
+const environment = 'local';
```
bo-dun-1, line 37: revert this please

src/utils/client.ts
```diff
@@ -48,9 +49,11 @@ const TYPE_MESSAGES_TO_IGNORE = [
 export default class Client {
   formKey: any;
   ignoreNetworkErrors: any; // this should be a ref
-  constructor(formKey?: any, ignoreNetworkErrors?: any) {
+  authCb: any;
```
bo-dun-1, line 52: authCallback? may be a little more clear

src/utils/client.ts
```diff
@@ -347,11 +351,10 @@ export default class Client {
   submitAuthInfo({ authId, authPhone = '', authEmail = '' }: any) {
     const { userId } = initInfo();
     initState.authId = authId;
-    initState.redirectAfterLogin = true;
+    this.authCb();
```
bo-dun-1, line 354: we should probably check if authCb exists first

src/Form/index.tsx
```diff
@@ -212,12 +218,8 @@ function Form({
   const [render, setRender] = useState(false);
 
   const [loaders, setLoaders] = useState({});
-  const clearLoaders = (clearAuthLoader = false) => {
-    if (clearAuthLoader) setLoaders({});
-    else
-      setLoaders(({ auth }: any) => ({
-        auth
-      }));
+  const clearLoaders = () => {
```
bo-dun-1, line 221: nit:

const clearLoaders = () => setLoaders({});

src/Form/index.tsx
```diff
@@ -617,6 +584,9 @@ function Form({
       updateFieldOptions
     };
 
+    // Clear auth (and other) loaders and reset auth flag
+    redirectAfterLoginRef.current = false;
```
bo-dun-1, line 588: couldn't this be set in getNextAuthStep instead? so this logic is in the hook

bo-dun-1 (approved):

src/Form/index.tsx
obmelvin: sent on slack. posting here for posterity: 

yeah I wanted to move where the flag is reset to false, but earlier when I tried colocating with the other auth code it wasn’t working properly. It looks like with these other changes I can reset it in the auth hook

but I still need to return the ref rather than the value because otherwise the value is stale (stuck at true for the last render), which then prevents navigation to any other step because the code sees the flag as true so it sets the next step to the login start step

PR #681, opened 2023-01-12
fix infinite loading edge case due to cookies shared across ports

obmelvin:


src/Form/index.tsx
```diff
@@ -196,7 +197,9 @@ function Form({
   >({});
   const [, setRepeatChanged] = useState(false);
 
-  const [integrations, setIntegrations] = useState<Record<string, any>>({});
+  const [integrations, setIntegrations] = useState<null | Record<string, any>>(
+    null
```
obmelvin, line 201: I had to change the initial value to null so we can distinguish between integrations not being set yet vs the form not having integrations

bo-dun-1 (approved):

PR #682, opened 2023-01-12
handle form version conflicts

bo-dun-1:


astrozzc (approved):
Tested locally and the form is reloading if the version is outdated

PR #695, opened 2023-01-24
add prop to onSubmit context for fields not hidden by hide if rules

obmelvin:


src/utils/formHelperFunctions.ts
```diff
@@ -12,20 +13,21 @@ import { evalComparisonRule, ResolvedComparisonRule } from './logic';
...
     const servar = field.servar;
     let value;
     // Only use base64 for signature if these values will be presented to the user
     const val = fieldValues[servar.key];
     if (forUser && servar.type === 'signature') {
       value =
-        val !== null
-          ? // @ts-expect-error TS(2345): Argument of type 'string | number | boolean | stri... Remove this comment to see the full error message
-            Promise.resolve(val).then((file) => toBase64(file))
+        val !== null && (val instanceof File || val instanceof Promise)
```
obmelvin, line 27: had to add the instanceof to conditional to make TS happy

src/Form/index.tsx
```diff
@@ -879,7 +878,7 @@ function Form({
     if (typeof onSubmit === 'function') {
       let stepChanged = false;
       await runUserCallback(onSubmit, () => ({
-        submitFields: formattedFields,
+        submitFields: getNonHiddenFields(activeStep, formattedFields),
```
bo-dun-1, line 881: can this logic just live in the formatStepFields function?

src/utils/formHelperFunctions.ts
```diff
@@ -76,6 +78,23 @@ export function getDefaultFieldValue(field: any) {
   }
 }
 
+export function getNonHiddenFields(currentStep: any, formattedFields: any) {
+  const hiddenState: { key: string; hidden: boolean }[] =
+    currentStep.servar_fields.map((servar: any) => ({
+      key: servar.servar.key,
+      hidden: shouldElementHide({ element: servar })
```
bo-dun-1, line 85: cleaner to just call .filter and run the shouldElementHide function. check out how we use this function in other places

src/Form/index.tsx
obmelvin: formatAllFormFields calls formatStepFields per step, so that would remove the conditionally hidden fields from more things than we want

obmelvin: I guess we can just add a prop to formatStepFields

src/utils/formHelperFunctions.ts
```diff
@@ -2,30 +2,43 @@ import getRandomBoolean from './random';
...
+export const formatStepFields = (
+  step: any,
+  forUser: boolean,
+  onlyVisibleFields: boolean
+) => {
+  const formattedFields: Record<
+    string,
+    { value: any; type: string; displayText: string }
+  > = {};
+  const fields = onlyVisibleFields
```
bo-dun-1, line 26: let fields = step.servar_fields;
if (onlyVisibleFields) fields = fields.filter(....);

bo-dun-1 (approved):

PR #696, opened 2023-01-24
Support modal form vertical scrolling of content within the modal

ericpias:


src/Form/index.tsx
```diff
@@ -1506,35 +1506,43 @@ function Form({
...
-          css={{
-            ...stepCSS,
-            ...style,
-            position: 'relative',
-            marginBottom: addChin ? '80px' : '0',
-            display: 'flex'
-          }}
+        <div
+          css={
+            display === 'modal'
```
bo-dun-1, line 1511: why do you need another div vs adding this CSS to the BootstrapForm component directly?

ericpias: The div establishes the overflow-y and the limited height.  The form establishes the content height/width so that it scrolls on overflow inside its parent.  I tried using the form as the scroll container and ran into issues with the height/width of the content.  This was the cleanest solution I found.

PR #717, opened 2023-02-15
onClientReady callback function

bo-dun-1:


src/auth/internal/utils.ts
```diff
@@ -24,34 +12,27 @@ export function inferEmailLoginFromURL(featheryClient: Client) {
   else return emailLoginFirebase(featheryClient);
 }
 
-export function inferAuthLogout() {
-  let logout;
+export async function inferAuthLogout() {
+  // logout may not have a value in certain cases, i.e. stytch is already logged out so there is no jwt
```
obmelvin, line 16: i'd probably just get rid of this comment now, rather than rewrite it. the current conditional doesn't really need one

src/auth/internal/utils.ts
```diff
@@ -24,34 +12,27 @@ export function inferEmailLoginFromURL(featheryClient: Client) {
...
+  authState.authEmail = '';
+  authState.setAuthId('');
 }
 
 export function isAuthStytch() {
-  const authClient = getAuthClient();
-  if (!authClient) return;
-  const isAuthClientStytch = Object.getOwnPropertySymbols(authClient)
+  if (!authState.client) return;
+  return Object.getOwnPropertySymbols(authState.client)
```
obmelvin, line 33: I'll have to double check if this works. we are still using the old version of stytch, so can't remember if the old one has these symbols

obmelvin (approved):

PR #718, opened 2023-02-16
add context to LoginProvider

obmelvin:


src/auth/LoginProvider.tsx
```diff
@@ -137,17 +145,15 @@ const LoginProvider = ({
...
 
-        if (session) {
-          stytchClient.session.authenticate({
+        if (authState.client.session.getSync()) {
+          authState.client.session.authenticate({
             session_duration_minutes: 1440
           });
         } else if (hasAuthedRef.current) {
-          stytchClient.session.revoke().then(logoutActions);
+          // There is no session, so need to revoke it
```
obmelvin, line 155: I think this change will fix the sentry error about user needing to be authenticated

bo-dun-1 (approved):

PR #729, opened 2023-03-01
cast submit custom values based on field type

obmelvin:


src/utils/formHelperFunctions.ts
```diff
@@ -437,3 +437,35 @@ export function getInitialStep({
...
+  let castVal = val;
+  switch (servarType) {
+    case 'currency':
+    case 'integer_field':
+    case 'rating':
+    case 'slider':
+      castVal = Number(val);
+      break;
+    case 'checkbox':
+      castVal = val === 'true';
```
bo-dun-1, line 453: or 'True'?

bo-dun-1: actually might be better to default to truthy and otherwise case "false" and "False" to false

src/Form/index.tsx
```diff
@@ -672,8 +673,12 @@ function Form({
         .then(([session, steps]) => {
           updateBackNavMap(session.back_nav_map);
           setIntegrations(session.integrations);
-          if (!isObjectEmpty(initialValues))
-            clientInstance.submitCustom(initialValues, false);
+          if (!isObjectEmpty(initialValues)) {
+            clientInstance.submitCustom(initialValues, {
+              override: false,
+              servarKeyToTypeMap: getServarTypeMap(steps)
```
bo-dun-1, line 679: could you just cast initialValues before passing it into submitCustom? might be a cleaner API

src/Form/index.tsx
```diff
@@ -672,8 +674,13 @@ function Form({
         .then(([session, steps]) => {
           updateBackNavMap(session.back_nav_map);
           setIntegrations(session.integrations);
-          if (!isObjectEmpty(initialValues))
+          if (!isObjectEmpty(initialValues)) {
+            const servarKeyToTypeMap = getServarTypeMap(steps);
+            Object.entries(initialValues).map(([key, val]) => {
+              initialValues[key] = castVal(servarKeyToTypeMap[key], val);
```
bo-dun-1, line 680: you should maybe store it not in the original user-passed object since we probably dont want to generate unintended side effects for the user

src/utils/formHelperFunctions.ts
```diff
@@ -437,3 +437,35 @@ export function getInitialStep({
...
+      break;
+    default:
+      castVal = String(val);
+      break;
+  }
+
+  return castVal;
+}
+
+export function getServarTypeMap(steps: any) {
```
bo-dun-1, line 463: not needed to address in this PR, but this file has grown massive... probably worth spreading out at some point

bo-dun-1 (approved):

PR #732, opened 2023-03-07
login via redirect

obmelvin:


src/auth/LoginForm.tsx
```diff
@@ -187,6 +199,10 @@ const LoginForm = ({
         <JSForm {...formProps} _internalId={_internalId} />
       </div>
     );
+  } else if (apexForm) {
+    // If logged in and have finished onboarding questions for an apex form, then we need to redirect back to application
+    window.location.href = 'http://localhost:3000';
```
obmelvin, line 204: should this be set in the dashboard or via URL param when redirected to the apex form?

src/auth/LoginForm.tsx
```diff
@@ -187,6 +199,10 @@ const LoginForm = ({
         <JSForm {...formProps} _internalId={_internalId} />
       </div>
     );
+  } else if (apexForm) {
```
obmelvin, line 202: for my POC, I passed this prop in to LoginForm in hosted forms. for the full implementation we only want to pass it in for forms that are actually the apex form

do you think we identify that based on URL (lack of slug & feathery query params) or should a form have a flag indicating if it is the apex form?

src/integrations/stytch.ts
```diff
@@ -20,7 +20,12 @@ export function installStytch(stytchConfig: any) {
       if (authState.client) resolve(authState.client);
       else {
         const initializedClient = new StytchHeadlessClient(
-          stytchConfig.metadata.token
+          stytchConfig.metadata.token,
+          {
+            cookieOptions: {
+              availableToSubdomains: true
```
obmelvin, line 26: should this always be true now, or do we add a checkbox to the integration settings?

src/auth/LoginForm.tsx
```diff
@@ -48,6 +48,7 @@ export const authState = {
 };
 
 const LoginForm = ({
+  apexForm = false,
```
bo-dun-1, line 51: we dont need this anymore do we?

src/auth/LoginForm.tsx
```diff
@@ -187,6 +199,10 @@ const LoginForm = ({
         <JSForm {...formProps} _internalId={_internalId} />
       </div>
     );
+  } else if (initInfo().redirectCallbacks[_internalId]) {
+    // If logged in and have finished onboarding questions for an apex form, then we need to redirect back to application
+    initInfo().redirectCallbacks[_internalId]();
+    return <></>;
```
bo-dun-1, line 205: why return a fragment instead of null?

bo-dun-1 (approved):

PR #745, opened 2023-03-15
fix auth form completed race condition

obmelvin:


src/utils/client.ts
```diff
@@ -384,9 +388,15 @@ export default class Client {
...
-          ([formKey, session]) =>
-            (initState.formSessions[formKey].form_completed =
-              session.form_completed)
+          ([formKey, session]) => {
+            // Before resolving this bug where formSessions was empty, login wasn't working
+            // properly because this would silently fail and stop execution of this 'thread'
+            // - i.e. setAuthId below was never called..
+            if (!initState.formSessions[formKey])
+              initState.formSessions[formKey] = {};
+            initState.formSessions[formKey].form_completed =
```
obmelvin, line 397: Before resolving adding the guard above bug, login wasn't working properly because this would silently fail and stop execution of this 'thread' - i.e. setAuthId below was never called..

src/utils/client.ts
```diff
@@ -383,11 +384,14 @@ export default class Client {
...
-          ([formKey, session]) =>
-            (initState.formSessions[formKey].form_completed =
-              session.form_completed)
-        );
+        Object.entries<{ form_completed: boolean }>(
+          data.completed_forms
+        ).forEach(([formKey, session]) => {
+          if (!initState.formSessions[formKey])
+            initState.formSessions[formKey] = {};
+          initState.formSessions[formKey].form_completed =
```
obmelvin, line 392: Before resolving adding the guard above bug, login wasn't working properly because this would silently fail and stop execution of this 'thread' - i.e. setAuthId below was never called..

bo-dun-1 (approved):

PR #756, opened 2023-03-21
set stytch cookies manually

obmelvin:


package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "@feathery/react",
-  "version": "1.1.9",
+  "name": "@feathery-org/staging",
```
obmelvin, line 2: need to revert once I'm done with private packages

package.json
```diff
@@ -1,6 +1,6 @@
...
+  "version": "1.1.36",
```
obmelvin, line 3: should probably use alpha/beta tags in the future for something like this. wasn't too concerned since we don't need to use the private package very often

src/integrations/utils.ts
```diff
@@ -62,8 +62,10 @@ export async function initializeIntegrations(
 
   const gtm = integs['google-tag-manager'];
   if (gtm) initializeTagManager(gtm);
-  if (integs.firebase || integs.stytch)
-    return Auth.inferLoginFromURL(featheryClient);
+  if (integs.firebase || integs.stytch) {
+    const session = await Auth.inferLoginFromURL(featheryClient);
```
bo-dun-1, line 66: nit: just return direcrtly

src/auth/LoginForm.tsx
```diff
@@ -29,6 +30,7 @@ export const AuthContext = createContext<any>(null);
 export const authState = {
   client: null as any,
   authId: '',
+  isHostedOnFeathery: false,
```
bo-dun-1, line 33: featheryHosted?

src/utils/formHelperFunctions.ts
```diff
@@ -420,6 +420,19 @@ export function remountAllForms() {
...
 
+/**
+ *
+ * @returns Url hash without the #, or '' if decodeURI fails
+ */
+export function getUrlHash() {
+  try {
+    return decodeURI(location.hash.substr(1));
+  } catch (e) {
+    console.log(e);
```
bo-dun-1, line 431: console.warn?

src/integrations/stytch.ts
```diff
@@ -84,7 +85,18 @@ export function stytchLoginOnLoad(featheryClient: any) {
 
   // If there is an existing Stytch session when a user returns to an embedded
   // Feathery form, we need to update the auth info from Feathery's side
-  if (stytchSession) return stytchSubmitAuthInfo(featheryClient);
+  if (stytchSession) {
+    try {
+      const submitInfo = () => stytchSubmitAuthInfo(featheryClient);
```
bo-dun-1, line 90: this function seems unnecessary

bo-dun-1 (approved):

PR #761, opened 2023-03-24
Stripe fix: hidden field were not working as product qty fields for purchases

ericpias:
Thought I/we tested this originally, but apparently not.  Needed for the ox.work plan selector demo.

### DIFF
--- a/src/integrations/stripe.ts
+++ b/src/integrations/stripe.ts
@@ -2,6 +2,7 @@ import { loadStripe } from '@stripe/stripe-js';
 import { useState } from 'react';
 import { runningInClient, featheryDoc } from '../utils/browser';
 import { ActionData } from './utils';
+import { fieldValues } from '../utils/init';
 import { ACTION_COLLECT_PAYMENT } from '../utils/elementActions';
 const stripePromise = new Promise((resolve) => {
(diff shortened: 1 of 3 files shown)


src/integrations/stripe.ts
```diff
@@ -33,25 +34,25 @@ export function installStripe(stripeConfig: any) {
 // Returns mapping of servar key (or hidden field key) to value
 export function getFlatStripeCustomerFieldValues(
   stripeConfig: any,
-  fieldValues: any
+  _fieldValues: any
```
bo-dun-1, line 37: instead of passing fieldValues in everywhere, can you use it directly from the import?

ericpias: Doing it this way makes it convenient for test cases.  I don't actually pass it at runtime and it does use the import at runtime.  It is only passed from the test case.

bo-dun-1: i would prefer we don't structure production code around test case convenience. You can always initialize fieldValues at a global level (we do that for all other test cases)

ericpias: Changes made.

bo-dun-1 (approved):

PR #774, opened 2023-03-30
improve logout listeners

obmelvin:
* fixes issue where stytch wasn't redirecting to login step after clicking logout button. due to `setAuthId('')` being called twice, `rerenderAllForms` was being called twice before Form could re-render, so `render` wouldn't actually change and the form wouldn't execute `getNewStep`
* adds session listener for firebase, achieves parity with stytch

### DIFF
--- a/src/auth/internal/AuthIntegrationInterface.ts
+++ b/src/auth/internal/AuthIntegrationInterface.ts
@@ -86,21 +86,30 @@ function oauthRedirect(oauthType: string) {
 }
 
 function initializeAuthClientListeners() {
+  if (!authState.client) return;
+
   if (isAuthStytch()) {
     if (getStytchJwt()) setStytchDomainCookie();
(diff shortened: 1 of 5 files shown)


src/auth/internal/AuthIntegrationInterface.ts
```diff
@@ -45,14 +45,16 @@ function isThereAnExistingSession(): boolean {
 async function inferAuthLogout() {
   if (!authState.client) return;
 
+  // We don't want to setAuthId('') here because user's can also revoke the
+  // session themselves with the auth client. So that should be performed in the
+  // auth client listeners
   if (isAuthStytch()) {
     await authState.client.session.revoke();
+    authState.onLogout();
```
obmelvin, line 53: I think authState.onLogout should also be moved to the client listeners, but wanted to get your opinion

bo-dun-1 (approved):

PR #782, opened 2023-04-01
remove setAuthId guard, not needed after improvement to setRender

obmelvin:


src/auth/LoginForm.tsx
```diff
@@ -100,18 +100,16 @@ const LoginForm = ({
       history.replaceState(null, '', loginPath + window.location.search);
     }
 
+    authState._featheryHosted = _featheryHosted;
```
obmelvin, line 103: this never should have been in the setAuthId fn definition....that is completely my bad for not realizing the indentation

bo-dun-1 (approved):

PR #785, opened 2023-04-04
hideif repeat support

bo-dun-1:


ericpias (commented):
1. I set up a repeating group of fields (text field and dropdown).  Both a required.  I was able to submit without filling either.  Also tried multiple rows and was able to submit empty.  Seems wrong to me.
2. I then set up a custom validation on one of the repeating fields (isFilled).  It then seemed to validate both fields and would not allow either to be empty and submit.  This is good, but the default required field (not optional) should have been enough (i.e. item 1 above).
3. Inline validation errors do not seem to be working at all, but this may have already been broken.  I will investigate and write an asana task for this if broken in master.
4. Hide if is working nicely with repeat.

PR #789, opened 2023-04-05
Fixes repeat specific validation logic issues

ericpias:
1. Validation rules and hide rules on a repeating element must be specifically evaluated against sibling repeat elements if they are in the rule.
2. Validation/hide rules on a non-repeating element that involve repeating fields must evaluate against any/all of the repeating field's values.

2 above was not working properly prior to this fix.  I think 1 was working.

### DIFF
--- a/src/utils/hideAndRepeats.ts
+++ b/src/utils/hideAndRepeats.ts
@@ -221,6 +221,9 @@ function getVisibleElements(
   elementTypes: string[] = [],
   repeat = false
 ) {
+  const repeatGrid = step.subgrids.filter((grid: any) => grid.repeated)[0];
+  const repeatKey = repeatGrid ? getPositionKey(repeatGrid) : '';
+
   return elementTypes.flatMap((type) =>
     step[type].flatMap((el: any) => {
       const elKey = getPositionKey(el);
@@ -230,7 +233,9 @@ function getVisibleElements(
         if (flag && (repeat || !elements.length)) {
           elements.push({
             element: el,
-            repeat: index,
+            repeat: getPositionKey(el).startsWith(repeatKey)
+              ? index
+              : undefined,
             last: index === flags.length - 1
           });
         }
--- a/src/utils/validation.ts
+++ b/src/utils/validation.ts
@@ -104,7 +104,7 @@ function validateElement(
     };
     validations?: ResolvedCustomValidation[];
   },
-  repeat: number
+  repeat: any
 ): string {
   // First priority is standard validations for servar fields
   const { servar, validations } = element;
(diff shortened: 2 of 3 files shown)


src/Form/grid/Element/index.tsx
```diff
@@ -158,7 +158,8 @@ const Element = ({ node: el, form, flags }: any) => {
         if (isFieldActuallyRequired(element, activeStep)) {
           const servar = element.servar;
           let fieldVal: any = fieldValues[servar.key];
-          if (servar.repeated) fieldVal = fieldVal[repeat];
+          if (servar.repeated && repeat !== undefined)
```
bo-dun-1, line 161: doesn't servar.repeated = True imply repeated isn't undefined?

ericpias: I think so, but wanted to be explicit on the check.

src/utils/hideAndRepeats.ts
```diff
@@ -221,6 +226,8 @@ function getVisibleElements(
   elementTypes: string[] = [],
   repeat = false
 ) {
+  const [repeatKey] = getRepeatParentInfo(step);
```
bo-dun-1, line 229: this function is running a lot of unnecessary functionality here to get numRepeats, which isn't used here. I think this would be better as it was before, and you can get repeatKey manually here.

bo-dun-1, line 229: this function is running a lot of unnecessary functionality here to get numRepeats, which isn't used here. I think this would be better as it was before, and you can get repeatKey manually here.

src/utils/validation.ts
```diff
@@ -104,13 +104,13 @@ function validateElement(
...
   },
-  repeat: number
+  repeat: number | undefined
 ): string {
   // First priority is standard validations for servar fields
   const { servar, validations } = element;
   if (servar) {
     let fieldVal: any = fieldValues[servar.key];
-    if (servar.repeated) fieldVal = fieldVal[repeat];
+    if (servar.repeated && repeat !== undefined) fieldVal = fieldVal[repeat];
```
bo-dun-1, line 113: do we need both checks? in what case is servar.repeated true but repeat is undefined

src/Form/grid/Element/index.tsx
ericpias: Ok, not generally the way I like to code, but made the change.

src/utils/validation.ts
ericpias: Ok, not generally the way I like to code, but made the change.

src/utils/validation.ts
```diff
@@ -104,7 +104,7 @@ function validateElement(
     };
     validations?: ResolvedCustomValidation[];
   },
-  repeat: number
+  repeat: any
```
bo-dun-1, line 107: number | undefined seemed good to me

bo-dun-1 (approved):

src/utils/validation.ts
ericpias: Typescript complains.  If I go with that, then I need to add the undefined guard back in.

PR #790, opened 2023-04-05
fix redirect when no onboarding questions

obmelvin:


src/auth/internal/utils.ts
```diff
@@ -26,13 +26,55 @@ export function isTerminalStepAuth(
...
+  // protected steps. That means there are no onboarding questions.
+  const isOnboarding =
+    authIntegration.auth_gate_steps.length === 0 &&
+    authIntegration.login_step === '' &&
+    authIntegration.logout_step !== '';
+
+  return !isOnboarding;
+}
+
+// TODO: need to set local formCompleted so that the flag is set when verifying sms code
```
obmelvin, line 59: still need to do this

src/Form/index.tsx
```diff
@@ -1246,19 +1286,41 @@ function Form({
...
+          );
         } else {
           setElementError('A valid email is needed to send your magic link.');
           break;
         }
-      } else if (type === ACTION_OAUTH_LOGIN)
-        Auth.oauthRedirect(action.oauth_type);
-      else if (type === ACTION_LOGOUT) await Auth.inferAuthLogout();
+      } else if (type === ACTION_OAUTH_LOGIN) {
+        // TODO: should we check to see if completed is already true and short circuit this logic?
```
obmelvin, line 1316: curious on your thoughts @bo-dun-1

src/Form/index.tsx
```diff
@@ -688,19 +693,38 @@ function Form({
...
+          //   redirectUrl
+          // });
+
+          const onboardingExists = areThereOnboardingQuestions(
+            session.integrations
+          );
+          const isFormComplete = session.form_completed;
+
+          // can I drop onboardingExists?
+          if (isFormComplete && !onboardingExists && authState.authId)
```
obmelvin, line 709: should probably make this a shared fn between here + sms verify. just would need to have optional pass in of redirectUrl, since this case can't rely on the CBs

src/Form/index.tsx
```diff
@@ -688,19 +693,31 @@ function Form({
...
           }
 
-          // User is authenticating. auth hook will set the initial stepKey once auth has finished
-          if (authState.redirectAfterLogin) return;
+          const onboardingExists = areThereOnboardingQuestions(
+            session.integrations
+          );
+          const isFormComplete = session.form_completed;
+
+          // can I drop onboardingExists?
```
obmelvin, line 701: pretty sure I don't need to check if onboardingExists, because all that really matters is if the form is complete. Forms without onboarding will be marked complete during login process, forms with onboarding will be marked complete after finishing the questions

src/auth/internal/utils.ts
```diff
@@ -33,6 +33,19 @@ export function isTerminalStepAuth(
   );
 }
 
+export function areThereOnboardingQuestions(integrations: any): boolean {
```
bo-dun-1, line 36: hasOnboardingSteps?

PR #791, opened 2023-04-05
make loader a general form prop

obmelvin:


src/Form/index.tsx
```diff
@@ -257,21 +261,14 @@ function Form({
...
-        {(data as any).loader}
-      </div>
-    ) : (
-      (data as any).loader
-    );
-  }, [loaders]);
+  // When the active step changes, recalculate the dimensions of the new step
+  const stepCSS = useMemo(() => calculateStepCSS(activeStep), [activeStep]);
+
+  const { clearLoaders, loader, loaders, setLoaders } = useLoader({
```
bo-dun-1, line 267: why does loaders still need to be returned here? isn't it not used now?

src/hooks/useLoader.tsx
```diff
@@ -0,0 +1,88 @@
...
+    return (data as any).type === 'default' ? (
+      <div style={{ height: '20vh', width: '20vh' }}>
+        {(data as any).loader}
+      </div>
+    ) : (
+      (data as any).loader
+    );
+  }, [loaders]);
+
+  const loaderContainer = isShowExplicitlyFalse ? null : (
```
bo-dun-1, line 63: const showLoader = !isShowExplicitlyFalse && (_isAuthLoading || stepLoader);
const loaderContainer = <LoaderContainer showLoader={showLoader}...

src/Form/index.tsx
obmelvin: I had removed it completely and then ran into a crash because it's used here - https://github.com/feathery-org/feathery-react/blob/master/src/Form/grid/Element/index.tsx#L184

src/Form/index.tsx
```diff
@@ -1449,58 +1443,59 @@ function Form({
...
-        )}
-        {formSettings.showBrand && (
-          <Watermark
-            addChin={addChin}
-            brandPosition={formSettings.brandPosition}
-          />
-        )}
-      </BootstrapForm>
-    </ReactPortal>
+    <div style={{ position: 'relative', height: '100%', width: '100%' }}>
```
obmelvin, line 1452: will explore whether this div can be conditionally rendered only if the stepLoader is present

but obv don't want to cause remounts of the entire form - so wasn't sure we could do it. but theoretically with a key we can

bo-dun-1 (approved):

PR #799, opened 2023-04-16
Button group fix for responsive width

ericpias:


src/elements/fields/index.tsx
```diff
@@ -180,6 +180,7 @@ function applyFieldStyles(field: any, styles: any) {
         ['button_width', 'button_width_unit', 'content_responsive'],
         (a: any, b: any, c: boolean) => ({
           [c ? 'minWidth' : 'width']: `${a}${b}`,
+          maxWidth: c ? 'max-content' : '',
```
bo-dun-1, line 183: could we just revert to the previous css here? you may just be able to remove the flex property

ericpias: Ok, I had it in my head that the flex was legacy css.  However, I mast have added it with my last change.  I reverted it and the maxWidth and it seems to work fine with responsive in my testing.

bo-dun-1 (approved):

PR #816, opened 2023-05-02
Drafts phase 3 - draft preview

ericpias:


src/Form/components/DevNavBar.tsx
```diff
@@ -141,7 +147,22 @@ export default function DevNavBar({ allSteps, curStep, history }: any) {
...
         >
-          Preview
+          <span
+            css={{
+              color: '#007bff',
+              paddingRight: '8px',
+              borderRight: '2px solid #cfd4dd'
+            }}
+          >
+            {draft ? 'Draft Preview' : 'Publish Preview'}
```
bo-dun-1, line 157: Maybe `Test Form` instead of `Publish Preview`?

src/Form/index.tsx
```diff
@@ -845,7 +847,7 @@ function Form({
       return { key, [(val as any).type]: newVal };
     });
     const stepPromise =
-      featheryFields.length > 0
+      featheryFields.length > 0 && !_draft
```
bo-dun-1, line 850: maybe it would be cleaner to initialize the client with the draft boolean and check the flag on the write calls within the client functions

src/utils/client.ts
```diff
@@ -462,8 +468,11 @@ export default class Client {
       method: 'POST',
       body: JSON.stringify(data)
     };
-    if (promise) return promise.then(() => this._fetch(url, options));
-    else return this._fetch(url, options);
+    // no events for draft
+    const doRegister = () =>
```
bo-dun-1, line 472: promise = promise || Promise.resolve();
return promise.then(() => this.draft && this._fetch(...

src/utils/formHelperFunctions.ts
```diff
@@ -194,10 +194,18 @@ export const getStepDepthMap = (steps: any, hasProgressBar = false) => {
 
     const incr = missingBar ? 0 : 1;
     step.next_conditions.forEach((condition: any) => {
-      stepQueue.push([steps[condition.next_step_key], depth + incr]);
+      const step = condition.next_step_key
```
bo-dun-1, line 197: why is this check necessary?

src/Form/index.tsx
```diff
@@ -878,71 +880,74 @@ function Form({
...
-      servar: pm ? pm.servar : null,
-      client,
-      updateFieldValues,
-      integrationData: integrations?.stripe,
-      targetElement: pm ? getCardElement(pm.servar.key) : null
-    });
-    if (errors) {
-      const { errorMessage, errorField } = errors;
-      await setFormElementError({
+    if (!_draft) {
```
bo-dun-1, line 904: why does this draft check need to be here? even if we can't trigger integrations, can't we fetch the integrations for the draft form?

ericpias: I don't understand the question.  I think we are in agreement that we don't want to trigger integrations.  For Stripe this means making the card setup and the collect payment inert.  That is what I am doing with this check.  What do you mean by "fetch the integrations"?

bo-dun-1 (approved):

PR #844, opened 2023-05-21
Stripe 7: Redesigned product selection and payment

ericpias:


src/Form/index.tsx
```diff
@@ -913,16 +925,16 @@ function Form({
...
     if (!_draft)
       for (let i = 0; i < activeStep.servar_fields.length; i++) {
-        const servar = activeStep.servar_fields[i].servar;
-        if (servar.type === 'payment_method') {
+        const field = activeStep.servar_fields[i];
+        if (field.servar.type === 'payment_method') {
           const integrationData = integrations?.stripe;
           const actionData: ActionData = {
-            servar,
+            servar: field,
```
bo-dun-1, line 932: why is a field being passed as the servar? if we need a field, then we should just call it a field

src/Form/index.tsx
```diff
@@ -564,6 +568,14 @@ function Form({
       updateFieldOptions
     };
 
+    // This could be a redirect from Stripe following a successful payment checkout
+    checkForPaymentCheckoutCompletion(
```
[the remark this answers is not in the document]
ericpias: Ok, moved it up a few lines to where I think you suggested.

src/utils/logic.ts
```diff
@@ -149,16 +149,23 @@ const COMPARISON_FUNCTIONS: {
...
         l,
         r
-      )
-    ),
-  selections_dont_include: (l, r) =>
-    l.every((l: any) =>
+      );
+    return l.some((l: any) => {
+      if (l && detectType(l) === 'object')
+        // Special behavior for left side objects.  Their keys will be compared to the right side
```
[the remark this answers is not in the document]
ericpias: Added this comment:
// While generally useful, this is specifically for the case where the left side
        // is the feathery.payments.selections field which is an object with keys that are the
        // product ids selected for purchase.

src/Form/index.tsx
```diff
@@ -900,6 +914,13 @@ function Form({
       // need to include value === '' so that we can clear out hidden fields
       if (value !== undefined) hiddenFields[fieldKey] = value;
     });
+    // submit feathery reserved hidden fields
+    fieldValues[FEATHERY_PAYMENTS_SELECTIONS] &&
```
bo-dun-1, line 918: could we put these into standard if statements? it's a little hard to read in this format

src/integrations/stripe.ts
```diff
@@ -271,6 +338,64 @@ export async function collectPayment(
...
+  if (queryParams.get('_feathery_paid') === 'true') {
+    // remove the _feathery_paid parameter from the url
+    queryParams.delete('_feathery_paid');
+    const { origin, pathname, hash } = featheryWindow().location;
+    const newUrl = `${origin}${pathname}?${queryParams}${hash}`;
+    featheryWindow().history.replaceState({}, '', newUrl);
+
+    // search all step buttons and subgrids for the purchase action and complete the payment
+    const paymentElement = [...step.buttons, ...step.subgrids].find((element) =>
+      element.properties?.actions?.find(
```
bo-dun-1, line 358: no need for elvis operator on properties (properties?.)

bo-dun-1 (approved):

PR #888, opened 2023-06-15
update stytch. add error msgs for sms, email, oauth

obmelvin:


src/integrations/stytch.ts
```diff
@@ -115,7 +124,12 @@ export async function stytchLoginOnLoad(featheryClient: any) {
     return authFn()
       .then(() => stytchSubmitAuthInfo(featheryClient))
       .catch((e: any) => {
-        authState.showError();
+        const type = e.error_type;
+        let errorMsg;
+        if (type === 'bad_domain_for_stytch_sdk')
```
bo-dun-1, line 129: could handleLoginOrCreateErrors be used here too? And what if we just call it handleStytchLoginErrors

src/Form/index.tsx
```diff
@@ -1293,8 +1297,12 @@ function Form({
       } else if (type === ACTION_SEND_MAGIC_LINK) {
         const email = fieldValues[action.auth_target_field_key] as string;
         if (validators.email(email)) {
-          // Don't block to make potential subsequent navigation snappy
-          Auth.sendMagicLink(email);
+          try {
+            await Auth.sendMagicLink(email, setElementError);
+          } catch (e) {
+            elementClicks[id] = false;
```
bo-dun-1, line 1303: do we not want to set a user facing error in this case?

src/integrations/stytch.ts
obmelvin: hm, yeah I guess it probably makes sense to use it here too even though some of the error types can't happen for oauth, since I'm already using that fact to share the same fn between sms and email

obmelvin: oh, I could use the same fn - but oauth doesn't error until after the redirect, which is why I had to update the LoginError component & why this is different

bo-dun-1 (approved):

PR #901, opened 2023-06-21
save url params

obmelvin:


src/Form/utils.ts
```diff
@@ -0,0 +1,28 @@
+import { mapUrlParamsToHiddenFields } from '../utils/browser';
+import Client from '../utils/client';
+
+export function saveUrlParams(
```
bo-dun-1, line 4: this utils file seems to serve the same purpose as formHelperFunctions, should probably be merged into one file

src/Form/index.tsx
```diff
@@ -63,7 +63,12 @@ import DevNavBar from './components/DevNavBar';
 import FeatherySpinner from '../elements/components/Spinner';
 import { isObjectEmpty } from '../utils/primitives';
 import CallbackQueue from '../utils/callbackQueue';
-import { featheryWindow, openTab, runningInClient } from '../utils/browser';
+import {
+  featheryWindow,
+  mapUrlParamsToHiddenFields,
```
bo-dun-1, line 68: this function doesn't appear to be used in this file

src/utils/browser.ts
```diff
@@ -39,3 +39,14 @@ export function getStytchJwt() {
 
 export const openTab = (url: any) =>
   featheryWindow().open(url, '_blank', 'noopener');
+
+// maps url query params to an object, used to set hidden fields
+export const mapUrlParamsToHiddenFields = () => {
```
bo-dun-1, line 44: can this function just be unified into the single place where it's used in saveUrlParams?

bo-dun-1: also shoudl hosted forms be updated to use this functionality instead?

src/utils/browser.ts
```diff
@@ -39,3 +39,14 @@ export function getStytchJwt() {
...
+  const params = new URLSearchParams(featheryWindow().location.search);
+  const hiddenFields: any = {};
+  params.forEach((value, key) => {
+    if (key === '_slug') return;
```
obmelvin, line 48: Also discard feathery_ params

src/utils/formHelperFunctions.ts
```diff
@@ -504,3 +506,34 @@ export function isStepTerminal(step: any) {
...
 }
+
+export function saveUrlParams(
+  updateFieldValues: (newFieldValues: any, rerender?: boolean) => boolean,
+  client: Client
+) {
+  const params = new URLSearchParams(featheryWindow().location.search);
+  const hiddenFields: any = {};
+  params.forEach((value, key) => {
+    if (['feathery_1', 'feathery_2', '_slug'].includes(key)) return;
```
bo-dun-1, line 517: i think you can ignore feathery_1 and feathery-2 now, it's been long enough and i've removed that handling logic form hosted forms too

src/utils/formHelperFunctions.ts
```diff
@@ -504,3 +506,34 @@ export function isStepTerminal(step: any) {
...
+    hiddenFields[key] = value;
+  });
+  if (Object.getOwnPropertyNames(hiddenFields).length === 0) return;
```
bo-dun-1, line 520: what is this check for? and if we want to short circuit it, could we just short circuit by checking params above?

bo-dun-1 (approved):

src/utils/formHelperFunctions.ts
obmelvin: I wanted to make sure that the stytch query params didn't get saved

src/utils/formHelperFunctions.ts
obmelvin: hosted forms can have the _slug query param so I only wanted to proceed if there were other query params

this seems like the most direct way to do that, rather than needing a separate check for _slug (and the others) above

bo-dun-1: ah that makes sense

src/utils/formHelperFunctions.ts
```diff
@@ -504,3 +506,39 @@ export function isStepTerminal(step: any) {
...
   return !hasNext;
 }
+
+export function saveUrlParams(
+  updateFieldValues: (newFieldValues: any, rerender?: boolean) => boolean,
+  client: Client
+) {
+  // If there is only one query param and it is the slug, don't save it
+  if (
+    featheryWindow().location.search.startsWith('?_slug=') &&
```
bo-dun-1, line 516: couldn't we do this check on the params object?

bo-dun-1 (approved):

PR #903, opened 2023-06-22
fix: Fix percentage heights on elements

ondrosh:


src/Form/grid/StyledContainer/index.tsx
```diff
@@ -74,7 +76,9 @@ export const StyledContainer = forwardRef<HTMLDivElement, StyledContainerProps>(
         className={classNames('styled-container', type, className)}
         {...props}
       >
-        {children}
+        <div className='inner-container' css={innerStyles}>
```
bo-dun-1, line 79: why are inner containers necessary? can you leave comments?

bo-dun-1 (approved):

PR #908, opened 2023-06-23
fix rerender when form complete

obmelvin:


src/utils/formHelperFunctions.ts
```diff
@@ -486,7 +486,7 @@ export function getServarTypeMap(steps: any) {
 }
 
 export function isStepTerminal(step: any) {
-  // If step is navigatable to another step, it's not terminal
+  // If step is navigable to another step, it's not terminal
```
obmelvin, line 489: just changed because the previous word was underlined by spell checkeer

bo-dun-1 (approved):

PR #911, opened 2023-06-26
callback function api refactor

bo-dun-1:


src/Form/grid/Element/index.tsx
```diff
@@ -9,7 +9,7 @@ import { isObjectEmpty, stringifyWithNull } from '../../../utils/primitives';
 import { isFieldValueEmpty } from '../../../utils/validation';
 import { justInsert, justRemove } from '../../../utils/array';
 import { fieldValues } from '../../../utils/init';
-import { ACTION_STORE_FIELD } from '../../../utils/elementActions';
+import { ACTION_NEXT, ACTION_STORE_FIELD } from '../../../utils/elementActions';
```
ericpias, line 12: ACTION_NEXT not used here.

src/Form/index.tsx
```diff
@@ -1172,20 +1121,15 @@ function Form({
...
-          fieldOnChange({
-            fieldIds: data.fieldIDs,
-            fieldKeys: data.fieldKeys,
-            elementRepeatIndex: element.repeat
-          })();
-        }
-      } else if (type === ACTION_TRIGGER_PLAID) {
+      await runUserCallback(onAction, () => ({
+        trigger,
+        action: type
```
ericpias, line 1126: This adds 'action' to the callback context but it is not in ContextOnAction (Form.ts).  It should be, no?

src/utils/formHelperFunctions.ts
```diff
@@ -145,7 +145,11 @@ export const getAllElements = (step: any) => {
   ];
 };
 
-export const lookUpTrigger = (step: any, elementID: any, elementType: any) => {
+export const lookUpTrigger = (
+  step: any,
+  elementID: any,
+  elementType: any
+): Record<string, any> => {
```
ericpias, line 152: The return object for this function has the 'elementType' prop instead of the 'type' prop specified in the 'Trigger' type defined in Form.ts.  This return is used as the trigger that is part of ContextOnAction.  This mismatch seems to be an old issue though because it is also an issue with **OnError and onSubmit**.

src/types/Form.ts
```diff
@@ -35,55 +30,31 @@ export type FieldData = {
...
+  trigger: Trigger;
+}
+
 export interface ContextOnLoad extends FormContext {
-  fields: FieldData;
-  stepName: string;
-  previousStepName: string;
-  lastStep: boolean;
-  setStep: SetStep;
   firstStepLoaded: boolean;
```
ericpias, line 42: Is firstStepLoaded ever set anywhere?

bo-dun-1: nope, removed

src/Form/index.tsx
bo-dun-1: yes, thank you for catching

src/utils/formHelperFunctions.ts
bo-dun-1: changed elementType to type here

ericpias (approved):

PR #916, opened 2023-06-28
reset state for popup forms when tracking off, fix z index bug and is…

obmelvin:
fixes
* popup forms will now reset the field values, step and user ID when the popup is closed & tracking is off
* fixes a z-index issue with phone field & popup forms
* fixes an issue where react-bootstrap's overlay fn was causing the page to scroll to the top when clicking to open the phone country dropdown. replaces it with ElevatedMenu from the dashboard, which resolved the issue

### DIFF
--- a/src/Form/index.tsx
+++ b/src/Form/index.tsx
@@ -43,7 +43,7 @@ import {
   getVisiblePositions
 } from '../utils/hideAndRepeats';
 import { validators, validateElements } from '../utils/validation';
-import { initState, fieldValues, FieldValues } from '../utils/init';
+import { initState, fieldValues, FieldValues, initInfo } from '../utils/init';
 import { isEmptyArray, justInsert, justRemove } from '../utils/array';
 import Client from '../utils/client';
 import { useFirebaseRecaptcha } from '../integrations/firebase';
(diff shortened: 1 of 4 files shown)


src/Form/index.tsx
```diff
@@ -329,6 +332,39 @@ function Form({
...
+      history.replace(location.pathname + location.search);
+      const newKey = getInitialStep({
+        initialStepId,
+        steps,
+        sessionCurrentStep: session.current_step_key
+      });
+      // if I set this, then the hash shows while the form is closed. If I don't
+      // set this, the hash stays cleared until the form is both reopened &
+      // navigated to a new step to another step
+      // setUrlStepHash(history, steps, newKey);
```
obmelvin, line 363: is it fine if the hash doesn't show up until navigating to the 2nd step in this case? happy to adjust this any way - but wasn't sure what you'd expect the bx to be

src/elements/fields/PhoneField/index.tsx
```diff
@@ -197,42 +199,44 @@ function PhoneField({
             '&:hover': { backgroundColor: '#e6e6e633' }
           }}
           ref={triggerRef}
-          onClick={() => setShow(!show)}
+          onClick={(e: any) => {
+            // TODO: position should be set staticly based on the input, not on the click. but refs were all null?
+            setMenuPosition({
```
obmelvin, line 204: small change to make here. I forgot that I commented out `ref={triggerRef}` 🙈 when I was trying to do this originally, so pretty sure that was my issue

PR #930, opened 2023-07-10
Run user-provided snippets

bo-dun-1:


src/utils/formContext.ts
```diff
@@ -47,6 +47,7 @@ export const getFormContext = (formUuid: string) => ({
       : {};
     return {
       totalSteps: Object.keys(state.steps).length,
+      stepId: step.id ?? '',
```
bo-dun-1, line 50: why is stepId being returned in the form context? this should be an internal identifier

ericpias: The steps filter on logicRule is an array of step ids.  Step ids are better than names/keys which can change.  So I need the step id here.  We need not document it so it stays internal.  See runUserLogic where I filter by any steps and use this.

src/Form/index.tsx
```diff
@@ -589,6 +677,24 @@ function Form({
             };
           if (res.save_url_params) saveUrlParamsFormSetting = true;
           setFormSettings(mapFormSettingsResponse(res));
+          setLogicRules(res.logic_rules);
+
+          // Add any logic_rule.elements to viewElements so that onView called for then too.
+          // Make sure there are no duplicate entries.
+          if (res.logic_rules) {
+            const newViewElements: string[] = [...viewElements];
```
bo-dun-1, line 685: IMO we should handle this viewElements aggregation on the backend before returning to frontend

src/Form/index.tsx
```diff
@@ -455,25 +476,92 @@ function Form({
...
     }
+    // filter the logicRules that have trigger_event matching the trigger event (type_)
+    if (logicRules) {
+      const logicRulesForEvent = logicRules.filter(
+        (logicRule: any) => logicRule.trigger_event === event
+      );
+      // Run the logic rules in sequence!
+      for (const logicRule of logicRulesForEvent) {
+        // don't run disabled or invalid rules
+        if (!logicRule.enabled || !logicRule.valid || !logicRule.code) continue;
```
bo-dun-1, line 520: we should filter out disabled and invalid rules on the backend so we don't need this check or these flags here

src/Form/index.tsx
ericpias: This is aggregating any view rule elements with any onViewElements that might have been passed as a form prop (embedded situation).  Doing this aggregation on the BE means that onViewElements would need to be sent to the BE via a new end point.  Unless I am missing something, that would not make sense.

src/utils/formContext.ts
ericpias: Ok, made this change

ericpias (approved):

PR #952, opened 2023-07-26
Stripe 8 - Product/Price Cache, Pricing Model and Cost Calculation

ericpias:


src/integrations/stripe.ts
```diff
@@ -417,3 +526,155 @@ export function usePayments(): [
...
+export interface Product {
+  id: string;
+  default_price_id: string;
+  prices: Price[];
+}
+
+interface ProductPriceCacheConfig {
+  product_price_cache: { [key: string]: Product };
+}
+export function calculateSelectedProductsTotal(
```
bo-dun-1, line 558: i think this type of hefty calculation logic should live on the backend, and we can issue a request when we need a calculated total. react lib is insecure and hard to update

ericpias: Ok, but I stated really clearly in the spec document that these calculations would be in the FE.  I even put it in bold.  See: https://docs.google.com/document/d/1ADF3b3u-u008Jyhws6g9DT9VEdYzLkQFLNltiLQAhAs/edit#heading=h.17h2et7tytnr.  I am not necessarily arguing your point, but I will point out that quite a few hours were spent writing and testing this code that is essentially throw-away.  There was a break-down in the review process here that is causing inefficiency.

ericpias: As we discussed, this will be deferred to a later phase.

bo-dun-1 (approved):

PR #954, opened 2023-07-31
Button validation errors will only be triggered if the button is the …

ericpias:
…trigger

### DIFF
--- a/src/Form/index.tsx
+++ b/src/Form/index.tsx
@@ -76,7 +76,8 @@ import {
   ContextOnView,
   ElementProps,
   PopupOptions,
-  ContextOnAction
+  ContextOnAction,
+  Trigger
 } from '../types/Form';
 import usePrevious from '../hooks/usePrevious';
 import ReactPortal from './components/ReactPortal';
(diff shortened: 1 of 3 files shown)


src/Form/index.tsx
```diff
@@ -1182,8 +1185,6 @@ function Form({
     if (id && elementClicks[id]) return;
     elementClicks[id] = true;
 
-    if (Object.keys(inlineErrors).length > 0) setInlineErrors({});
```
ericpias, line 1185: This line was causing inline errors to be cleared simply by clicking anywhere on the form.  This was a change in behavior from what I remember.  Not sure of the original purpose of this line, but I did not notice any side effects by removing it.

bo-dun-1: The intention here is so next time a button is clicked, the inline errors should be reevaluated and updated

ericpias: This re-validation happening.  In fact, after the first submit attempt, fields are auto-validated on any field change so that as things are fixed, inline errors are cleanup up.  Also, if a button has submit and validate checked, the validation runs again on any later clicks and updates the errors.  

So based on this, this code was causing harm and not needed anyway.

bo-dun-1 (approved):

PR #1002, opened 2023-09-06
Refactor rule (trigger) elements for change event to use servar Ids instead of keys

ericpias:


src/types/Form.ts
```diff
@@ -14,6 +14,7 @@ export interface PopupOptions {
 
 export type Trigger = {
   id: string;
+  servarId?: string;
```
bo-dun-1, line 17: should we name this _servarId if we want it to be private?

bo-dun-1 (approved):

PR #1010, opened 2023-09-15
Hide rules are now show/hide rules and have a related show property

ericpias:


src/utils/__test__/formHelperFunctions.spec.js
```diff
@@ -37,7 +37,8 @@ describe('formHelperFunctions', () => {
               name: 'Name 1',
               metadata: {}
             },
-            hide_ifs: []
+            hide_ifs: [],
+            show: false
```
bo-dun-1, line 41: i left a coment on jay's PR as well -- IMO show is a little generic, could we maybe call it like show_logic?

bo-dun-1 (approved):

PR #1014, opened 2023-09-20
Implementing a fields obj collection in sdk context and in rule feathery context

ericpias:


src/utils/formHelperFunctions.ts
```diff
@@ -90,6 +91,123 @@ export const formatAllFormFields = (steps: any, forUser = false) => {
...
+export const getAllFields = (fieldKeys: string[]): Record<string, Field> => {
+  const fields: Record<string, Field> = {};
+  fieldKeys.forEach((key) => {
+    const newField = new Field(key);
+    fields[key] = newField;
+  });
+  return fields;
+};
+
+export function isValidFieldIdentifier(str: string) {
```
bo-dun-1, line 103: would it be easier to check if an identifier is present within our list of field ids rather than checking that it is *not* every possible global keyword? which seems brittle / edge-casey

ericpias: The issue is that the form designer could assign a field id that collides with a javascript reserved word.  They will get a validation error should they try to use it in a rule.  However, even if they do not use it in a rule, the runtime injects that field and this causes an exception at runtime due to the reserved word being used.  So to keep things robust we need to avoid injecting fields with reserved word ids/keys.  Yes it is an edge case but if someone were to hit it, then it would be a big time waster.  So I really think this check against reserved words needs to remain in place.

bo-dun-1: can we add a comment explaining why we're doing this check then?

ericpias: sure

src/utils/init.ts
```diff
@@ -174,25 +178,20 @@ async function updateUserId(newUserId: string, merge = false): Promise<void> {
...
-  return val instanceof File ? Promise.resolve(val) : val;
-}
-
 /**
  * If customers provide files through setFieldValues
  * we need to explicitly convert any files to file Promises
  * since they may not have done so
  */
 function setFieldValues(userVals: FieldValues, rerender = true): void {
+  console.warn(
```
bo-dun-1, line 187: isn't this still undeprecated at the global level?

src/utils/Field.ts
```diff
@@ -0,0 +1,311 @@
...
+      }, null);
+    }
+    return this._sourceField;
+  }
+
+  // read-only props that are only relevant to form fields
+  get type(): string {
+    const { type, onThisForm } = this._getFormSpecificProps();
+    if (!onThisForm)
+      console.warn(
```
bo-dun-1, line 109: should these be console.error instead?

src/utils/init.ts
ericpias: Yes. Good catch.  Forgot to remove it there.

src/utils/formContext.ts
```diff
@@ -26,7 +39,11 @@ export const getFormContext = (formUuid: string) => ({
       completed: flag
     });
   },
+  // @deprecated
```
bo-dun-1, line 42: can we add todos on all of the deprecated code to remove them?

src/utils/Field.ts
ericpias: Per our discussion, leaving it as warnings.  console.error is a bit too in your face.  It expands with a stack trace.  It could be a bit annoying for someone using console to debug a logic rule, etc.

bo-dun-1 (approved):

PR #1039, opened 2023-10-04
Filtering unhandled errors at the global level and only passing to sentry if not logic rule error.  Recaptcha is a TODO.

ericpias:


src/Form/index.tsx
```diff
@@ -624,20 +624,20 @@ function Form({
...
-              e
-            );
+              // );
+            ).catch((e: any) => {
+              // catch unhandled rejections in async user code (if a promise is returned)
+              // handle any errors in async code that actually returns a promise
+              handleRuleError(e.message, logicRule);
+            });
+
+            await result;
```
bo-dun-1, line 637: why was the await pulled out to here?

bo-dun-1 (approved):

src/Form/index.tsx
ericpias: Yeah this is left-over from when I was trying several different things.  Let me put it back and retest that.

PR #1048, opened 2023-10-11
Stripe Phase 9 - Product Selector and Cart DX improvements

ericpias:


src/utils/primitives.ts
```diff
@@ -57,9 +57,22 @@ function filterKeys(obj: any, allowedKeys: any) {
...
+    maximumFractionDigits: decimalDigits
+  }).format(number);
+}
 function formatMoneyUSD(number: number, locales = ['en-US']) {
   return formatNumeric(number, { style: 'currency', currency: 'USD' }, locales);
 }
+function formatMoney(number: number, currency = 'USD', locales?: any) {
+  return formatNumeric(
+    number,
+    { style: 'currency', currency: currency },
```
bo-dun-1, line 72: nit: no need for currency: currency, just currency

src/integrations/stripe/stripe.ts
```diff
@@ -241,9 +353,17 @@ export function addToCart(
...
+  stripeConfig: any,
+  productionEnv: boolean
 ) {
-  const { product_id: productId, clear_cart: clearCart } = paymentAction;
+  const { product_id: configuredProductId, clear_cart: clearCart } =
+    paymentAction;
+  const productId = getLiveOrTestProduct(
+    configuredProductId,
+    stripeConfig,
+    productionEnv
```
bo-dun-1, line 364: could we set productionEnv into initState so you can import it directly everywhere instead of needing to pass it? the productionEnv should be consistent for all forms in teh current session anyways

src/elements/components/TextNodes.tsx
```diff
@@ -82,56 +84,87 @@ function TextNodes({
...
-                onClick = () => openTab(link);
-                cursor = 'pointer';
-              } else if (
-                attrs.fullSpan ||
-                (isNum(attrs.start) && isNum(attrs.end))
-              ) {
-                onClick = () => textSpanOnClick(attrs.start, attrs.end);
-                cursor = 'pointer';
+        {element.properties.text_mode === 'data' && textFromData !== null ? (
+          <span
```
bo-dun-1, line 111: this span is basically the same as the span on line 152, except it's missing the onClick & cursor style (which i think is a bug). We can probably define these spans as the same component

src/elements/components/TextNodes.tsx
```diff
@@ -82,56 +84,87 @@ function TextNodes({
...
-                );
-                onClick = () => openTab(link);
-                cursor = 'pointer';
-              } else if (
-                attrs.fullSpan ||
-                (isNum(attrs.start) && isNum(attrs.end))
-              ) {
-                onClick = () => textSpanOnClick(attrs.start, attrs.end);
-                cursor = 'pointer';
+        {element.properties.text_mode === 'data' && textFromData !== null ? (
```
bo-dun-1, line 110: can we move this ternary before the return statement? IMO it's getting a little too complex to represent cleanly inline with the return

src/elements/components/TextNodes.tsx
```diff
@@ -82,56 +84,87 @@ function TextNodes({
       });
     } else if (actions.length > 0) delta = applyNewDelta(delta);
 
+    // If text_mode property is set to 'data', then we don't want to render the text_formatted
+    // property, instead we the text from the data element specified in the text_source property
+    let textFromData = null;
```
bo-dun-1, line 89: lines 89 - 93 seem like they could be moved inside of the if statement on line 95

src/elements/components/TextNodes.tsx
ericpias: Cleaned it up a bit to make a little clearer.  TBH, a bit unsure exactly what you wanted and did not want to repeat the outer span.

src/elements/components/TextNodes.tsx
```diff
@@ -82,56 +107,81 @@ function TextNodes({
...
+        textSource = textSource.replace('feathery', '$');
+      try {
+        textFromData = jsonpath.value(featheryContext, textSource);
+      } catch (e) {
+        // Ignoring errors due to things like unset text_source, deleted products etc.
+      }
+    }
+    const textIsFromData =
+      element.properties.text_mode === 'data' && textFromData !== null;
+    console.log('textFromData', textFromData);
```
bo-dun-1, line 127: remove console logs please

ericpias: done

bo-dun-1 (approved):

PR #1051, opened 2023-10-12
fix: Fit containers having incorrect width after changing viewport

ondrosh:
## Changes

- Listen for transitions ending on the canvas before resizing fit containers on the editor

### DIFF
--- a/src/Form/grid/StyledContainer/hooks.ts
+++ b/src/Form/grid/StyledContainer/hooks.ts
@@ -6,7 +6,8 @@ import {
   getStyle,
   isFitContainer,
   isFitElement,
-  resizeFitContainer
+  resizeFitContainer,
+  whichTransitionEvent
 } from './utils';
 import { formatContainerStyles } from './transform';
 import {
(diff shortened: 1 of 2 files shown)


src/Form/grid/StyledContainer/utils.ts
```diff
@@ -222,3 +223,20 @@ export const resizeFitContainer = (div: any) => {
...
   div.style.width = `calc(100% - ${totalMargin}px)`;
 };
+
+export const whichTransitionEvent = () => {
+  let t;
+  const el = featheryDoc().createElement('fakeelement');
+  const transitions: any = {
+    transition: 'transitionend',
+    OTransition: 'oTransitionEnd',
+    MozTransition: 'transitionend',
```
bo-dun-1, line 233: some ends here have the e capitalized and some dont, is that intentional?

src/Form/grid/StyledContainer/hooks.ts
```diff
@@ -197,16 +214,19 @@ export const useContainerEngine = (node: any, rawNode: any, ref: any) => {
...
-
-      resizeFitContainer(div);
-
-      if (parentFitContainers) {
-        parentFitContainers.collapse();
+      resizeCurrentFitContainer();
+
+      // For the editor, we need to also resize after transitions finish (between viewport changes)
+      if (node.uuid) {
+        canvas = document.querySelector('div[data-testid="editor-canvas"]');
```
bo-dun-1, line 221: is there any way all of hte editor-related logic can be passed in from the editor via a prop? maybe have the editor pass in a hook or something so we don't create editor dependencies in the react library (e.g. assuming the editor-canvas attribute here)

ondrosh: I feel like it would be difficult to do due to some of the modifications of the editor vs. hosted form containers being so deep within logic. It's just slight changes.

I can start thinking about how we can navigate towards this approach if you want though.

bo-dun-1 (approved):

PR #1053, opened 2023-10-13
Impl priority order for actions + fix stripe bug

ericpias:


src/Form/index.tsx
```diff
@@ -1335,6 +1319,22 @@ function Form({
     clearLoaders();
   };
 
+  // Reorders by leaving non-execution order actions in place and moving actons with specific
+  // exceution orders to the end.  Non-priority order actions are
+  // essenially order 0 (before others).
+  const prioritizeActions = (actions: any[]) => {
```
bo-dun-1, line 1325: can we move this into formHelperFunctions? in general we should put any code that doesn't need to be in the Form component somewhere else since it's quite large already

bo-dun-1: or into the elementActions file might be better

src/integrations/stripe.ts
```diff
@@ -368,10 +368,18 @@ export async function checkForPaymentCheckoutCompletion(
     featheryWindow().history.replaceState({}, '', newUrl);
 
     // search all step buttons and subgrids for the purchase action and complete the payment
-    const paymentElement = [...step.buttons, ...step.subgrids].find((element) =>
-      element.properties.actions?.find(
-        (action: any) => action.type === ACTION_PURCHASE_PRODUCTS
-      )
+    const paymentElement: any = Object.values(steps).reduce(
```
bo-dun-1, line 371: why does this need to be a reduce? versus .find

bo-dun-1 (approved):

src/integrations/stripe.ts
ericpias: Find on steps would not work since what this bit of code does is find particular element in one of the steps.  Reduce seems appropriate to me.

PR #1074, opened 2023-11-03
feat: Allow typing signatures

ondrosh:
## Changes

- Update the signature field to allow typing or drawing signatures


src/elements/fields/SignatureField/components/SignatureModal.tsx
```diff
@@ -0,0 +1,375 @@
...
+                  margin: 0,
+                  padding: 0
+                }}
+              >
+                Use your cursor or trackpad to draw your signature in the box
+                below.
+              </p>
+              <div
+                css={{ position: 'relative', width: '100%', height: '200px' }}
+              >
```
ericpias, line 297: It would be nice to scale the image canvas so that for reasonably sized signature fields, the resulting signature image extends over the whole field undistorted.  I did this on the original signature dialog I put in place last week.  There are limits of course.  For really wide signature fields (on the form canvas), have some max width to scale to so the canvas on the dialog is reasonably sized and signable.

src/elements/fields/SignatureField/components/SignatureModal.tsx
```diff
@@ -0,0 +1,375 @@
...
+                  css={{
+                    position: 'relative',
+                    width: '100%',
+                    height: '100px',
+                    backgroundColor: '#f6f6f6',
+                    display: 'flex',
+                    alignItems: 'center',
+                    justifyContent: 'center',
+                    fontSize: '2em',
+                    fontFamily: 'cursive'
```
ericpias, line 174: Overflow hidden.  Edge case: Long names can overflow the box.  
image

src/elements/fields/SignatureField/components/SignatureModal.tsx
```diff
@@ -0,0 +1,375 @@
...
+    []
+  );
+
+  const handleCancel = () => {
+    setShow(false);
+    resetState();
+  };
+
+  const handleSubmit = () => {
+    onEnd(signatureFile);
```
ericpias, line 66: Edge case: With the old signature control, it was possible to unsign in that the clear icon would appear on hover and you could remove the signature.  Clearing would clear the value of the signature field.  It seems like that is not present here and there is no way to do it.  I think we need it.  Should be as simple as clicking sign with no name entered or no signature, clears the value.

ericpias: And maybe the Sign label become "Unsign" until something is entered?

src/elements/fields/SignatureField/components/SignatureModal.tsx
```diff
@@ -0,0 +1,395 @@
...
+                gap: '15px'
+              }}
+            >
+              <p
+                css={{
+                  margin: 0,
+                  padding: 0
+                }}
+              >
+                Use your cursor or trackpad to draw your signature in the box
```
bo-dun-1, line 295: Maybe just "Draw your signature in the box below"?

bo-dun-1 (approved):

PR #1088, opened 2023-11-27
Stripe: Extend API for Add/Remove cart actions

ericpias:


src/integrations/stripe/Cart.ts
```diff
@@ -121,4 +128,31 @@ export default class Cart {
     const total = (fieldValues[FEATHERY_CART_TOTAL] as number) ?? 0;
     return currency ? formatMoney(total, currency) : formatDecimal(total, 2);
   }
+
+  addProductToCart(productId: string, quantity: number, overwrite = true) {
```
bo-dun-1, line 132: replace instead of overwrite everywhere

bo-dun-1 (approved):

PR #1094, opened 2023-11-28
feat: Add API Connector utilities

ondrosh:
## Changes

- Add `connectToAPI` method on `feathery.http` to utilize new API Connectors built by the no-code rule builder

### DIFF
--- a/src/utils/client.ts
+++ b/src/utils/client.ts
@@ -514,35 +514,34 @@ export default class Client {
 
   // Logic custom APIs
   runCustomRequest(
-    method: string,
-    url: string,
-    userData: Record<string, any> | any[],
-    headers: Record<string, string>
+    payload:
+      | string
+      | {
+          method: string;
+          url: string;
+          userData: Record<string, any> | any[];
+          headers: Record<string, string>;
+        }
   ) {
     const { userId } = initInfo();
     const data = {
       fuser_key: userId,
-      form_key: this.formKey,
-      method,
-      url,
-      user_data: userData,
-      headers
+      form_key: this.formKey
     };
-    return this._fetch(`${API_URL}custom_request/`, {
-      headers: { 'Content-Type': 'application/json' },
-      method: 'POST',
-      body: JSON.stringify(data)
-    }).then((response) => (response ? response.json() : Promise.resolve()));
-  }
 
-  async runCustomRequestByName(name: string) {
-    const { userId } = initInfo();
-    const data = {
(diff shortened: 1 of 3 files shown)


src/utils/formHelperFunctions.ts
```diff
@@ -820,5 +826,14 @@ export function httpHelpers(client: any) {
         headers: Record<string, string>
       ) => client.runCustomRequest(method, url, data, headers))
   );
+
+  helpers.connectToAPI = async (name: string) => {
```
bo-dun-1, line 830: should this actually just be integrated into the runCustomRequest interface? maybe the url parameter can be polymorphic and also support `name`? This way we have just one interface and users can still pass in custom data and headers to adjust the api connector if they want

ondrosh: Yeah, I can do that -- ideally we move completely away from the older payload all together in the future.

src/utils/client.ts
```diff
@@ -530,20 +530,39 @@ export default class Client {
...
-      form_key: this.formKey,
-      method,
-      url,
-      user_data: userData,
-      headers
+      form_key: this.formKey
     };
+
+    if (typeof payload === 'string') {
+      Object.assign(data, {
```
bo-dun-1, line 550: data.name = payload;

src/utils/client.ts
```diff
@@ -530,20 +530,39 @@ export default class Client {
...
+        name: payload
+      });
+    } else {
+      Object.assign(data, {
```
bo-dun-1, line 554: why not just assign one-by-one?

data.method = payload.method;
data.url = payload.url;

etc.

src/utils/formHelperFunctions.ts
bo-dun-1: i think we were going to rename this to connect as well

src/utils/formHelperFunctions.ts
```diff
@@ -811,15 +817,45 @@ export function mapFormSettingsResponse(res: any) {
   };
 }
 
-export function httpHelpers(client: any) {
+export function httpHelpers(client: any, connectorFields: string[] = []) {
   const helpers: Record<string, any> = {};
   ['GET', 'PATCH', 'POST', 'PUT', 'DELETE'].forEach(
```
bo-dun-1, line 822: could we also support lowercase get, patch, etc.?

bo-dun-1 (approved):

PR #1096, opened 2023-11-30
Multiple non-nested repeats per step support

ericpias:


src/Form/index.tsx
```diff
@@ -417,13 +425,15 @@ function Form({
     updateFieldValues(updatedValues);
   }
 
-  function removeRepeatedRow(index: number) {
+  function removeRepeatedRow(element: any) {
```
bo-dun-1, line 428: most of the logic in this function is the same as the logic in the addRepeatedRow function. Can we unify?

src/Form/index.tsx
```diff
@@ -395,11 +401,13 @@ function Form({
...
   }, [stepKey]);
 
-  function addRepeatedRow() {
-    // Collect a list of all repeated elements
-    const repeatedServarFields = activeStep.servar_fields.filter(
-      (field: any) => field.servar.repeated
-    );
+  function addRepeatedRow(repeatContainer: Subgrid | undefined) {
+    // Collect a list of all relevant repeated elements.
+    // Legacy: if no repeatContainer is provided, then all repeated elements are relevant
```
bo-dun-1, line 406: instead of leaving legacy logic in the react library, i think we should do a backend migration so all forms will now specify the repeating container. should be a backwards compatible backend change i believe

src/Form/index.tsx
ericpias: I don't really agree.  They are significantly different and commonizing them would be difficult with no real gain.

src/Form/index.tsx
ericpias: BE migration added.  This code was then simplified since repeat_container always set.

bo-dun-1 (approved):

PR #1101, opened 2023-12-08
Collaborator logic

ericpias:


src/utils/Collaborator.ts
```diff
@@ -0,0 +1,39 @@
+/**
+ * Represents a Collaborator on a form.  Part of the SDK coding model.
+ */
+
+export default class Collaborator {
```
bo-dun-1, line 5: maybe we should make a folder to put Collaborator.ts, Field.ts, etc. into

bo-dun-1 (approved):

PR #1109, opened 2023-12-20
feat: Fixed containers

ondrosh:
## Changes

- Add the ability to fix containers (allowing sticky headers and footers)

### DIFF
--- a/src/Form/components/DevNavBar.tsx
+++ b/src/Form/components/DevNavBar.tsx
@@ -129,7 +129,7 @@ export default function DevNavBar({
     <>
       <div
         css={{
-          zIndex: 2,
+          zIndex: 1000,
           position: 'fixed',
           left: '0',
           top: '0',
@@ -188,7 +188,7 @@ export default function DevNavBar({
       </div>
       <div
         css={{
-          zIndex: 2,
+          zIndex: 1000,
           position: 'fixed',
           bottom: '30px',
           left: '50%',
@@ -238,7 +238,7 @@ export default function DevNavBar({
   ) : (
     <div
       css={{
-        zIndex: 2,
+        zIndex: 1000,
         position: 'fixed',
         top: '0',
         backgroundColor: 'white',
--- a/src/Form/components/DevNavBar.tsx
+++ b/src/Form/components/DevNavBar.tsx
@@ -9,6 +9,7 @@ import {
   DiagonalArrowIcon
 } from '../../elements/components/icons';
 import { noTextSelectStyles } from '../../elements/styles';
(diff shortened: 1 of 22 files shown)


src/Form/components/DevNavBar.tsx
```diff
@@ -188,7 +188,7 @@ export default function DevNavBar({
       </div>
       <div
         css={{
-          zIndex: 2,
+          zIndex: 1000,
```
bo-dun-1, line 191: as discussed, would be great if we could clean up and centralize z indices

src/Form/grid/StyledContainer/index.tsx
```diff
@@ -52,6 +62,84 @@ export const StyledContainer = forwardRef<HTMLDivElement, StyledContainerProps>(
...
+            const { top, left } = container.getBoundingClientRect();
+            const isParentRoot = !_node.parent.parent;
+            const isFirstChild = _node.parent.children[0].id === _node.id;
+            const isLastChild =
+              _node.parent.children[_node.parent.children.length - 1].id ===
+              _node.id;
+
+            if (isParentRoot && isFirstChild) {
+              fixedContainerRef.current.style.top = '0';
+              fixedContainerRef.current.style.left = `${left}px`;
```
bo-dun-1, line 102: this line can probably be pulled out instead of written three times i think

src/Form/grid/StyledContainer/index.tsx
```diff
@@ -52,6 +62,84 @@ export const StyledContainer = forwardRef<HTMLDivElement, StyledContainerProps>(
...
+            }
+
+            if (isParentRoot && isLastChild) {
```
bo-dun-1, line 105: should probably be else ifs. also would be great to add a comment explaining the snapping logic

src/Form/grid/StyledContainer/index.tsx
```diff
@@ -52,6 +62,84 @@ export const StyledContainer = forwardRef<HTMLDivElement, StyledContainerProps>(
 
     useContainerEngine(node, rawNode, ref);
 
+    useEffect(() => {
+      nodeRef.current = node || rawNode;
+    }, [node, rawNode]);
+
+    useEffect(() => {
```
bo-dun-1, line 69: should this be pulled out into a hook? like useFixedContainer. maybe the isFixed logic in useContainerStyles could be added to it as well

bo-dun-1: the isFixedContainer function could be put in that file as well. i like the idea of unifying our fixed logic into one place

bo-dun-1 (approved):

PR #1133, opened 2024-02-06
multiple collaborator emails

ericpias:


src/elements/components/FormOff.tsx
```diff
@@ -4,12 +4,18 @@ import { openTab } from '../../utils/browser';
...
-    ? 'You have successfully filled out the form.'
-    : "This form isn't currently collecting responses.";
+  const FILLED_OUT = 'filled_out';
+  const CLOSED = 'closed';
+  const COLLAB_COMPLETED = 'collab_completed';
+  const messages: any = {
+    [FILLED_OUT]: 'You have successfully filled out the form.',
+    [CLOSED]: "This form isn't currently collecting responses.",
+    [COLLAB_COMPLETED]:
+      'This form has already been completed for this collaboration role.'
```
bo-dun-1, line 17: Your collaboration group has completed this form

src/elements/components/FormOff.tsx
```diff
@@ -4,12 +4,18 @@ import { openTab } from '../../utils/browser';
...
 export default function FormOff({
   width = 400,
-  noEdit = false,
+  message = 'closed',
   showCTA = true
 }) {
-  const message = noEdit
-    ? 'You have successfully filled out the form.'
-    : "This form isn't currently collecting responses.";
+  const FILLED_OUT = 'filled_out';
```
bo-dun-1, line 10: since you have these enums defined, i think you should pull them out into global constants and pass in constants instead of raw strings

src/Form/index.tsx
```diff
@@ -934,9 +935,12 @@ function Form({
         // @ts-expect-error TS(2345): Argument of type 'Promise<any[]>' is not assignabl... Remove this comment to see the full error message
         .fetchSession(formPromise, true)
         .then(([session, steps]) => {
-          if (!session) {
+          if (!session || session.collaborator?.invalid) {
             setFormSettings({ ...formSettings, formOff: true });
             return;
+          } else if (session.collaborator?.completed) {
+            setCollaboratorCompleted(true);
```
bo-dun-1, line 942: instead of creating a new collaboratorcompleted state variable, i wonder if we can make **formOff** an enum instead of a boolean instead

bo-dun-1 (approved):

PR #1146, opened 2024-02-23
Added SMS Code and Telesign Voice No Code Actions

summep:


src/utils/sensitiveActions.ts
```diff
@@ -0,0 +1,21 @@
+import internalState, { setFormInternalState } from './internalState';
+import Auth from '../auth/internal/AuthIntegrationInterface';
+
+export const getSensitiveActions = (formUuid: string) => {
```
bo-dun-1, line 4: getPrivateActions maybe?

bo-dun-1: i think we shoudl leave a omment here as well explaining what this is for

src/utils/sensitiveActions.ts
```diff
@@ -0,0 +1,21 @@
...
+  if (!internalState[formUuid]) {
```
bo-dun-1, line 5: this code doesn't seem necessary for this function

bo-dun-1 (approved):

PR #1150, opened 2024-02-27
No Code Email Actions

summep:


src/utils/client.ts
```diff
@@ -1036,6 +1036,21 @@ export default class Client {
       } else throw Error(parseError(await response.json()));
     }
   }
+
+  async sendEmail(templateId: string) {
```
bo-dun-1, line 1040: could we pull all of the integration actions into a separate file? maybe it can be a base class inherited by the client.ts class here. we've gotten to the point IMO where we have too many functions in one place here.

src/utils/integClient.ts
```diff
@@ -0,0 +1,439 @@
...
+          (this.ignoreNetworkErrors?.current ||
+            TYPE_MESSAGES_TO_IGNORE.includes(e.message)) &&
+          e instanceof TypeError
+        )
+          return;
+        throw e;
+      });
+  }
+
+  // THIRD-PARTY INTEGRATIONS
```
bo-dun-1, line 90: This comment can maybe be moved to the top of the class definition to describe the purpose of this client

src/utils/client.ts
```diff
@@ -70,89 +70,7 @@ export const updateRegionApiUrls = (region: string) => {
...
-          (this.ignoreNetworkErrors?.current ||
-            TYPE_MESSAGES_TO_IGNORE.includes(e.message)) &&
-          e instanceof TypeError
-        )
-          return;
-        throw e;
-      });
-  }
-
+export default class Client extends IntegClient {
```
bo-dun-1, line 73: this would probably be better as a folder called client. The main client can live at index.ts, and the integration base can live at IntegrationClient. Let's use the full word for important definitions. abbreviations are better for local variables.

also now that we have two client definitions, maybe the overall class should be called `FeatheryClient` for clarity

bo-dun-1: IDEs make it easy to rename objects being used in many places: https://code.visualstudio.com/Docs/editor/editingevolved#_rename-symbol

PR #1161, opened 2024-03-12
Request Queue

summep:


src/utils/offlineRequestHandler.ts
```diff
@@ -0,0 +1,268 @@
...
+      case 'arrayBuffer':
+        return body;
+      case 'text':
+      default:
+        return body;
+    }
+  }
+}
+
+function createOfflineRequestHandler(): OfflineRequestHandler {
```
bo-dun-1, line 256: we may not need this function, could do something like:

const offlineRequestHandler = new OfflineRequestHandler(...);

src/utils/featheryClient/index.ts
```diff
@@ -86,14 +89,24 @@ export default class FeatheryClient extends IntegrationClient {
...
       },
       method: 'POST',
       body: JSON.stringify(data)
     };
-    return this._fetch(url, options);
+    if (navigator.onLine) {
+      return this._fetch(url, options);
+    } else {
+      const { sdkKey } = initInfo();
+      options.headers = {
```
bo-dun-1, line 103: options.headers.Authorization = ...

if your'e running into a type error, you could do something like 

headers: {'Content-Type': 'application/json'} as Record<string, string> in the initial option definition

src/utils/offlineRequestHandler.ts
```diff
@@ -0,0 +1,268 @@
...
+    try {
+      const getTransaction = async () => {
+        const tx = db.transaction(this.storeName, 'readwrite');
+        const store = tx.objectStore(this.storeName);
+        return { tx, store };
+      };
+      const requestClone: Request = requestToSave.clone();
+      let serializedBody: SerializedRequestBody = { type: 'text', body: '' };
+
+      if (requestClone.method !== 'GET' && requestClone.method !== 'HEAD') {
```
bo-dun-1, line 104: if (!['GET', 'HEAD'].includes(requestClone.method))

PR #1168, opened 2024-03-19
feat: Add a polling mechanism to AI Extractions

ondrosh:
## Changes

- Add a polling mechanism to AI Extraction to avoid timeouts and lower server load

### DIFF
--- a/src/utils/featheryClient/index.ts
+++ b/src/utils/featheryClient/index.ts
@@ -552,13 +552,51 @@ export default class FeatheryClient extends IntegrationClient {
     const data = {
       fuser_key: userId,
       extraction_id: extractionId,
-      run_async: runAsync
+      run_async: true
     };
-    return this._fetch(`${AI_URL}ai/vision/`, {
+
+    this._fetch(`${AI_URL}ai/vision/`, {
       headers: { 'Content-Type': 'application/json' },
       method: 'POST',
       body: JSON.stringify(data)
-    }).then((res) => (res ? res.json() : Promise.resolve({})));
+    });
+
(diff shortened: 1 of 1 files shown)


src/utils/featheryClient/index.ts
```diff
@@ -552,13 +552,51 @@ export default class FeatheryClient extends IntegrationClient {
     const data = {
       fuser_key: userId,
       extraction_id: extractionId,
-      run_async: runAsync
+      run_async: true
```
bo-dun-1, line 555: seems like we don't need this flag anymore?

src/utils/featheryClient/index.ts
```diff
@@ -552,13 +552,51 @@ export default class FeatheryClient extends IntegrationClient {
...
       body: JSON.stringify(data)
-    }).then((res) => (res ? res.json() : Promise.resolve({})));
+    });
+
+    return new Promise((resolve) => {
+      const CHECK_INTERVAL = 2000;
+      const MAX_TIME = 3 * 60 * 1000;
+      const MAX_ATTEMPTS = MAX_TIME / CHECK_INTERVAL;
+      let attempts = 0;
+
```
bo-dun-1, line 569: maybe here we can have 

if (runAsync) return resolve({});

const checkCompletion = ...

src/utils/featheryClient/index.ts
ondrosh: I was leaving it for backwards compatibility but we might not need to do that if no one is embedded using this feature right now

bo-dun-1 (approved):

PR #1182, opened 2024-03-30
Logic Rule Request Offline Handling

summep:


src/utils/featheryClient/index.ts
```diff
@@ -574,11 +574,27 @@ export default class FeatheryClient extends IntegrationClient {
       data.field_values = fieldValues;
     }
 
-    return this._fetch(`${API_URL}custom_request/`, {
+    const url = `${API_URL}custom_request/`;
```
bo-dun-1, line 577: const run = () => this._fetch(url, options);
if (typeof payload === 'string' && payload.method === 'GET') {
  return run().then(response...)
}

return this.offlineRequestHandler...

src/utils/featheryClient/index.ts
```diff
@@ -574,11 +574,27 @@ export default class FeatheryClient extends IntegrationClient {
...
+    const options = {
       headers: { 'Content-Type': 'application/json' },
       method: 'POST',
       body: JSON.stringify(data)
-    }).then((response) => (response ? response.json() : Promise.resolve()));
+    };
+    if (
+      typeof payload !== 'string' &&
```
bo-dun-1, line 584: why do we check the payload type here?

src/utils/offlineRequestHandler.ts
```diff
@@ -295,23 +295,32 @@ export class OfflineRequestHandler {
...
-            resolve(requests);
+            noFormKeyRequests.sort((a, b) => a.timestamp - b.timestamp);
+            resolve([...noFormKeyRequests, ...requests]);
           }
         };
       });
 
-      const submitRequests = allRequests.filter((req) => req.type === 'submit');
+      const noFormKeyRequests = allRequests.filter((req) => !req.formKey);
+      await this.replayRequestsInParallel(noFormKeyRequests);
```
bo-dun-1, line 319: can these be replayed in parallel with the submit requests after this as well?

src/utils/featheryClient/index.ts
summep: payload could be a string and payload.method will give TS error

src/utils/offlineRequestHandler.ts
```diff
@@ -295,24 +295,34 @@ export class OfflineRequestHandler {
         SerializedRequest[]
       >((resolve) => {
         const requests: SerializedRequest[] = [];
+        const noFormKeyRequests: SerializedRequest[] = [];
```
bo-dun-1, line 298: why not just store these in requests since you have to filter out the noFormKeyRequests on line 318 anyways?

src/utils/offlineRequestHandler.ts
```diff
@@ -311,7 +311,9 @@ export class OfflineRequestHandler {
         };
       });
 
-      const submitRequests = allRequests.filter((req) => req.type === 'submit');
+      const submitRequests = allRequests.filter(
+        (req) => req.type === 'submit' || req.type === 'customRequest'
```
bo-dun-1, line 315: ['submit', 'customRequest'].includes(req.type)

bo-dun-1 (approved):

PR #1199, opened 2024-04-24
Kailash/phone number node bugfix

avnkailash:
Tested scenarios:
1. Loaded a really long form and intereacted with the phone number dropdown.
2. Switched between US and afganistan multiple times.
3. Switched to a completely different country and back to US
4. Entered a phone number for US. Then, I changed the country to Afghanistan and entered a different phone number.

Outcome:
1. The auto-scrolling to the top of page is not observed.
2. There was no Node not found error which was observed earlier.

### DIFF
--- a/src/elements/fields/PhoneField/index.tsx
+++ b/src/elements/fields/PhoneField/index.tsx
@@ -231,6 +231,7 @@ function PhoneField({
           show={show}
           onHide={() => setShow(false)}
           placement='bottom-start'
+          container={triggerRef.current}
         >
           {(props) => {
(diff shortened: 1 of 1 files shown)


avnkailash:
The `container` prop of the overlay lets us specify which element will hold the overlay. If we do not specify the container, there are chances that the page's body is taken as the container. This can cause the auto-scrolling issue that we observed.

The `key` prop will force React to remount the component if the container changes. This is very helpful when react is dealing with left over components. It will be able to perform cleanup and avoid the node not found type of errors.

bo-dun-1 (approved):

PR #1203, opened 2024-04-27
3) Static Embed Id: Support Embeds via Form ID

soundsbywise:
# Overview
As a part of the static embed ID project, this PR adds support for embedding your form via its live slug, passed as `formId`.

[DESIGN DOC](https://docs.google.com/document/d/12zPSUgPpQscbbHnQ0n18qi2cN9RGnK2YfzlCrl2jleQ/edit)
[ASANA](https://app.asana.com/0/home/1207048571643766/1206468676347913)

### DIFF
--- a/src/Form/index.tsx
+++ b/src/Form/index.tsx
@@ -165,7 +165,11 @@ export * from './grid/StyledContainer';
 export type { StyledContainerProps } from './grid/StyledContainer';
 
 export interface Props {
-  formId: string;
+  formId?: string;
+  /**
+   * @deprecated use formId instead
+   */
+  formName: string; // TODO: remove support for formName
   onChange?: null | ((context: ContextOnChange) => Promise<any> | void);
   onLoad?: null | ((context: FormContext) => Promise<any> | void);
   onFormComplete?: null | ((context: FormContext) => Promise<any> | void);
@@ -219,6 +223,7 @@ function Form({
   _internalId,
   _isAuthLoading = false,
   _bypassCDN = false,
+  formName: formNameProp,
   formId, // The 'live' env slug
   onChange = null,
   onLoad = null,
@@ -243,11 +248,17 @@ function Form({
   children,
   _draft = false
 }: InternalProps & Props) {
-  const [formName, setFormName] = useState('');
+  // TODO: remove support for formName
+  if (formNameProp)
+    console.warn(
+      'The `formName` parameter is deprecated and support will be removed in a future library version. Please use `formId` instead.'
+    );
+  const [formName, setFormName] = useState(formNameProp || '');
+  const formKey = formId || formName; // prioritize formID but fall back to name
   const clientRef = useRef<any>();
   const client = clientRef.current;
   const history = useHistory();
-  const session = initState.formSessions[formId];
+  const session = initState.formSessions[formKey];
 
   const [autoValidate, setAutoValidate] = useState(false);
 
@@ -912,7 +923,7 @@ function Form({
     if (clientRef.current) return;
 
     clientRef.current = new FeatheryClient(
-      formId,
+      formKey,
       hasRedirected,
       _draft,
       _bypassCDN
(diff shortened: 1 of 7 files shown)


src/Form/index.tsx
```diff
@@ -219,7 +219,7 @@ function Form({
   _internalId,
   _isAuthLoading = false,
   _bypassCDN = false,
-  formName,
+  formId, // The 'live' env slug
```
bo-dun-1, line 222: the sdk still needs to support taking in a formName prop (that we should mark as deprecated w/ a TODO) because some users will upgrade their version and Feathery should still work for them, but we can emit a warning message to console to notify them it's deprecated and will be removed soon

src/Form/index.tsx
```diff
@@ -922,6 +924,7 @@ function Form({
       .fetchForm(initialValues, language)
       .then(async (data: any) => {
         await updateCustomHead(data.custom_head ?? '');
+        setFormName(data.form_name); // API definition guarantees this.
```
bo-dun-1, line 927: let's move this to the next .then function, this one is special for the custom_head sinc eit needs to occur before the other stuff

src/integrations/utils.ts
```diff
@@ -99,7 +99,7 @@ export async function initializeIntegrations(
   const gtm = integs['google-tag-manager'];
   if (gtm) initializeTagManager(gtm);
   if (integs.firebase || integs.stytch) {
-    authState.authFormKey = featheryClient.formKey;
+    authState.authFormKey = featheryClient.formKey; // TODO: refactor to form slug
```
bo-dun-1, line 102: do these TODOs need to be addressed?

bo-dun-1 (changes requested):

src/Form/index.tsx
soundsbywise: done

src/integrations/utils.ts
soundsbywise: acc left the comment, its been addressed. will remove

src/Form/index.tsx
```diff
@@ -165,7 +165,11 @@ export * from './grid/StyledContainer';
 export type { StyledContainerProps } from './grid/StyledContainer';
 
 export interface Props {
-  formName: string;
+  formId?: string;
+  /**
+   * @deprecated use formId instead
+   */
+  formName: string; // TODO: remove support for formName
```
soundsbywise, line 172: leaving `formName` required for now. I think it makes more sense to remove `formName` and make `formId` required when we're ready to release a version that makes `formId` the default. Also, makes sense to do that in a diff version since its a breaking change

src/Form/index.tsx
```diff
@@ -243,10 +248,17 @@ function Form({
   children,
   _draft = false
 }: InternalProps & Props) {
+  // TODO: remove support for formName
+  if (formNameProp)
```
soundsbywise, line 252: i think it's fine to leave te warning here even if there are re-renders but we can move to a side effect if need be

bo-dun-1: i think there are a lot of re-renders so IMO a side effect would be a better experience

src/Form/index.tsx
bo-dun-1: IMO it'd be better to make the formId the default now, it won't break anything but will be annoying and push users to switch

src/utils/internalState.ts
```diff
@@ -17,7 +17,8 @@ interface FormInternalState {
   products: Record<string, SimplifiedProduct>;
   cart: Cart;
   collaborator: Collaborator;
-  formName: string;
+  formName: string; // TODO: remove support for formName
```
bo-dun-1, line 20: where is formName still needed for internal state?

src/Form/index.tsx
soundsbywise: cool, should we make `formName` optional then? so new users that are on the latest version don't write anything based on `formName` and existing users can remove the prop when they eventually refactor to `formId`

src/utils/internalState.ts
soundsbywise: i don't think it is needed since we set `formName` for the dev nav bar in local state - removed it.

src/Form/index.tsx
```diff
@@ -248,12 +248,7 @@ function Form({
   children,
   _draft = false
 }: InternalProps & Props) {
-  // TODO: remove support for formName
-  if (formNameProp)
-    console.warn(
-      'The `formName` parameter is deprecated and support will be removed in a future library version. Please use `formId` instead.'
-    );
-  const [formName, setFormName] = useState(formNameProp || '');
+  const [formName, setFormName] = useState(formNameProp || ''); // TODO: remove support for formName
```
bo-dun-1, line 251: formName (deprecated)

bo-dun-1 (approved):

PR #1205, opened 2024-05-01
Debounce submitCustom

soundsbywise:
# Overview

The `submitCustom` method can be called several times in rapid succession which leads to BE race conditions. To offset this we debounce the method so that each successive call has to wait 1 second.

# Notes
1. Updates are aggregated so that the latest key value pairs are sent to the BE when a call goes through.
2. If the method is called and the form is about to be completed, we send the pending changes to the backend regardless of the wait time

# Testing
Submitted the `Book A Demo` feathery form before and after changes - works the same.

### DIFF
--- a/src/utils/featheryClient/index.ts
+++ b/src/utils/featheryClient/index.ts
@@ -84,19 +84,12 @@ export default class FeatheryClient extends IntegrationClient {
    * Used to aggregate field value updates for sucessive calls to
    * submitCustom within the debounce window
    */
-  pendingSubmitCustomUpdates: { [key: string]: any };
+  pendingCustomFieldUpdates: { [key: string]: any };
 
   /**
    * Debounced implementation of submitCustom
    */
   debouncedSubmitCustom: DebouncedFunc<(override: boolean) => Promise<void>>;
-
-  /**
-   * If there is a pending invocation of submitCustom, this method calls it immediately
-   */
-  flushPendingSubmitCustomUpdates: (
-    override?: boolean
-  ) => Promise<void> | undefined;
 
   constructor(
     formKey = '',
@@ -105,17 +98,11 @@ export default class FeatheryClient extends IntegrationClient {
     bypassCDN = false
   ) {
     super(formKey, ignoreNetworkErrors, draft, bypassCDN);
-    this.pendingSubmitCustomUpdates = {};
+    this.pendingCustomFieldUpdates = {};
     this.debouncedSubmitCustom = debounce(
       this._debouncedSubmitCustom.bind(this),
       SUBMIT_CUSTOM_DEBOUNCE_WINDOW
     );
-    this.flushPendingSubmitCustomUpdates = (override?: boolean) => {
-      // we call the debounced method and then the flush to immediately submit changes
-      // see: https://github.com/lodash/lodash/issues/4185#issuecomment-462388355
-      this.debouncedSubmitCustom(override ?? true);
-      return this.debouncedSubmitCustom.flush();
-    };
   }
 
   async _submitJSONData(servars: any, stepKey: string, noComplete: boolean) {
@@ -466,12 +453,10 @@ export default class FeatheryClient extends IntegrationClient {
    * Debounceable function responsible for pinging `/api/panel/custom/submit/<version>`
    */
   async _debouncedSubmitCustom(override: boolean) {
-    if (Object.keys(this.pendingSubmitCustomUpdates).length === 0) return;
+    if (Object.keys(this.pendingCustomFieldUpdates).length === 0) return;
 
-    console.log(' I AM FLUSHING?');
-
-    const customKeyValues = { ...this.pendingSubmitCustomUpdates };
-    this.pendingSubmitCustomUpdates = {}; // Clear pending updates after copying them
+    const customKeyValues = { ...this.pendingCustomFieldUpdates };
+    this.pendingCustomFieldUpdates = {}; // Clear pending updates after copying them
 
     const { userId } = initInfo();
     const url = `${API_URL}panel/custom/submit/v3/`;
@@ -531,7 +516,17 @@ export default class FeatheryClient extends IntegrationClient {
   }
 
   /**
-   * `beforeunload` event handler that flushes the pending submit custome changes
+   * If there is a pending invocation of submitCustom, this method calls it immediately
+   */
+  flushPendingSubmitCustomUpdates(override: boolean = true) {
+    // we call the debounced method and then flush() to immediately submit changes
+    // see: https://github.com/lodash/lodash/issues/4185#issuecomment-462388355
+    this.debouncedSubmitCustom(override);
+    return this.debouncedSubmitCustom.flush();
+  }
+
+  /**
+   * `beforeunload` event handler that flushes the pending submit custom changes
    * when a user is attempting to exit the page.
    * @param event `BeforeUnloadEvent`
    * @returns
@@ -544,22 +539,19 @@ export default class FeatheryClient extends IntegrationClient {
 
   async submitCustom(
     customKeyValues: { [key: string]: any },
-    options?: {
-      override?: boolean;
-      shouldFlush?: boolean;
-    }
+    // Options
+    {
+      override = true,
+      shouldFlush = false
+    }: { override?: boolean; shouldFlush?: boolean } = {}
   ) {
-    const override = options?.override ?? true;
-    const shouldFlush = options?.shouldFlush ?? false;
-
     if (this.draft || this.noSave) return;
     if (Object.keys(customKeyValues).length === 0 && !shouldFlush) return;
     // If there are values passed, aggregate them in the pending queue
     Object.entries(customKeyValues).forEach(
-      ([key, value]) => (this.pendingSubmitCustomUpdates[key] = value)
+      ([key, value]) => (this.pendingCustomFieldUpdates[key] = value)
     );
-    console.log('\nPENDING UPDATES: ', this.pendingSubmitCustomUpdates);
-    if (Object.keys(this.pendingSubmitCustomUpdates).length) {
+    if (Object.keys(this.pendingCustomFieldUpdates).length) {
       // if there are pending changes, prevent user from exiting page and losing them
       featheryWindow().addEventListener(
         'beforeunload',
@@ -570,7 +562,6 @@ export default class FeatheryClient extends IntegrationClient {
     }
     // if we don't want to override the existing values or the caller tells us to flush, immediately flush
     if (!override || shouldFlush) {
-      console.log('\nSHOULD FLUSH!!!');
       return this.flushPendingSubmitCustomUpdates(override);
     }
     // otherwise, ping the API in normal debounced cadence
@@ -629,12 +620,11 @@ export default class FeatheryClient extends IntegrationClient {
       body: JSON.stringify(data)
     };
 
-    const stepKey =
-      eventData.event === 'load'
-        ? eventData.previous_step_key
-        : eventData.step_key;
-    // If the user completed a step or skipped it, submit pending changes to BE
-    if (['complete', 'skip'].includes(eventData.event)) {
+    let stepKey;
+    if (eventData.event === 'load') {
+      stepKey = eventData.previous_step_key;
+    } else {
+      stepKey = eventData.step_key;
       this.flushPendingSubmitCustomUpdates();
     }
     return this.offlineRequestHandler.runOrSaveRequest(
--- a/src/utils/featheryClient/index.ts
+++ b/src/utils/featheryClient/index.ts
@@ -541,8 +541,8 @@ export default class FeatheryClient extends IntegrationClient {
       shouldFlush?: boolean;
     }
   ) {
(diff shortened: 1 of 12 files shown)


src/utils/featheryClient/index.ts
```diff
@@ -474,6 +528,23 @@ export default class FeatheryClient extends IntegrationClient {
     );
   }
 
+  async submitCustom(customKeyValues: { [key: string]: any }, override = true) {
+    if (this.draft || this.noSave) return;
+    const shouldFlushPendingUpdates = this.formContext?.isLastStep();
```
bo-dun-1, line 533: IMO instead of doing the check here, you can have a separate boolean parameter that triggers the flush, and you can trigger the flush each time submitStep or registerEvent (when the event type is a step completion) is called. That will be more robust and guarantee we get the data the user intends to set before critical event times (e.g. integration runs) while also solving the problem of a ton of submitCusotm calls stacked on each other

src/utils/featheryClient/index.ts
```diff
@@ -474,6 +528,23 @@ export default class FeatheryClient extends IntegrationClient {
     );
   }
 
+  async submitCustom(customKeyValues: { [key: string]: any }, override = true) {
```
bo-dun-1, line 531: if override is false, we should flush it immediately. override=false and override=true calls have different behavior (they either override exisitng field data on the backend or not) and can't be combined. Alos override=false is only used in one place that doesn't need to be debounced so it's fine to trigger immediately

src/utils/featheryClient/index.ts
```diff
@@ -474,6 +522,51 @@ export default class FeatheryClient extends IntegrationClient {
...
+    let override = options?.override ?? true;
+    let shouldFlush = options?.shouldFlush ?? false;
+
+    if (this.draft || this.noSave) return;
+    if (Object.keys(customKeyValues).length === 0 && !shouldFlush) return;
+    // If there are values passed, aggregate them in the pending queue
+    Object.entries(customKeyValues).forEach(
+      ([key, value]) => (this.pendingSubmitCustomUpdates[key] = value)
+    );
+    if (Object.keys(this.pendingSubmitCustomUpdates).length) {
```
soundsbywise, line 553: added: if pending changes, listen for before unload event

src/utils/featheryClient/index.ts
```diff
@@ -474,6 +522,51 @@ export default class FeatheryClient extends IntegrationClient {
     );
   }
 
+  /**
+   * `beforeunload` event handler that flushes the pending submit custome changes
+   * when a user is attempting to exit the page.
+   * @param event `BeforeUnloadEvent`
+   * @returns
+   */
+  _flushPendingChangesBeforeUnload(event: BeforeUnloadEvent) {
```
soundsbywise, line 531: added: event listener - when user tries to exits page, flush pending changes

src/utils/featheryClient/index.ts
```diff
@@ -465,7 +509,11 @@ export default class FeatheryClient extends IntegrationClient {
       method: 'POST',
       body: formData
     };
-
+    // Here we can safely remove the listener becasue offlineRequestHandler has its own beforeunload
+    featheryWindow().removeEventListener(
```
soundsbywise, line 513: added: once pending changes cleared, stop flushing in case user tries to exit page

src/utils/featheryClient/index.ts
```diff
@@ -530,6 +623,10 @@ export default class FeatheryClient extends IntegrationClient {
       eventData.event === 'load'
         ? eventData.previous_step_key
         : eventData.step_key;
+    // If the user completed a step or skipped it, submit pending changes to BE
+    if (['complete', 'skip'].includes(eventData.event)) {
```
soundsbywise, line 627: added: events that trigger a flush

src/utils/featheryClient/index.ts
```diff
@@ -497,7 +590,7 @@ export default class FeatheryClient extends IntegrationClient {
     const fileServars = servars.filter(isFileServar);
     this.submitQueue = Promise.all([
       this.submitQueue,
-      this.submitCustom(hiddenFields),
+      this.submitCustom(hiddenFields, { shouldFlush: true }),
```
soundsbywise, line 593: addedL submitCustom should flush pending changes

src/utils/featheryClient/index.ts
```diff
@@ -72,7 +74,50 @@ export const updateRegionApiUrls = (region: string) => {
...
+    super(formKey, ignoreNetworkErrors, draft, bypassCDN);
+    this.pendingSubmitCustomUpdates = {};
+    this.debouncedSubmitCustom = debounce(
+      this._debouncedSubmitCustom.bind(this),
+      SUBMIT_CUSTOM_DEBOUNCE_WINDOW
+    );
+    this.flushPendingSubmitCustomUpdates = (override: boolean = true) => {
+      // we call the debounced method and then the flush to immediately submit changes
+      // see: https://github.com/lodash/lodash/issues/4185#issuecomment-462388355
+      this.debouncedSubmitCustom(override);
```
soundsbywise, line 116: added: to immediately submit changes we must call the flush method in this way

src/utils/featheryClient/index.ts
```diff
@@ -72,7 +74,50 @@ export const updateRegionApiUrls = (region: string) => {
...
+ * The number of milliseconds waited until another submitCustom call
+ */
+const SUBMIT_CUSTOM_DEBOUNCE_WINDOW = 1000;
+
 export default class FeatheryClient extends IntegrationClient {
+  /**
+   * Used to aggregate field value updates for sucessive calls to
+   * submitCustom within the debounce window
+   */
+  pendingSubmitCustomUpdates: { [key: string]: any };
```
bo-dun-1, line 87: pendingCustomFieldUpdates

src/utils/featheryClient/index.ts
```diff
@@ -72,7 +74,50 @@ export const updateRegionApiUrls = (region: string) => {
...
+    draft = false,
+    bypassCDN = false
+  ) {
+    super(formKey, ignoreNetworkErrors, draft, bypassCDN);
+    this.pendingSubmitCustomUpdates = {};
+    this.debouncedSubmitCustom = debounce(
+      this._debouncedSubmitCustom.bind(this),
+      SUBMIT_CUSTOM_DEBOUNCE_WINDOW
+    );
+    this.flushPendingSubmitCustomUpdates = (override?: boolean) => {
```
bo-dun-1, line 113: can this be statically defined as a class function instead of in the constructor dynamically

src/utils/featheryClient/index.ts
```diff
@@ -417,9 +462,16 @@ export default class FeatheryClient extends IntegrationClient {
...
-  async submitCustom(customKeyValues: any, override = true) {
-    if (this.draft || this.noSave) return;
-    if (Object.keys(customKeyValues).length === 0) return;
+  /**
+   * Debounceable function responsible for pinging `/api/panel/custom/submit/<version>`
+   */
+  async _debouncedSubmitCustom(override: boolean) {
+    if (Object.keys(this.pendingSubmitCustomUpdates).length === 0) return;
+
+    console.log(' I AM FLUSHING?');
```
bo-dun-1, line 471: please remove debug logs

src/utils/featheryClient/index.ts
bo-dun-1: also parameter can instead be override = true so no need for ?? true when passing it into debouncedSubmitCustom below

src/utils/featheryClient/index.ts
```diff
@@ -72,7 +74,50 @@ export const updateRegionApiUrls = (region: string) => {
...
+    this.pendingSubmitCustomUpdates = {};
+    this.debouncedSubmitCustom = debounce(
+      this._debouncedSubmitCustom.bind(this),
+      SUBMIT_CUSTOM_DEBOUNCE_WINDOW
+    );
+    this.flushPendingSubmitCustomUpdates = (override?: boolean) => {
+      // we call the debounced method and then the flush to immediately submit changes
+      // see: https://github.com/lodash/lodash/issues/4185#issuecomment-462388355
+      this.debouncedSubmitCustom(override ?? true);
+      return this.debouncedSubmitCustom.flush();
```
bo-dun-1, line 117: i'm not familiar with the flush function. will this return the promise from the network request so it can be awaited by the client of this function?

src/utils/featheryClient/index.ts
```diff
@@ -474,6 +530,53 @@ export default class FeatheryClient extends IntegrationClient {
     );
   }
 
+  /**
+   * `beforeunload` event handler that flushes the pending submit custome changes
```
bo-dun-1, line 534: nit: submit custom

src/utils/featheryClient/index.ts
```diff
@@ -474,6 +530,53 @@ export default class FeatheryClient extends IntegrationClient {
...
+  _flushPendingChangesBeforeUnload(event: BeforeUnloadEvent) {
+    event.preventDefault();
+    this.flushPendingSubmitCustomUpdates();
+    return (event.returnValue = '');
+  }
+
+  async submitCustom(
+    customKeyValues: { [key: string]: any },
+    options?: {
+      override?: boolean;
```
bo-dun-1, line 548: customKeyValues, {override=true, shouldFlush=false}

destructure directly with defaults will make this cleaner

src/utils/featheryClient/index.ts
bo-dun-1, line 113: i'm not sure this method is necessary actually. if there is data to flush, there will be a debounced, pending submitCustom call as well so no need to call this again or have this helper function. .flush() can be called directly.

src/utils/featheryClient/index.ts
```diff
@@ -530,6 +633,10 @@ export default class FeatheryClient extends IntegrationClient {
       eventData.event === 'load'
         ? eventData.previous_step_key
         : eventData.step_key;
```
bo-dun-1, line 635: let stepKey;
if (eventData.event === 'load') {
    stepKey = eventData.previous_step_key;
} else {
    stepKey = eventData.step_key;
    this.debouncedSubmitCustom.flush();
}

src/utils/featheryClient/index.ts
soundsbywise: whoops, my bad

src/utils/featheryClient/index.ts
soundsbywise: I think wrapping the flush call in this method creates for a cleaner API long term. 

There's currently 3 instances where we flush. Since the `lodash.flush` implementation has this [caveat](https://github.com/lodash/lodash/issues/4185#issuecomment-462388355) that requires us to 
1. call the debounced function
2. then  call `flush()`

It's easier to debug and refactor the flush call if its defined in one place.

Otherwise, there's going to be 3 instances of:

```{typescript}
this.debouncedSubmitCustom(override ?? true);
this.debouncedSubmitCustom.flush();
```

src/utils/featheryClient/index.ts
```diff
@@ -474,6 +515,59 @@ export default class FeatheryClient extends IntegrationClient {
...
+      shouldFlush = false
+    }: { override?: boolean; shouldFlush?: boolean } = {}
+  ) {
+    if (this.draft || this.noSave) return;
+    if (Object.keys(customKeyValues).length === 0 && !shouldFlush) return;
+    // If there are values passed, aggregate them in the pending queue
+    Object.entries(customKeyValues).forEach(
+      ([key, value]) => (this.pendingCustomFieldUpdates[key] = value)
+    );
+    if (Object.keys(this.pendingCustomFieldUpdates).length) {
```
bo-dun-1, line 554: this should happen after hte if statement on line 564 IMO. if there's no debounce no point in adding the unload handler

src/utils/featheryClient/index.ts
```diff
@@ -417,9 +449,14 @@ export default class FeatheryClient extends IntegrationClient {
...
   }
 
-  async submitCustom(customKeyValues: any, override = true) {
-    if (this.draft || this.noSave) return;
-    if (Object.keys(customKeyValues).length === 0) return;
+  /**
+   * Debounceable function responsible for pinging `/api/panel/custom/submit/<version>`
+   */
+  async _debouncedSubmitCustom(override: boolean) {
+    if (Object.keys(this.pendingCustomFieldUpdates).length === 0) return;
```
bo-dun-1, line 456: shouldn't event listener attempt to be removed here as well before exit

soundsbywise: ah yes

src/utils/featheryClient/index.ts
```diff
@@ -474,6 +522,59 @@ export default class FeatheryClient extends IntegrationClient {
     );
   }
 
+  /**
+   * If there is a pending invocation of submitCustom, this method calls it immediately
+   */
+  flushPendingSubmitCustomUpdates(override = true) {
```
bo-dun-1, line 528: can we prepend this with an underscore since this is only used by the client internally

bo-dun-1 (approved):

PR #1208, opened 2024-05-06
unify check for pending changes in one place

soundsbywise:


src/utils/featheryClient/index.ts
```diff
@@ -454,10 +454,10 @@ export default class FeatheryClient extends IntegrationClient {
    */
   async _debouncedSubmitCustom(override: boolean) {
     if (Object.keys(this.pendingCustomFieldUpdates).length === 0) {
-      // if no pending changes, no need to keep listening for unload events.
-      this._removeCustomFieldListener();
       return;
     }
+    // if there are changes, listen for beforeunload events
+    this._addCustomFieldListener();
```
soundsbywise, line 460: if debounced function continues after the pending changes check, we can add the before unload listener

src/utils/featheryClient/index.ts
```diff
@@ -507,9 +507,11 @@ export default class FeatheryClient extends IntegrationClient {
       body: formData
     };
     // Here we can safely remove the listener because offlineRequestHandler has its own beforeunload
-    this._removeCustomFieldListener();
     return this.offlineRequestHandler.runOrSaveRequest(
-      () => this._fetch(url, options, true, true),
+      () => {
+        this._fetch(url, options, true, true);
+        this._removeCustomFieldListener();
```
soundsbywise, line 513: after the fetch, we can remove the before unload listener because it's certain that changes have been submitted

src/utils/featheryClient/index.ts
```diff
@@ -454,10 +454,10 @@ export default class FeatheryClient extends IntegrationClient {
    */
   async _debouncedSubmitCustom(override: boolean) {
     if (Object.keys(this.pendingCustomFieldUpdates).length === 0) {
```
soundsbywise, line 456: when the debounced function is called and there are no pending changes, no need to make a request

soundsbywise:
closing in favor of an updated pr with actual fix (tbd)

PR #1210, opened 2024-05-07
fix: datepicker input triggering errors

trebitowski:
Fixes DateSelector change events from triggering errors while typing a date. Now errors are triggered on calendar select and field blur. This is similar to how phone input works, where errors are suppressed until a valid input is entered.

- Input text is selected on focus
- Calendar opens on focus and prevent errors from autofocusing calendar on mobile
- Datepicker onChange does not trigger errors
- Datepicker onComplete updates value and triggers errors

### DIFF
--- a/src/Form/grid/Element/index.tsx
+++ b/src/Form/grid/Element/index.tsx
@@ -273,8 +273,8 @@ const Element = ({ node: el, form }: any) => {
             {...fieldProps}
             value={fieldVal}
             onChange={(val: any) => {
-              const change = changeValue(val, el, index);
-              if (change) onChange();
+              const change = changeValue(val, el, index, false, false);
+              if (change) debouncedOnChange();
             }}
             setRef={(ref: any) => {
               if (focusRef.current === el.id) focusRef.current = ref;
--- a/src/Form/grid/Element/index.tsx
+++ b/src/Form/grid/Element/index.tsx
@@ -273,8 +273,11 @@ const Element = ({ node: el, form }: any) => {
             {...fieldProps}
             value={fieldVal}
             onChange={(val: any) => {
-              const change = changeValue(val, el, index, false, false);
-              if (change) debouncedOnChange();
+              changeValue(val, el, index, false, false);
+            }}
+            onComplete={(val: any) => {
+              changeValue(val, el, index);
+              onChange();
             }}
             setRef={(ref: any) => {
               if (focusRef.current === el.id) focusRef.current = ref;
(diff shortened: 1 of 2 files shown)


src/elements/fields/DateSelectorField/index.tsx
```diff
@@ -201,20 +200,24 @@ function DateSelectorField({
...
             if (isTouchDevice()) {
               // hide keyboard on mobile focus
               e.target.readOnly = true;
             }
+            // select all text on focus
+            e.target.select();
             setFocused(true);
           }}
-          onBlur={() => setFocused(false)}
+          onBlur={() => {
```
bo-dun-1, line 217: i believe onDateChange should also be triggered if enter is typed as well now

bo-dun-1: also, what if i type in a date and immediately hit the submit button on the step? are we guaranteed this blur event will fire first and propagate state change before the submit logic validates?

src/elements/fields/DateSelectorField/index.tsx
```diff
@@ -138,12 +140,20 @@ function DateSelectorField({
...
 
+  // Called when the input is changed
   const onDateChange = (newDate: any) => {
     newDate = newDate ?? '';
     setInternalDate(newDate);
     onChange(formatDateString(newDate, servarMeta));
   };
 
+  // Called when the date is selected, or field is blurred
+  const onDateComplete = (newDate: any) => {
```
bo-dun-1, line 151: can we unify this function with onDateChange? the only difference is which callback funciton is called

src/Form/grid/Element/index.tsx
```diff
@@ -273,8 +273,11 @@ const Element = ({ node: el, form }: any) => {
             {...fieldProps}
             value={fieldVal}
             onChange={(val: any) => {
-              const change = changeValue(val, el, index);
-              if (change) onChange();
+              changeValue(val, el, index, false, false);
+            }}
+            onComplete={(val: any) => {
+              changeValue(val, el, index);
+              onChange();
```
bo-dun-1, line 280: should we only run onChange() when changeValue returns truthy?

src/elements/fields/DateSelectorField/index.tsx
```diff
@@ -138,10 +140,15 @@ function DateSelectorField({
     return !disabledDates.includes(`${date.getMonth() + 1}-${date.getDate()}`);
   };
 
-  const onDateChange = (newDate: any) => {
+  // Called when the input is changed
+  const onDateChange = (newDate: any, isComplete = false) => {
```
bo-dun-1, line 144: const callback = isComplete ? onComplete : onChange;
newDate = newDate ?? '';
setInternalDate(newDate);
callback(formatDateString(newDate, servarMeta));

bo-dun-1 (approved):

PR #1216, opened 2024-05-08
add readonly prop to form

trebitowski:
Adds a readonly prop for `<Form>` that allows step navigation but no edits, validation, or submitting.

### DIFF
--- a/src/Form/index.tsx
+++ b/src/Form/index.tsx
@@ -1501,7 +1501,8 @@ function Form({
     } as Trigger;
     let submitPromise: Promise<any> = Promise.resolve();
 
-    if (submit) {
+    // If the step is readOnly, don't run validation or submit
+    if (submit && !readOnly) {
       // Do not proceed until user has gone through required flows
       if (
         !hasFlowActions(actions) &&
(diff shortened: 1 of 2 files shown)


src/Form/index.tsx
```diff
@@ -1509,41 +1513,45 @@ function Form({
...
-        setInlineErrors,
-        trigger
-      });
-      if (invalid) {
-        setAutoValidate(true);
-        elementClicks[id] = false;
-        return;
-      }
+      // If the step is readonly, don't run validation or submit
+      if (!readonly) {
```
bo-dun-1, line 1517: IMO this should be moved to line 1504, becoming `if (submit && !readonly)`

src/Form/index.tsx
```diff
@@ -193,6 +193,7 @@ export interface Props {
   className?: string;
   children?: JSX.Element;
   _draft?: boolean;
+  readonly?: boolean;
```
bo-dun-1, line 196: should this technically be readOnly since it's two words

src/Form/index.tsx
```diff
@@ -269,6 +271,8 @@ function Form({
   // No state since off reason is set in two locations almost simultaneously
   const formOffReason = useRef('');
   const [formSettings, setFormSettings] = useState({
+    readOnly: readOnly,
```
bo-dun-1, line 274: just readOnly,

bo-dun-1 (approved):

PR #1223, opened 2024-05-24
fix: iOS radio button and checkbox error message

trebitowski:
This fixes a bug where iOS devices do not scroll to radio buttons or checkboxes on focus. The bug was causing confusion where a user would receive no feedback of missing fields when trying to submit a step.

### Solution
Added a focus handler for all inputs of type `checkbox` or `radio`. On focus, if the input is not in the viewport and the device is running iOS, we manually scroll to the input. This matches the behavior on desktop and android devices.

### Testing
**Devices:** [Android - Chrome, Macbook - Chrome, iPhone 9 - Safari]
**Form:** https://form.feathery.io/to/pRq7c7 

A three step form where each step is either a validated radio group, checkbox group, or matrix. The submit button is below and requires scrolling past the input to press.

For each step I scrolled to the bottom and attempted to submit. The input would show a required error and scrolled into view. The behavior on Android, iOS, and Mac OS were equivalent.

### DIFF
--- a/src/elements/fields/RadioButtonGroupField.tsx
+++ b/src/elements/fields/RadioButtonGroupField.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo, useState } from 'react';
+import React, { useMemo, useState, FocusEvent } from 'react';
 import ReactForm from 'react-bootstrap/Form';
 import { bootstrapStyles } from '../styles';
 import {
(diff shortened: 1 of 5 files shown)


src/elements/fields/RadioButtonGroupField.tsx
```diff
@@ -63,6 +65,23 @@ function RadioButtonGroupField({
     }));
   }
 
+  // iOS devices do not scroll to focused radio buttons
+  // we manually scroll to the focused radio button to
+  // handle form errors and provide the same experience other devices have
+  const scrollOnFocus = (event: FocusEvent<HTMLInputElement, Element>) => {
```
bo-dun-1, line 71: since all of the scrollOnFocus functions are mostly the same, can we make it a browser utility

bo-dun-1 (approved):

PR #1225, opened 2024-05-28
fix: android keyboard breaking suggestion

trebitowski:
Fixes a bug where android keyboard would lose their place while typing and cause double letters when using keyboard suggestions.

### Issue
On embedded forms, the android keyboard was losing track of the current word whenever the first character was typed. This caused it to not fully replace the word when a word suggestion was clicked.
### Solution
By moving the onChange logic out to the Element instead of inside the TextField we create a workaround for the issue, where rerenders are now triggered properly. Further investigation and fixes need to be made in the future to fix the underlying issue with embedded forms, rerenders, and field state.
### Testing
1. [Created a form](https://app.feathery.io/forms/ff6300bd-6e29-48ec-825f-b37412ea040b/) which tests each property of a text field. 
2. Completed the form on an embedded static page running a local version of the React library. The keyboard issue was no longer present. 
3. Regression tested the text field functionality by successfully completing the form. Repeatable placeholders worked as they should.

### DIFF
--- a/src/Form/grid/Element/index.tsx
+++ b/src/Form/grid/Element/index.tsx
@@ -673,8 +673,14 @@ const Element = ({ node: el, form }: any) => {
         return (
           <Elements.TextField
             {...fieldProps}
-            onChange={(val: string) => {
-              const change = changeValue(val, el, index, false, false);
+            onAccept={(val: any, mask: any) => {
+              const newVal = mask._unmaskedValue === '' ? '' : val;
+              if (newVal === stringifyWithNull(fieldVal)) return;
+
+              const isOrWasEmpty = !newVal || !stringifyWithNull(fieldVal);
+              const rerender =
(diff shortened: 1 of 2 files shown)


src/Form/grid/Element/index.tsx
```diff
@@ -673,8 +673,14 @@ const Element = ({ node: el, form }: any) => {
         return (
           <Elements.TextField
             {...fieldProps}
-            onChange={(val: string) => {
-              const change = changeValue(val, el, index, false, false);
+            onAccept={(val: any, mask: any) => {
+              const newVal = mask._unmaskedValue === '' ? '' : val;
+              if (newVal === stringifyWithNull(fieldVal)) return;
+
+              const isOrWasEmpty = !newVal || !stringifyWithNull(fieldVal);
```
bo-dun-1, line 680: no need to call stringifyWithNull here

bo-dun-1 (approved):

PR #1243, opened 2024-06-26
feat: render pdfs in image element

trebitowski:
Adds PDF rendering capabilities to the image element. If a file upload field has a pdf uploaded, and an image element is mapped to the field, the pdf will be embedded instead of an image. If an image is uploaded, the image is displayed as before.

Removes the container's functionality to render uploaded PDFs

image

image


src/elements/basic/ImageElement.tsx
```diff
@@ -19,30 +19,39 @@ function ImageElement({
...
       }
-    } else setImageUrl(element.properties.source_image);
+    } else {
+      setDocumentUrl(element.properties.source_image);
+      setDocumentType('');
+    }
   }, [imageFieldSource, element.properties.source_image]);
 
+  const displayPDF = documentUrl && documentType === 'application/pdf';
+  const displayImage = !fieldKey || (documentUrl && !displayPDF);
```
bo-dun-1, line 54: is this needed? since we can just do !displayPDF. If the image / field doesn't exist, our existing behavior is we still have the image element render a placeholder image

PR #1245, opened 2024-06-27
fix: multiselect dropdown expands past max-width

trebitowski:
Fixes some sizing issues on multi-dropdown fields with long option labels, where the field width would expand and break user experience, especially on mobile devices.

### Changes
Removes css property `min-width: min-content` on the container around a multi-dropdown. This property didn't work well with the composition of the dropdown component and was causing issues where long labels would force the field to expand horizontally. By removing this property, we allow the the labels to wrap, and keep the field width to its expected size.

Adds extra padding so that the selected options do not expand into the dropdown indicator chevron on the right.

Very long labels can now wrap onto multiple lines in order to fit. There is a maximum of three lines before the remaining content is truncated.

image

### DIFF
--- a/src/Form/grid/StyledContainer/styles.ts
+++ b/src/Form/grid/StyledContainer/styles.ts
@@ -87,7 +87,7 @@ export const getContainerStyles = (
       if (contentResponsive)
         contentResponsive = getElementType(node) !== 'button_group';
 
-      // if element is mutli dropdown, don't apply min-width: min-content
+      // if element is multi dropdown, don't apply min-width: min-content
       if (getElementType(node) !== 'dropdown_multi') s.minWidth = 'min-content';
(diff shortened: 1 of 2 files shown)


src/Form/grid/StyledContainer/styles.ts
```diff
@@ -87,7 +87,8 @@ export const getContainerStyles = (
       if (contentResponsive)
         contentResponsive = getElementType(node) !== 'button_group';
 
-      s.minWidth = 'min-content';
+      // if element is mutli dropdown, don't apply min-width: min-content
```
ondrosh, line 90: Typo here on `multi`

ondrosh (approved):

bo-dun-1 (approved):

PR #1249, opened 2024-07-16
Tyler/otp

trebitowski:
Adds a listener for otp sms codes and autofills pin inputs with them


src/hooks/useOTPListener.tsx
```diff
@@ -0,0 +1,38 @@
+import { useEffect, useRef } from 'react';
```
bo-dun-1, line 1: i wonder if it would make sense to make PinInputField a folder like the TextField folder, then put the hook in there

bo-dun-1 (approved):

PR #1255, opened 2024-07-18
QR Code Scanner Reimplementation

trebitowski:
Reimplementation of the QR Code Scanner element to improve on reliability and functionality.

**Changes**
- Handles the component rendering ourselves instead of relying on the library's standard UI. This allows us to customize the functionality and UI.
- Improved setup and cleanup logic so that the camera is running only when necessary.
- Smarter default camera. Before, the user could pick any camera. This allowed them to choose a bad camera (like their phone's wide angle camera) that would not be able to scan a code and would frustrate the user. Now we pick the standard, backward-facing camera if available by default.
- Zoom functionality. Because we have access to the underlying media stream, we can apply camera settings to the video feed. This means that if a camera has zoom enabled, we show a slider input to control the zoom.
- Scan exits immediately after success. Once the scanner detects a code, it will automatically close, matching the behavior users would expect and stopping the field value from being cleared if the code isn't detected next frame.

image


src/elements/fields/QRScanner/index.tsx
```diff
@@ -1,27 +1,42 @@
...
-  if (
-    errorEl?.textContent?.trim() ===
-    'D: No MultiFormat Readers were able to detect the code.'
-  ) {
-    errorEl.textContent =
-      'No QR code detected. Please try with a different image.';
+async function createScanner(id: string) {
+  await qrPromise;
+  const window = featheryWindow();
+  for (let i = 0; i < 3; i++) {
```
bo-dun-1, line 24: is this for loop relevant anymore given we're no longer using the original library? i feel like we could just await on qrPromise and once it's loaded, we can intiialize the scanner which should be guarnteed to work

PR #1262, opened 2024-07-31
feat: update generateEnvelopes payload

johnlapp:
- update the payload for generateEnvelopes request to reflect the Quik Forms integration changes on the backend

### DIFF
--- a/src/utils/featheryClient/integrationClient.ts
+++ b/src/utils/featheryClient/integrationClient.ts
@@ -353,7 +353,13 @@ export default class IntegrationClient {
     };
 
     if (action.quik_tags_field_key) {
-      payload.quik_tags = fieldValues[action.quik_tags_field_key];
+      if (typeof fieldValues[action.quik_tags_field_key] === 'string') {
+        payload.quik_tags = (fieldValues[action.quik_tags_field_key] as string)
+          .split(',')
+          .map((tag) => tag.trim());
+      } else {
+        payload.quik_tags = fieldValues[action.quik_tags_field_key];
+      }
     }
 
     const url = `${API_URL}document/form/generate/`;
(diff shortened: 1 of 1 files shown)


src/utils/featheryClient/integrationClient.ts
```diff
@@ -319,12 +319,17 @@ export default class IntegrationClient {
...
       documents: action.documents ?? [],
-      signer_email: signer
+      signer_email: signer,
+      use_quik: action.quik_documents
     };
-    if (action.quik_documents)
-      payload.quik = fieldValues[action.quik_json_field_key];
+
+    if (action.quik_tags_field_key) {
+      payload.quik_tags = fieldValues[action.quik_tags_field_key];
```
bo-dun-1, line 330: let's add logic to parse the quik tags depending on the type of the field value that's being fetched here. i left the specific details in a comment on your backend PR

src/utils/featheryClient/integrationClient.ts
```diff
@@ -345,12 +345,23 @@ export default class IntegrationClient {
...
-    if (action.quik_documents)
-      payload.quik = fieldValues[action.quik_json_field_key];
+
+    if (action.quik_tags_field_key) {
+      if (typeof fieldValues[action.quik_tags_field_key] === 'string') {
+        payload.quik_tags = (fieldValues[action.quik_tags_field_key] as string)
+          .split(',')
+          .map((tag) => tag.trim());
+      } else {
+        payload.quik_tags = fieldValues[action.quik_tags_field_key];
```
bo-dun-1, line 361: this only works if the type is an array. we should also handle the case where the field value isn't an array or a string, in which case we should stringify it and wrap it with an array. also cleaner to do

const fieldVal = fieldValues[action.quik_tags_field_key];

and use that variable instead

bo-dun-1 (approved):

PR #1264, opened 2024-08-03
fix: date picker typed input

trebitowski:
Fixes an issue where date parsing would not work if the user typed their date without slashes. Now the input is masked so all input is properly formatted.

Also fixes an issue where the date might not update if onChange was called before onComplete with the same date while typing. Now the change is only triggered once the date is completely selected and parsed. This ensures that the date value is carried out through the form. It also stops extra onChange events from firing while the user is typing their date.

Testing form: https://form.feathery.io/to/zt3lqY

**Testing:**
- Date formats work with MM/DD/YYYY and DD/MM/YYYY
- Date restrictions work, specific dates, no weekends, past and future only
- Time formats work 12 hour format with AM/PM and 24 hour format
- Time restrictions work, earliest and latest time options

### DIFF
--- a/src/elements/fields/DateSelectorField/index.tsx
+++ b/src/elements/fields/DateSelectorField/index.tsx
@@ -297,7 +297,7 @@ const CustomMaskedInput = React.forwardRef((props: any, ref) => {
       blocks: {
         dd: { mask: IMask.MaskedRange, from: 1, to: 31, maxLength: 2 },
         MM: { mask: IMask.MaskedRange, from: 1, to: 12, maxLength: 2 },
-        yyyy: { mask: IMask.MaskedRange, from: 1900, to: 2099 },
+        yyyy: { mask: IMask.MaskedRange, from: 1, to: 9999, maxLength: 4 },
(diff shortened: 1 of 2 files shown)


src/elements/fields/DateSelectorField/index.tsx
```diff
@@ -286,3 +288,37 @@ function DateSelectorField({
...
+
+const CustomMaskedInput = React.forwardRef((props: any, ref) => {
+  const { onChange, dateMask, ...rest } = props;
+  const maskOptions = useMemo(
+    () => ({
+      mask: dateMask,
+      blocks: {
+        dd: { mask: IMask.MaskedRange, from: 1, to: 31, maxLength: 2 },
+        MM: { mask: IMask.MaskedRange, from: 1, to: 12, maxLength: 2 },
+        yyyy: { mask: IMask.MaskedRange, from: 1900, to: 2099 },
```
bo-dun-1, line 300: do we want to restrict it to this range? what about 1 to 9999?

trebitowski: Yeah good call. I made the range 1-9999.

bo-dun-1 (approved):

PR #1274, opened 2024-08-16
feat: quik docusign integration

johnlapp:
- add docusign params to quik payload

### DIFF
--- a/src/utils/featheryClient/integrationClient.ts
+++ b/src/utils/featheryClient/integrationClient.ts
@@ -371,18 +371,12 @@ export default class IntegrationClient {
     const payload: Record<string, any> = {
       form_key: this.formKey,
       fuser_key: userId,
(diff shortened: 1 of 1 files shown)


src/utils/featheryClient/integrationClient.ts
```diff
@@ -378,8 +378,27 @@ export default class IntegrationClient {
       if (!action.html_url_field)
         throw Error('No field selected for storing HTML URL returned by Quik!');
 
-      payload.html_url_field = action.html_url_field;
-      payload.html_url_field_type = action.html_url_field_type;
+      Object.assign(payload, {
+        html_url_field: action.html_url_field,
```
bo-dun-1, line 382: honestly feels like we can just pass the entire action into the payload and backend can parse what itneeds

bo-dun-1 (approved):

PR #1290, opened 2024-09-30
Flinks iframe url changed to only fetch 90 days of transaction history

EricHuang2FG:


src/integrations/flinks.ts
```diff
@@ -23,6 +23,9 @@ export async function openFlinksConnect(
       ? 'toolbox'
       : flinksConfig.metadata.instance;
 
+  const daysOfTransactions =
+    parseInt(flinksConfig.metadata.daysTransactions) || 365;
```
bo-dun-1, line 27: instead of hardcoding the default number in the react lib, yous hould instead pass back null to the backend and the backend can set the default number

PR #1303, opened 2024-10-23
feat: Show quik form bundle within modal

johnlapp:


src/Form/index.tsx
```diff
@@ -341,6 +342,9 @@ function Form({
   // Set to trigger conditional renders on field value updates, no need to use the value itself
   const [render, setRender] = useState({ v: 1 });
 
+  const [showQuikFormViewer, setShowQuikFormViewer] = useState(false);
```
bo-dun-1, line 345: i think this quik state and hte quik component can all be merged into one quik hook. Something like:

const {setQuikPayload, showQuikViewer, quikViewer} = useQuik(...);

bo-dun-1 (approved):

PR #1352, opened 2025-03-12
feat: Support field set extractions

trebitowski:
Adds support for field set overrides in document extractions.

Previously the function signature was `runAIExtraction(extractionId: string, runAsync: boolean, pages?: number[])`. We would have needed to add overrideId as another optional, unnamed parameter.

Instead, now runAIExtraction accepts an options object as the second parameter, where waitForCompletion, pages, and overrideId can be configured: 
```
runAIExtraction(extractionId: string, options: {
   waitForCompletion: boolean; 
   pages?: number[]; 
   overrideId?: string;
})
```
There is code to support both the original parameters and the new options parameters, depending on if the second parameter is an object.

### DIFF
--- a/src/Form/index.tsx
+++ b/src/Form/index.tsx
@@ -128,6 +128,7 @@ import { getPrivateActions } from '../utils/sensitiveActions';
 import { v4 as uuidv4 } from 'uuid';
 import internalState, {
   ApplyAlloyJourney,
+  ExtractionActionOptions,
   RunIntegrationActions,
   setFormInternalState
 } from '../utils/internalState';
@@ -963,13 +964,7 @@ function Form({
         },
         runAIExtraction: async (
           extractionId: string,
-          options:
-            | {
-                waitForCompletion?: boolean;
-                pages?: number[];
-                variantId?: string; // uuid
-              }
-            | boolean,
+          options: ExtractionActionOptions | boolean,
           // deprecated, pages should be in options
           pages?: number[]
         ) => {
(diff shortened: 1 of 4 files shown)


src/utils/formContext.ts
```diff
@@ -143,9 +143,9 @@ export const getFormContext = (formUuid: string) => {
     ) => formState.runIntegrationActions(actionIds, options),
     runAIExtraction: async (
       extractionId: string,
-      waitForCompletion = false,
+      options = { waitForCompletion: true },
       pages?: number[]
-    ) => formState.runAIExtraction(extractionId, !waitForCompletion, pages),
+    ) => formState.runAIExtraction(extractionId, options, pages),
```
bo-dun-1, line 148: doesn't waitForCompletion still need to default to false to be backwards compatible?

src/Form/index.tsx
```diff
@@ -963,7 +963,14 @@ function Form({
         },
         runAIExtraction: async (
           extractionId: string,
-          runAsync: boolean,
+          options:
+            | {
+                waitForCompletion: boolean;
```
bo-dun-1, line 968: think this should also be optional

src/Form/index.tsx
```diff
@@ -1949,6 +1956,7 @@ function Form({
           await submitPromise;
           const data = await client.extractAIDocument({
             extractionId: action.extraction_id,
+            extractionVariantId: action.variant_id,
```
bo-dun-1, line 1959: i dont think this is the right way to call the new client method definition

bo-dun-1 (approved):

PR #1355, opened 2025-03-17
feat: Add formName and formId

trebitowski:
Adds support for `feathery.formName` and `feathery.formId` in code logic rules.

### DIFF
--- a/src/Form/index.tsx
+++ b/src/Form/index.tsx
@@ -879,9 +879,9 @@ function Form({
           )
         ),
         trackHashes: trackHashes.current,
-        formRef,
         formId,
         formName,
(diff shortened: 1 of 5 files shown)


src/Form/index.tsx
```diff
@@ -887,6 +887,8 @@ function Form({
         setInlineErrors,
         setUserProgress,
         steps,
+        formName: formName,
```
bo-dun-1, line 890: just formName, formId

bo-dun-1: nit: maybe move these right above formRef in this list of args

bo-dun-1 (approved):

PR #1380, opened 2025-04-30
fix: Firebase auth

trebitowski:
replaces firebase auth redirect flow with popup flow instead. Since popup flow does not trigger a form reload like redirecting does, we need to manually navigate to the login step on auth success.

### DIFF
--- a/src/Form/index.tsx
+++ b/src/Form/index.tsx
@@ -193,6 +193,7 @@ import { isNum } from '../utils/primitives';
 import { getSignUrl } from '../utils/document';
 import QuikFormViewer from '../elements/components/QuikFormViewer';
 import { createSchwabContact } from '../integrations/schwab';
+import { getLoginStep } from '../auth/utils';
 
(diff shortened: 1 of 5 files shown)


src/Form/index.tsx
```diff
@@ -1526,6 +1526,22 @@ function Form({
     }
   };
 
+  const goToLoginStep = async () => {
```
bo-dun-1, line 1529: can we put this in the auth folder and import it in here to use

src/auth/utils.tsx
```diff
@@ -0,0 +1,8 @@
+import { getAuthIntegrationMetadata } from './internal/utils';
+
+export const getLoginStep = (steps: any, integrations: any) => {
+  const authIntegration = getAuthIntegrationMetadata(integrations);
+  return Object.values(steps).find(
```
bo-dun-1, line 5: can't you just do steps[authIntegration.login_step]?

PR #1405, opened 2025-06-04
feat: Handle casting hidden field values

arykhoda:


src/utils/formHelperFunctions.ts
```diff
@@ -889,23 +893,25 @@ export function saveInitialValuesAndUrlParams({
...
   // Submit initial values & URL params
   let valuesToSubmit: Record<string, any> = {};
   if (!isObjectEmpty(initialValues)) {
     rerenderRequired = true;
-    const servarAttrMap = getServarAttrMap(steps);
+    const fieldAttrMap = getAllFieldAttrMap(steps, hiddenFields);
     valuesToSubmit = { ...initialValues };
     Object.entries(valuesToSubmit).map(([key, val]) => {
-      const attrs = servarAttrMap[key] ?? {};
+      const attrs = fieldAttrMap[key];
```
bo-dun-1, line 914: maybe we should leave the `?? {}` actually just to be safe

src/utils/formHelperFunctions.ts
```diff
@@ -785,24 +785,19 @@ export function getInitialStep({
   return initialStepId || sessionCurrentStep || (getOrigin as any)(steps).key;
 }
 
-export function castVal(
-  servarType: string | undefined,
-  val: any,
-  repeated = false
-): any {
+export function castVal(fieldType: string, val: any, repeated = false): any {
```
bo-dun-1, line 788: can we make separate functions castServarVal and castHiddenVal instead of putting them in the same one?

PR #1416, opened 2025-06-13
feat: Add dropdown sync feature

arsalansids:
Adding the ability for Dropdown and MultiDropdown components to fetch dynamic field values from Salesforce.

### DIFF
--- a/src/elements/fields/DropdownField.tsx
+++ b/src/elements/fields/DropdownField.tsx
@@ -8,7 +8,7 @@ import { getStateOptions, hasState } from '../components/data/states';
 import { css, Global } from '@emotion/react';
 import { hoverStylesGuard, iosScrollOnFocus } from '../../utils/browser';
 import { fieldValues } from '../../utils/init';
-import FeatheryClient from '../../utils/featheryClient';
+import useSalesforceSync from '../../hooks/useSalesforceSync';
 
 export default function DropdownField({
   element,
@@ -36,8 +36,9 @@ export default function DropdownField({
   const containerRef = useRef(null);
   const servar = element.servar;
   const short = servar.metadata.store_abbreviation;
-  const [dynamicOptions, setDynamicOptions] = useState<string[]>([]);
-  const [loadingDynamicOptions, setLoadingDynamicOptions] = useState(false);
+  const { dynamicOptions, loadingDynamicOptions } = useSalesforceSync(
+    servar.metadata.salesforce_sync
+  );
 
   useEffect(() => {
     if (servar.type === 'gmap_state') {
@@ -49,27 +50,6 @@ export default function DropdownField({
       setCurCountry(code || 'us');
     }
   }, [countryCode, setCurCountry]);
-
-  useEffect(() => {
-    const salesforceSync = servar.metadata.salesforce_sync;
-    if (!salesforceSync) return;
-
-    const fetchSalesforceOptions = async () => {
-      setLoadingDynamicOptions(true);
-      try {
-        const client = new FeatheryClient();
-        const data = await client.fetchSalesforceFieldValues(salesforceSync);
-        setDynamicOptions(data.options || []);
-      } catch (error) {
-        console.error('Failed to fetch Salesforce options:', error);
-        setDynamicOptions([]);
-      } finally {
-        setLoadingDynamicOptions(false);
-      }
-    };
-
-    fetchSalesforceOptions();
-  }, [servar.metadata.salesforce_sync]);
 
   let options;
   if (dynamicOptions.length > 0) {
--- a/src/elements/fields/DropdownField.tsx
+++ b/src/elements/fields/DropdownField.tsx
@@ -54,8 +54,8 @@ export default function DropdownField({
   let options;
   if (dynamicOptions.length > 0) {
     options = dynamicOptions.map((option) => (
-      <option key={option} value={option}>
-        {option}
+      <option key={option.value} value={option.value}>
+        {option.label}
       </option>
     ));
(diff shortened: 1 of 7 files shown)


src/elements/fields/DropdownMultiField.tsx
```diff
@@ -92,11 +118,16 @@ export default function DropdownMultiField({
...
-  const servar = element.servar;
   const labels = servar.metadata.option_labels;
   const labelMap: Record<string, string> = {};
   let options;
-  if (
+
+  if (dynamicOptions.length > 0) {
+    options = addFieldValOptions(dynamicOptions).map((option: string) => {
+      labelMap[option] = option;
+      return { value: option, label: option, tooltip: '' };
```
arsalansids, line 128: So value and labels can be different in Salesforce too, for example the value can be `AdminRole` with label being `Admin Role`. Is the label sufficient or do we need the value as well?

trebitowski: It should match Salesforce IMO, so it displays the label and the value is saved

src/elements/fields/DropdownField.tsx
```diff
@@ -47,8 +50,35 @@ export default function DropdownField({
...
+        console.error('Failed to fetch Salesforce options:', error);
+        setDynamicOptions([]);
+      } finally {
+        setLoadingDynamicOptions(false);
+      }
+    };
+
+    fetchSalesforceOptions();
+  }, [servar.metadata.salesforce_sync]);
+
```
trebitowski, line 73: It would be nice if the effect and state were contained in a custom hook.

src/utils/featheryClient/integrationClient.ts
```diff
@@ -718,4 +718,25 @@ export default class IntegrationClient {
       return { ok: true, payload: await res.json() };
     else return { ok: false, error: (await res?.text()) ?? '' };
   }
+
+  async fetchSalesforceFieldValues(salesforceSync: {
```
bo-dun-1, line 722: fetchSalesforcePicklistOptions?

src/utils/featheryClient/integrationClient.ts
```diff
@@ -718,4 +718,25 @@ export default class IntegrationClient {
...
+    object_name: string;
```
bo-dun-1, line 723: why not just pass these parameters in directly instead of making it an object?

src/utils/featheryClient/integrationClient.ts
```diff
@@ -718,4 +718,25 @@ export default class IntegrationClient {
...
+    field_name: string;
+    credential_key: string;
+  }) {
+    const url = `${API_URL}integration/salesforce/field/values/`;
+    const params = new URLSearchParams({
+      object_name: salesforceSync.object_name,
+      field_name: salesforceSync.field_name,
+      credential_key: salesforceSync.credential_key
+    }).toString();
+    const response = await this._fetch(`${url}?${params}`, { method: 'GET' });
```
bo-dun-1, line 733: method: 'GET' not necessary, it defaults to that

src/elements/fields/DropdownMultiField.tsx
```diff
@@ -92,11 +97,16 @@ export default function DropdownMultiField({
...
-  const servar = element.servar;
   const labels = servar.metadata.option_labels;
   const labelMap: Record<string, string> = {};
   let options;
-  if (
+
+  if (dynamicOptions.length > 0) {
+    options = dynamicOptions.map((option) => {
+      labelMap[option.value] = option.label;
+      return { value: option.value, label: option.label, tooltip: '' };
```
bo-dun-1, line 107: tooltip is an optional property, no need to include

src/utils/featheryClient/integrationClient.ts
arsalansids: Really for readability and that they're 3 different parameters that made sense to group together since they're all salesforce related values. If the team prefers explicit parameters, specifically for typechecking, I'm happy to change it

bo-dun-1: i find this pattern is nicer for a set of optional options, but when they're all required plus only three params, might be easier to make it explicit

src/utils/featheryClient/integrationClient.ts
```diff
@@ -718,4 +718,25 @@ export default class IntegrationClient {
...
+    const url = `${API_URL}salesforce/field/options/`;
+    const params = new URLSearchParams({
+      object_name: salesforceSync.object_name,
+      field_name: salesforceSync.field_name,
+      credential_key: salesforceSync.credential_key
+    }).toString();
+    const response = await this._fetch(`${url}?${params}`);
+    if (response && response.ok) {
+      const data = await response.json();
+      return {
```
bo-dun-1, line 736: could we just return data directly?

bo-dun-1 (approved):

src/elements/fields/DropdownField.tsx
```diff
@@ -35,6 +36,9 @@ export default function DropdownField({
   const containerRef = useRef(null);
   const servar = element.servar;
   const short = servar.metadata.store_abbreviation;
+  const { dynamicOptions, loadingDynamicOptions } = useSalesforceSync(
```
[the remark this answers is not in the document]
arsalansids: This will be addressed in subsequent PRs

bo-dun-1 (approved):

src/elements/fields/DropdownField.tsx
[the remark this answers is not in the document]
arsalansids: Nevermind, this is now addressed

PR #1420, opened 2025-06-20
fix: Do not load manual options, turn off fetch in edit mode

arsalansids:
## Changes
Do not load manual options for dynamic options fields

- [x] Tested end to end
refresh loading test

## Related pull requests

### DIFF
--- a/src/elements/fields/DropdownField.tsx
+++ b/src/elements/fields/DropdownField.tsx
@@ -37,7 +37,8 @@ export default function DropdownField({
   const servar = element.servar;
   const short = servar.metadata.store_abbreviation;
   const { dynamicOptions, loadingDynamicOptions } = useSalesforceSync(
-    servar.metadata.salesforce_sync
+    servar.metadata.salesforce_sync,
+    editMode
   );
 
   useEffect(() => {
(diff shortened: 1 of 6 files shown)


src/elements/fields/DropdownMultiField.tsx
```diff
@@ -109,13 +110,16 @@ export default function DropdownMultiField({
...
-  let options;
+  let options: any[] = [];
 
-  if (dynamicOptions.length > 0) {
-    options = dynamicOptions.map((option) => {
-      labelMap[option.value] = option.label;
-      return { value: option.value, label: option.label };
-    });
+  if (servar.metadata.salesforce_sync && !editMode) {
+    if (loadingDynamicOptions) options = [];
```
bo-dun-1, line 116: idk if the loadingDynamicOptions needs to be returned from the hook. You could just check this internal to the hook and return `[]` for the options if it's true

src/elements/fields/DropdownField.tsx
```diff
@@ -52,7 +53,7 @@ export default function DropdownField({
   }, [countryCode, setCurCountry]);
 
   let options;
-  if (dynamicOptions.length > 0) {
+  if (servar.metadata.salesforce_sync && !editMode) {
```
bo-dun-1, line 56: maybe this should be a boolean returned from the hook as well, like `shouldSyncSalesforce`

PR #1435, opened 2025-07-25
fix: Step submission

trebitowski:
## Changes
* Failed requests are now retried, even if the user was always online.
* Requests are retried with exponential backoff.
* Failed requests are never dropped.
* If a request fails the retry after 5 attempts we show an alert to the user, warning them that their results aren't being saved. (The user can either refresh or continue with this submission)
image

* The form completion event is blocked until all pending/failed requests are successful.
* Once the user is back online/requests are succeeding, the user is able to try submitting their step/form again and all step data will be submitted (in the correct order).

- [x] Tested end to end

## Related pull requests

### Set up

* Set up a multi step form with field and file uploads

* All scenarios tested on `hosted-forms` and with a html page with js embedded form

* Set up network request blocking to mock individual requests failing
image

* Set up network throttling to mock offline behavior
image

# Test Scenarios

## Offline Scenarios

| Scenario | Description | Expected Behavior | Result |
|----------|-------------|-------------------|---------|
| **1. Online Success** | User is online and all requests succeed | Submission is stored and marked as completed | Success |
| **2. Offline Form Completion** | User is offline and tries to complete form | User cannot complete form, all requests are saved | Offline |
| **2a. Return Online** | User comes back online after offline attempt | User is able to submit successfully | Back Online |
| **3. Temporary Offline** | User goes offline for a step, but comes back online before completing form | Form completes successfully | Temporary Offline |
| **4. Permanently Offline** | User goes offline after first step and never comes back online | Form remains incomplete | Permanently Offline |

## Failed Request Scenarios

| Scenario | Description | Expected Behavior | Result |
|----------|-------------|-------------------|---------|
| **5. Step Submit Failure** | User is online, but step submit request fails | User cannot complete form initially | Step Fail |
| **5a. Recovery** | Requests are unblocked after initial failure | User completes form successfully | Recovery |
| **6. Retry Success** | User is online, step submit fails initially but succeeds on retry | Form completes successfully after retry | Retry Success |
| **7. Multiple Step Failures** | User is online but multiple step submits fail until unblocked | Form completes after failures are resolved | Multiple Failures |
| **8. Final Submit Failure (Recovery)** | User is online, final submit fails until unblocked | Form completes after final submit is unblocked | Final Submit Recovery |
| **9. Final Submit Failure (No Recovery)** | User is online but only final submit fails and user never completes | Form remains incomplete | Final Submit Fail |

### DIFF
--- a/src/Form/index.tsx
+++ b/src/Form/index.tsx
@@ -1053,7 +1053,7 @@ function Form({
   useEffect(() => {
     if (clientRef.current) return;
 
-    const networkErrorCallback = (error: string) => {
+    const networkErrorCallback = () => {
       if (!showNetworkErrorAlert.current) return;
       showNetworkErrorAlert.current = false;
       featheryWindow().alert(
--- a/src/Form/index.tsx
+++ b/src/Form/index.tsx
@@ -294,7 +294,6 @@ function Form({
   const [formName, setFormName] = useState(formNameProp || ''); // TODO: remove support for formName (deprecated)
   const formKey = formId || formName; // prioritize formID but fall back to name
   const clientRef = useRef<any>(undefined);
-  const showNetworkErrorAlert = useRef<any>(true);
   const client = clientRef.current;
   const navigate = useNavigate();
   const location = useLocation();
@@ -1053,20 +1052,11 @@ function Form({
   useEffect(() => {
     if (clientRef.current) return;
 
-    const networkErrorCallback = () => {
-      if (!showNetworkErrorAlert.current) return;
-      showNetworkErrorAlert.current = false;
-      featheryWindow().alert(
-        'There was a network error while submitting the form. Please refresh the page and try again.'
-      );
-    };
-
     clientRef.current = new FeatheryClient(
       formKey,
       hasRedirected,
       _draft,
-      _bypassCDN,
-      networkErrorCallback
+      _bypassCDN
     );
     const newClient = clientRef.current;
     let saveUrlParamsFormSetting = false;
@@ -1494,7 +1484,7 @@ function Form({
         if (submitPromise) await submitPromise;
 
         // Check if there are any failed requests before completing
-        if (client.offlineRequestHandler.hasPendingFailedRequests()) {
+        if (await client.offlineRequestHandler.dbHasRequest()) {
           return;
         }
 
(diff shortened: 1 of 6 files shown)


src/Form/index.tsx
```diff
@@ -1370,7 +1367,11 @@ function Form({
     if (hasSubmitAfter) {
       // If running submit logic rule after, must finish
       // submit data first
-      await stepPromise;
+      try {
+        await stepPromise;
```
bo-dun-1, line 1371: why aren't we bubbling the error if stepPromise fails?

src/Form/index.tsx
```diff
@@ -1920,11 +1918,21 @@ function Form({
       } else if (type === ACTION_LOGOUT) await Auth.inferAuthLogout();
       else if (type === ACTION_NEW_SUBMISSION) await updateUserId(uuidv4());
       else if (type === ACTION_NEXT) {
+        // Wait for step submission to succeed
+        if (submitPromise && submitPromise !== undefined) {
+          try {
+            await submitPromise;
```
bo-dun-1, line 1924: we shouldn't do this. we very deliberately didn't await the promise here because we wanted navigation to be snappy and not depend on the submit returning. My proposal was that we just check the offline flag here instead

src/utils/featheryClient/index.ts
```diff
@@ -707,33 +686,15 @@ export default class FeatheryClient extends IntegrationClient {
...
-                return new Promise((resolve) => setTimeout(resolve, 5000));
-              throw e;
-            })
-          ),
-        url,
-        options,
-        'registerEvent',
-        stepKey
+      this.submitQueue.then(() =>
+        wrapUnload(() =>
```
bo-dun-1, line 695: if you're removing the typerror await handler, saad should be looped in for this. we added it initially for hiscox.

src/utils/offlineRequestHandler.ts
```diff
@@ -432,6 +459,12 @@ export class OfflineRequestHandler {
     }
   }
 
+  private getDelay(attemptNum: number): number {
```
bo-dun-1, line 462: getExponentialDelay?

bo-dun-1 (approved):

PR #1446, opened 2025-08-20
feat: Added generateDocument logic action

arykhoda:

- [ ] Tested end to end

## Related pull requests


src/utils/featheryClient/index.ts
```diff
@@ -994,4 +994,21 @@ export default class FeatheryClient extends IntegrationClient {
...
+    const data = await this.generateEnvelopes(action);
+    const envAction = action.envelope_action;
+
+    if (envAction === 'download') {
+      await downloadAllFileUrls(data.files);
+    } else if (envAction === 'save') {
+      let files = data.files;
+      if (files.length === 1) files = files[0];
+      const newValues = { [action.save_document_field_key]: files };
+      setFieldValues(newValues);
```
arykhoda, line 1008: not sure if setFieldValues is the right function to use here

src/utils/featheryClient/index.ts
```diff
@@ -994,4 +994,39 @@ export default class FeatheryClient extends IntegrationClient {
...
+        body: JSON.stringify(payload)
+      },
+      false
+    ).then(async (response) => {
+      if (response) {
+        if (response.ok) {
+          const data = await response.json();
+          const files = data.files;
+          if (downloadEnvelopes) await downloadAllFileUrls(files);
+          return files;
```
bo-dun-1, line 1027: let's return this in a format like { files } so in the future if we want to add other parameters to return it'll be easy to do it in a backwards compatible way

PR #1447, opened 2025-08-20
feat: Add code action for updating theme

trebitowski:
## Changes
Allows users to change the theme from inside a logic rule. I added a parameter to remountAllForms which will save the current step for each form. Then on form loading we set the initial step to the saved step if it exists. This allows us to remount the form without losing the form position.

- [ ] Tested end to end


src/Form/index.tsx
```diff
@@ -1192,19 +1193,24 @@ function Form({
         if (authState.redirectAfterLogin || authState.hasRedirected || stepKey)
           return;
 
-        const newKey = getInitialStep({
-          initialStepId,
-          steps,
-          sessionCurrentStep: session.current_step_key
-        });
+        // Check if there's a saved step key from remounting (for theme switching)
+        const savedStepKey = getSavedStepKey(_internalId);
```
bo-dun-1, line 1197: should we put getSavedStepKey logic in getInitialStep?

PR #1449, opened 2025-08-28
fix: Autocomplete Issue

JuneRhoe:
## Changes
- Use `new-password` to disable autocomplete rather than simply using `off`
  - I previously used the same method to disable autocomplete, and there were no other issues related to it.
  - https://developer.mozilla.org/en-US/docs/Web/Security/Practical_implementation_guides/Turning_off_form_autocompletion?utm_source=chatgpt.com
  

https://github.com/user-attachments/assets/df86a512-1d0e-449e-aedb-cf837ceecfef

- [ ] Tested end to end

## Related pull requests


src/Form/index.tsx
trebitowski, line 1: Can we revert the change here too. Form elements only accept `on` or `off` for autocomplete

trebitowski (approved):

src/Form/index.tsx
JuneRhoe: Reverted!

PR #1463, opened 2025-09-17
feat: Compact Display on Multi Select Dropdown

JuneRhoe:
## Changes
- Add a compact display feature to the multi select dropdown

https://github.com/user-attachments/assets/686fd536-b6da-4146-b5ac-2f74a4376c06

- [x] Tested end to end

## Related pull requests


src/elements/fields/DropdownMultiField.tsx
```diff
@@ -67,6 +73,115 @@ const TooltipOption = ({ children, ...props }: OptionProps<OptionData>) => {
...
+
+      const containerWidth = el.clientWidth;
+
+      const optionItems = Array.from(
+        el.querySelectorAll('.react-select__multi-value')
+      ) as HTMLElement[];
+
+      let itemTotalWidth = 0;
+
+      for (let i = 0; i < optionItems.length; i++) {
```
trebitowski, line 112: I think the math here has an issue. I see the +more item sooner than i should be. 
image

src/elements/fields/DropdownMultiField.tsx
```diff
@@ -67,6 +73,115 @@ const TooltipOption = ({ children, ...props }: OptionProps<OptionData>) => {
...
+  const [moreCount, setMoreCount] = useState(0);
+
+  const valueContainerRef = useRef<HTMLDivElement | null>(null);
+  const moreIndicatorRef = useRef<HTMLDivElement | null>(null);
+
+  const totalItems = props.getValue().length;
+
+  // Calculate the total width of all the option items
+  // in order to show the +N more indicator
+  useEffect(() => {
```
trebitowski, line 97: I see some flashes too when selecting items. I think it's the combination of useEffect and useState causing any update to require two rerenders. On the first render the item is displayed and overflows, and on the second render it gets added to moreCount and is hidden. 

I'm wondering if we can fix the flashing, maybe with layout effect or useMemo?

JuneRhoe:
@trebitowski I've addressed all your comments!

PR #1472, opened 2025-10-07
feat: Run form server-side logic rule

arykhoda:

- [ ] Tested end to end

## Related pull requests

### DIFF
--- a/src/Form/index.tsx
+++ b/src/Form/index.tsx
@@ -14,7 +14,6 @@ import { calculateGlobalCSS, calculateStepCSS } from '../utils/hydration';
 import {
   clearBrowserErrors,
   getAllElements,
-  httpHelpers,
   isElementInViewport,
   lookUpTrigger,
   mapFormSettingsResponse,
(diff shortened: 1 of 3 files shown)


src/Form/index.tsx
```diff
@@ -769,6 +771,24 @@ function Form({
...
+                setFieldValues(response?.field_data, true, true);
+              } else if (response?.error) {
+                handleRuleError(response?.error, logicRule);
+              }
+            } catch (e: any) {
+              const errorMessage =
+                e.reason?.message ?? e.error?.message ?? e.message;
+              handleRuleError(errorMessage, logicRule);
+            }
+            continue;
```
bo-dun-1, line 789: i think this would read clearer as an if-else rather than a continue in the if statement

src/utils/featheryClient/index.ts
```diff
@@ -802,6 +802,34 @@ export default class FeatheryClient extends IntegrationClient {
...
+      method: 'POST',
+      body: JSON.stringify(data),
+      keepalive: false
+    };
+
+    const run = () =>
+      this._fetch(url, options).then((response) =>
+        response ? response.json() : Promise.resolve()
+      );
+    return this.offlineRequestHandler.runOrSaveRequest(
```
bo-dun-1, line 825: let's not pass this through the offline request handler. let's just call _fetch directly

PR #1613, opened 2026-04-28
fix: Mobile datepicker doesn't show browser errors since input was readonly

trebitowski:
## Problem
On mobile, the date picker was not displaying validation errors when using the browser error reporting (e.g. "Please fill out this field" for required fields).

**Root cause:** To suppress the virtual keyboard on mobile focus, the onFocus handler was setting `e.target.readOnly = true` directly on the DOM element. Browsers exclude readOnly inputs from constraint validation, so native errors are silently skipped.

## Solution
Remove the readOnly workaround and replace it with `inputMode="none"` on the custom input element. This is the correct approach to tell the browser not to open the virtual keyboard.


tommy-tran-01:
Looks good to me, tested both before & after on my phone's browser 👍

tommy-tran-01 (approved):

PR #1616, opened 2026-05-05
feat: Better phone validation and parsing

trebitowski:
## Summary                                                                                                 
                                                                                                             
  Improves how the PhoneField resolves and displays externally-loaded phone numbers (form load, logic rules, 
  prefilled state) and pasted values, replacing brittle inline parsing in the component with a shared        
  resolver in `validation.ts`.                                                                               
                               
  ## What changed                                
                                                 
  **`PhoneField/validation.ts`** — new `resolvePhoneNumber(LPN, fullNumber, defaultCountry, browserCountry?, 
  selectedCountry?)` helper. Resolution order:
                                                                                                             
  1. Strict-valid international parse, when the input has country code already.
  2. Tries to match with field's default country and user's local country.
  3. Validates against all countries.                       
                                 
  **`PhoneField/tests/PhoneField.spec.tsx`**                                                                 
                           
  - Removed the synthetic regex-based mock for `validation.phoneLib`. Tests now run against real             
  `libphonenumber-js` via the production lazy-load path: `beforeAll` calls `validation.loadPhoneValidator()`
  and awaits `validation.phoneLibPromise`. Production lazy-load semantics are preserved.                                                                                     
  - Added smoke test suite covering different valid inputs and tests them on expected country detection.

### DIFF
--- a/src/elements/fields/PhoneField/index.tsx
+++ b/src/elements/fields/PhoneField/index.tsx
@@ -367,7 +367,6 @@ function PhoneField({
               const LPN = validation.phoneLib;
               if (!LPN) return;
 
-              e.preventDefault();
               const resolved = resolvePhoneNumber(
                 LPN,
                 pasted,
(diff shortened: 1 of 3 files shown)


src/elements/fields/PhoneField/index.tsx
```diff
@@ -350,6 +361,26 @@ function PhoneField({
                 onEnter(e);
               } else if (e.key === '+') setShow(true);
             }}
+            onPaste={(e) => {
+              const pasted = e.clipboardData.getData('text');
```
tommy-tran-01, line 365: Not sure what is the correct behaviour here but do we want to allow users to paste partial numbers in or only full numbers? (insert vs replace)

src/elements/fields/PhoneField/index.tsx
```diff
@@ -350,6 +361,26 @@ function PhoneField({
                 onEnter(e);
               } else if (e.key === '+') setShow(true);
             }}
+            onPaste={(e) => {
```
tommy-tran-01, line 364: Pasting with disable_other_countries=True might require a guard using since pasting a full number will allow users to change the country code through pasting

joondotcode:
Tested user interaction

https://github.com/user-attachments/assets/1dd8d69a-c492-488c-adfa-d80094657f52

tommy-tran-01 (approved):
Looks good to me, not sure how much we want to restrict paste behaviour, it seems tricky with all the possible numbers

faizalsomani (commented):
Ran `resolvePhoneNumber` against the real libphonenumber build to check the resolution order and idempotency. Convergence is fine - I fed each result back through the function the way the `[fullNumber]` effect does and every sample settled in one pass, no oscillation. One gap below in the final safety check.

PR #1635, opened 2026-06-01
fix: Repeatable field support in assistant tools

lanthony42:
## Changes

- Added repeatable field and actions support
- Improved change value behaviour

- [ ] Tested end to end

## Related pull requests

- https://github.com/feathery-org/ai-services/pull/469

### DIFF
--- a/src/Form/grid/Element/index.tsx
+++ b/src/Form/grid/Element/index.tsx
@@ -11,13 +11,9 @@ import {
   numMatchingItems,
   stringifyWithNull
 } from '../../../utils/primitives';
-import { isFieldValueEmpty } from '../../../utils/validation';
 import { justRemove } from '../../../utils/array';
 import { fieldValues, initState } from '../../../utils/init';
-import {
-  ACTION_STORE_FIELD,
-  NAVIGATION_ACTIONS
-} from '../../../utils/elementActions';
+import { isButtonDisabled } from '../../../utils/button';
 import {
   getInlineError,
   handleCheckboxGroupChange,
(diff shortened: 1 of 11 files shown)


varygx (approved):
# PR 1635 review comments

## Summary comment

No blockers — the core repeat plumbing traced out correctly (`field.repeat` / `element.repeat` consumption, phantom row handling, `getServarRepeatNum` extraction is behavior-identical, prior values, ai-services schema match). Two suggestions worth addressing before merge, one optional nit.

---

## 1. `src/assistant/tools/utils.ts:7` (`getContainerRowCount`)

Row counts here (and `maxRowCountByContainer` in panelRuntime.ts) only count repeated servar fields, but the renderer computes rows as `Math.max(repeatCountByFields, repeatCountByTextVariables, 1)` (hideAndRepeats.ts:166). So the assistant can disagree with what's on screen:

- A repeated container driven only by a text variable (repeated text + icon button over a hidden field array) renders N rows, but the assistant computes `rowCount 0` — every per-row click fails with "(none yet - add a row first)".
- Mixed case: repeated servar with 2 rows + text variable bound to a 5-element array renders 5 rows, but indices 2-4 are rejected as out of range.

`state.visiblePositions[positionKey]` already holds one flag per rendered row (panelRuntime already reads it for visibility), so `visiblePositions[getPositionKey(element)].length` would match the renderer by construction and could replace this helper and the `maxRowCountByContainer` loop.

---

## 2. `src/assistant/tools/setFieldValue.ts:380` (apply loop try/catch)

`changeValue` mutates form state immediately, so if entry 3 of 5 throws here, entries 1-2 have already been applied to the form — but the catch reports all 5 as `apply_failed`, so the LLM thinks nothing was set and may re-apply or mislead the user. Suggest a per-entry try/catch so already-applied entries still report `ok: true`.

---

## 3. `src/assistant/AssistantChat.tsx:455` (nit)

`clickElement`'s `repeatIndex` is parsed here (missing → `null`) while `setFieldValue`'s parsing lives inside `dispatchSetFieldValue` (missing → `undefined`). Both are handled fine by `validateRepeatIndex`, but moving this into `dispatchClickElement` would match the setFieldValue structure.

PR #1648, opened 2026-06-08
feat: Update Segment integration

trebitowski:
## Summary
Migrates the Segment integration from the hand-vendored analytics.js v1 snippet to the official `@segment/analytics-next` (analytics.js 2.0) npm package, loaded dynamically in its own chunk. Motivated by reports of missing events.

### Why this is more reliable
- **Delivery on unload/redirect:** analytics-next sends cloud-mode events via `navigator.sendBeacon`/`keepalive`, which survive the completion-redirect (`Form/index.tsx`) that was cancelling the legacy v1 XHRs.
- **No pre-load event loss:** a synchronous buffering facade on `window.analytics` queues `track/page/identify` and replays them once the chunk loads.

### Failure handling added
- Chunk-load failure → caught and logged; integration no-ops gracefully.
- SDK init failure → surfaced via `.load().catch()` instead of buffering forever.

### Changes
- `@segment/analytics-next@1.84.0` added.
- `src/integrations/segment.ts` rewritten: lazy `import()` into its own chunk, buffered facade, error handling, `page()` + optional `identify()` parity, guard against clobbering a customer's existing Segment instance.
- `src/types/global.d.ts`: `analytics: Array` → `analytics: any`.
- `rollup.config.mjs`: `browser: true` on node-resolve so the esm/cjs builds resolve Segment's browser `fetch` path (excludes `node-fetch`), matching what the webpack UMD build already does.
- No changes to event-fire site (`utils.ts`) or `Form/index.tsx`.

## Test plan
- [x] Webpack UMD build emits a lazy named `segment.<hash>.js` chunk (~104K); `node-fetch` excluded from main entry.
- [x] Rollup esm/cjs build splits analytics-next into its own lazy chunk; `node-fetch` excluded.
- [x] Lint passes on changed files.
- [ ] Verify events land in a Segment source (page, step load, form complete) against a test form.
- [ ] Verify completion-redirect no longer drops the `FeatheryFormComplete` event.

## Testing Results
**Prod Form (old Segment integration)**
5 events
| Events | Page: /to/zc6smq | Identify: cd753... | Track: FeatheryStepLoad | Track: FeatheryStepSubmit | Track: FeatheryFormComplete |
|--------|--------|--------|--------|--------|--------|
| image | image | image | image | image | image |

**Test Form (new Segment integration)**
5 events
| Events | Page: /to/zc6smq | Identify: cd753... | Track: FeatheryStepLoad | Track: FeatheryStepSubmit | Track: FeatheryFormComplete |
|--------|--------|--------|--------|--------|--------|
| image | image | image | image | image | image |

**Segment chunk loading dynamically (after form load)**
image
* segment.js is fetched from localhost
* queued events are played to segment: p (page), i (identify), t (track)

### DIFF
--- a/rollup.config.mjs
+++ b/rollup.config.mjs
@@ -6,6 +6,7 @@ import copy from 'rollup-plugin-copy';
 import replace from '@rollup/plugin-replace';
 import babel from '@rollup/plugin-babel';
 import { readFileSync } from 'fs';
+import path from 'path';
 
 // Read package.json
 const pkg = JSON.parse(readFileSync('./package.json', 'utf-8'));
@@ -57,8 +58,27 @@ export default {
         __PACKAGE_VERSION__: JSON.stringify(pkg.version)
       }
     }),
+    // @segment/analytics-next's `browser` field swaps its deprecated Node
+    // entry (which drags in node-fetch and breaks the build) for a browser
+    // stub. node-resolve only honors that field via the global `browser: true`
+    // option, which would also flip every other dependency to its browser
+    // variant (e.g. Stripe and react-datepicker to their UMD builds), so scope
+    // the swap to just this package.
+    {
+      name: 'segment-browser-entry',
+      resolveId(source, importer) {
+        if (
+          source === './node' &&
+          importer?.includes(`@segment${path.sep}analytics-next`)
+        ) {
+          return path.resolve(
+            'node_modules/@segment/analytics-next/dist/pkg/node/node.browser.js'
+          );
+        }
+        return null;
+      }
+    },
     resolve({
-      browser: true,
       extensions: ['.ts', '.tsx', '.js', '.jsx']
     }),
     commonjs(),
--- a/rollup.config.mjs
+++ b/rollup.config.mjs
@@ -58,6 +58,7 @@ export default {
       }
     }),
     resolve({
+      browser: true,
       extensions: ['.ts', '.tsx', '.js', '.jsx']
     }),
     commonjs(),
--- a/src/integrations/segment.ts
+++ b/src/integrations/segment.ts
@@ -16,10 +16,13 @@ export function installSegment(segmentConfig: any) {
   if (!segmentConfig || segmentInstalled) return;
   segmentInstalled = true;
 
-  // If a real (customer-loaded) Segment instance is already on the page, don't
-  // clobber it — just respect it and optionally identify the user.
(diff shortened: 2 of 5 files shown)


src/integrations/segment.ts
```diff
@@ -1,99 +1,65 @@
...
-  if (!segmentConfig || segmentInstalled) return Promise.resolve();
+// Methods the rest of the SDK calls on `window.analytics` (see trackEvent in
+// ./utils). We stub these synchronously so any event fired before the Segment
+// chunk finishes downloading is queued and replayed, rather than silently
+// dropped by the `if (window.analytics)` guard at the call site.
+const BUFFERED_METHODS = ['track', 'page', 'identify'] as const;
+
+type QueuedCall = [typeof BUFFERED_METHODS[number], any[]];
+
+export async function installSegment(segmentConfig: any) {
```
joondotcode: awaiting the import() here makes `fetchSession` block first render for Segments forms - i think the buffer already covers the gap so we can let it load in the background instead of waiting on it

src/integrations/segment.ts
```diff
@@ -1,99 +1,65 @@
...
   segmentInstalled = true;
 
-  // Script from https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/quickstart/#step-2-add-the-segment-snippet
-  // Create a queue, but don't obliterate an existing one!
-  const analytics = (featheryWindow().analytics =
-    featheryWindow().analytics || []);
+  // If a real (customer-loaded) Segment instance is already on the page, don't
+  // clobber it — just respect it and optionally identify the user.
+  const existing = featheryWindow().analytics;
+  if (existing && existing.initialize) {
```
joondotcode: the old snippet stepped aside if the page already had Segment loading (.invoked), but this only checks .initialize so if a customer's own Segment is mid-load we'd overwrite it and drop their queued events right? is coexisting with a host-page Sgement something we still want to support, or are we intentionally dropping that here?

rollup.config.mjs
```diff
@@ -58,6 +58,7 @@ export default {
       }
     }),
     resolve({
+      browser: true,
```
joondotcode: Was the dist/cjs output tested as well? with linking the built output to an app? Apparently with browser: true, date picker and stripe payment is swapped to their UMD builds so worth a check on that as well to see if they both mount and function properly too

joondotcode (approved):

PR #1649, opened 2026-06-08
feat: Assistant table actions

lanthony42:
## Changes

- Added assistant tools for adding rows, deleting rows, editing values, and triggering table actions

- [ ] Tested end to end

## Related pull requests

- https://github.com/feathery-org/ai-services/pull/477

### DIFF
--- a/src/assistant/tools/tableMutations.ts
+++ b/src/assistant/tools/tableMutations.ts
@@ -1,0 +1,273 @@
+import {
+  findTableOnCurrentStep,
+  getTableCapabilities,
+  type TableLookupErrorType
+} from './utils';
+
+type TableMutationErrorType =
+  | TableLookupErrorType
+  | 'not_allowed'
+  | 'row_out_of_range'
+  | 'unknown_field'
+  | 'not_mounted'
+  | 'dispatch_failed';
+
+type TableMutationFailure = {
+  ok: false;
+  errorType: TableMutationErrorType;
+  error: string;
+};
+
+type TableMutationResult = { ok: true } | TableMutationFailure;
+
+type CellResult =
+  | {
+      rowIndex: number;
+      fieldKey: string;
+      ok: true;
+      value: unknown;
+      priorValue: unknown;
+    }
+  | {
+      rowIndex: number;
+      fieldKey: string;
+      ok: false;
+      errorType: TableMutationErrorType;
+      error: string;
+    };
+
+type CellBatchResult = { results: CellResult[] };
+
+const validateRowIndex = (
+  rowIndex: number,
+  rowCount: number
+): TableMutationFailure | null => {
+  if (!Number.isInteger(rowIndex) || rowIndex < 0) {
+    return {
+      ok: false,
+      errorType: 'shape_mismatch',
+      error: 'rowIndex must be a non-negative integer.'
+    };
+  }
+  if (rowIndex >= rowCount) {
+    return {
+      ok: false,
+      errorType: 'row_out_of_range',
+      error: `Row ${rowIndex} is out of range (table has ${rowCount} row${
+        rowCount === 1 ? '' : 's'
+      }).`
+    };
+  }
+  return null;
+};
+
+type Capability = 'add' | 'del' | 'edit';
+
+const CAPABILITY_LABELS: Record<Capability, string> = {
+  add: 'adding rows',
+  del: 'deleting rows',
+  edit: 'cell editing'
+};
+
+const requireCapability = (
+  table: any,
+  tableId: string,
+  capability: Capability,
+  rowCount: number
+): TableMutationFailure | null => {
+  const caps = getTableCapabilities(table, rowCount);
+  const allowed =
+    (capability === 'add' && caps.canAddRows) ||
+    (capability === 'del' && caps.canDeleteRows) ||
+    (capability === 'edit' && caps.canEditCells);
+  if (allowed) return null;
+  return {
+    ok: false,
+    errorType: 'not_allowed',
+    error: `Table '${tableId}' does not allow ${CAPABILITY_LABELS[capability]}.`
+  };
+};
+
+export async function dispatchAddTableRow(
+  formUuid: string | undefined,
+  tableId: string
+): Promise<TableMutationResult> {
+  const lookup = findTableOnCurrentStep(formUuid, tableId);
+  if (!lookup.ok) return lookup;
+  const { state, table, rowCount } = lookup.found;
+
+  const cap = requireCapability(table, tableId, 'add', rowCount);
+  if (cap) return cap;
+
+  try {
+    if (!state.assistantClient.addTableRow(tableId)) {
+      return {
+        ok: false,
+        errorType: 'not_mounted',
+        error: `Table '${tableId}' is not mounted yet.`
+      };
+    }
+  } catch (err) {
+    return {
+      ok: false,
+      errorType: 'dispatch_failed',
+      error: err instanceof Error ? err.message : String(err)
+    };
+  }
+  return { ok: true };
+}
+
+export async function dispatchDeleteTableRow(
+  formUuid: string | undefined,
+  tableId: string,
+  rowIndex: number
+): Promise<TableMutationResult> {
+  const lookup = findTableOnCurrentStep(formUuid, tableId);
+  if (!lookup.ok) return lookup;
+  const { state, table, rowCount } = lookup.found;
+
+  const cap = requireCapability(table, tableId, 'del', rowCount);
+  if (cap) return cap;
+  const rowErr = validateRowIndex(rowIndex, rowCount);
+  if (rowErr) return rowErr;
+
+  try {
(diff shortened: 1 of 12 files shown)


src/assistant/tools/tableMutations.ts
```diff
@@ -0,0 +1,216 @@
...
+        rowIndex,
+        fieldKey,
+        ok: false,
+        errorType: rowErr.errorType,
+        error: rowErr.error
+      });
+      continue;
+    }
+    if (
+      !state.assistantClient.editTableCell(tableId, rowIndex, fieldKey, value)
```
varygx: **Cell values: no validation + immediate persist.** `editTableCell` → `handleCellEdit` writes the raw LLM value into the column array and `submitCustom` persists it to the server immediately (useTableMutations.ts:116-119). Two gaps vs the rest of the system:

- The UI only ever produces strings here — `EditableCell` is a plain textarea for every column type, and clear writes `''`. The ai-services tool description tells the LLM to send `number` for numeric, `boolean` for checkbox, and `null` to clear — so assistant edits introduce value types into these arrays that the UI never produces, and they're saved server-side instantly.
- `setFieldValue` shape-validates via `checkValueAgainstField` before writing; this tool has no equivalent.

Suggest either stringifying non-null values + mapping `null` → `''` (match the UI convention), or validating against the column's `field_type`. Either way the `setTableCellValue` description in feathery-org/ai-services#477 should be updated to match.

src/assistant/tools/tableUtils.ts
```diff
@@ -0,0 +1,102 @@
...
+    return Array.isArray(v) ? Math.max(max, v.length) : max;
+  }, 0);
+  return { ok: true, found: { state, table, columns, rowCount } };
+};
+
+export const getTableCapabilities = (
+  table: any
+): { canEditCells: boolean; canAddRows: boolean; canDeleteRows: boolean } => {
+  const props = table?.properties ?? {};
+  const canEditCells = !!props.enable_editing && !props.transpose;
```
varygx: **Transpose gating drifts from the UI.** `canEditCells = !!props.enable_editing && !props.transpose`, but the table component computes `isTransposed = enableTranspose && baseNumRows > 0` (useTableData.ts:489). For a transpose-configured table with zero rows and editing enabled, the UI un-transposes and shows a working "+ Add Row" button while the snapshot omits `canAddRows` and `addTableRow` refuses with `not_allowed`. Computing capabilities with the row count (`!(transpose && rowCount > 0)`) keeps the assistant in agreement with what's rendered.

src/assistant/tools/triggerTableAction.ts
```diff
@@ -0,0 +1,114 @@
...
+      error: `Table '${tableId}' is not mounted yet.`
+    };
+  }
+
+  const rowData = buildRowData(lookup.found, rowIndex);
+  const fromStepKey = state.currentStep.key;
+  const errorsBefore = snapshotInlineErrors(state);
+
+  try {
+    await state.assistantClient.runTableAction({
```
varygx, line 88: **`ok: true` for tables with nothing wired.** The dispatch never checks the table has an Element-Click logic rule — `runTableAction` fires `runElementActions` with `actions: []`, so a rule-less table no-ops and returns `{ok: true, navigated: null}`. The prompt tells the LLM to require `hasLogicRules: true`, but the mutation tools enforce their flags at the boundary (`not_allowed`) while this one trusts the prompt. `elementHasLogicRules` already computes exactly this for the snapshot (panelRuntime.ts) — reusing it here and returning a typed error keeps the LLM from reporting success on a no-op.

src/assistant/tools/triggerTableAction.ts
```diff
@@ -0,0 +1,114 @@
...
+    return {
+      ok: false,
+      errorType: 'row_out_of_range',
+      error: `Row ${rowIndex} is out of range (table has ${rowCount} row${
+        rowCount === 1 ? '' : 's'
+      }).`
+    };
+  }
+
+  if (!state.assistantClient) {
```
varygx: nit: this guard is for a missing `state.assistantClient`, which clickElement.ts:79 classifies as `no_form_state` — and this tool never touches the table registry, so `not_mounted` (documented in ai-services as "the table's registry entry hasn't been published yet") is misleading, as is the "Table 'X' is not mounted yet" message. Folding the client check into `findTableOnCurrentStep`'s `no_form_state` would align it.

src/assistant/tools/tableMutations.ts
```diff
@@ -0,0 +1,216 @@
...
+  });
+
+  const lookup = findTableOnCurrentStep(formUuid, tableId);
+  if (!lookup.ok) return fanOut(lookup.errorType, lookup.error);
+  const { state, table, columns, rowCount } = lookup.found;
+
+  const cap = requireCapability(table, tableId, 'edit');
+  if (cap) return fanOut(cap.errorType, cap.error);
+
+  if (!state.assistantClient?.hasTable(tableId)) {
```
varygx: nit: the `hasTable` pre-check is redundant — every cell's `editTableCell` call already returns `false` → `not_mounted` per cell when the registry entry is missing. Dropping the pre-check loses nothing.

src/elements/basic/TableElement/index.tsx
```diff
@@ -100,6 +101,18 @@ function TableElement({
       onMutate
     });
 
+  const tableId = element?.id;
+  // Lets the assistant invoke this table's mutations through the same handlers the user UI calls
+  useEffect(() => {
+    if (!assistantClient || !tableId) return;
+    assistantClient.registerTable(tableId, {
```
varygx: nit: the registry gets the unwrapped `handleCellEdit`, but the UI edit path goes through `wrappedHandleCellEdit`, which clears the `pendingAddRows` bookkeeping. Today that set is write-only so there's no functional difference, but if pending-row state ever drives behavior, assistant edits will silently diverge from user edits. Registering the wrapped version keeps exact parity.

src/assistant/tools/tableUtils.ts
```diff
@@ -0,0 +1,102 @@
...
+export const buildRowData = (
+  found: FoundTable,
+  rowIndex: number
+): Record<string, any> => {
+  const rowData: Record<string, any> = {};
+  for (const col of found.columns) {
+    if (!col?.field_key) continue;
+    const v = found.state.fields?.[col.field_key]?.value;
+    const cValue = Array.isArray(v) ? v[rowIndex] : v;
+    rowData[col.name ?? col.field_key] = cValue;
```
varygx: nit: `col.name ?? col.field_key` — in what situation is `name` missing? The `Column` type requires it and the UI's own rowData construction (Actions.tsx:99) indexes `col.name` directly, so the fallback is dead.

varygx (approved):

src/assistant/tools/tableMutations.ts
lanthony42: Coerced to what typing produces: `null` → `''`, everything else → `String(value)`, so assistant edits land identical to user edits. Tool description + prompt updated to match - format guidance is now grounded in the column's existing values (table columns carry no data types).

src/assistant/tools/tableUtils.ts
lanthony42: `getTableCapabilities` now takes `rowCount` and only blocks editing when `transpose && rowCount > 0`, matching `isTransposed`. Snapshot and dispatch share the same call so the flag and the refusal can't disagree.

src/assistant/tools/triggerTableAction.ts
lanthony42: Folded the client check into `findTableOnCurrentStep` as `no_form_state` and dropped this block - `not_mounted` now only fires from registry misses.

src/assistant/tools/tableMutations.ts
lanthony42: Dropped the pre-check - per-cell `editTableCell` misses report `not_mounted` per entry.

src/elements/basic/TableElement/index.tsx
lanthony42: Registry now gets wrapped handlers for all three - added `wrappedHandleAddRow`/`wrappedHandleDeleteRow` too. Beyond the pending-row bookkeeping, the wraps dismiss an open delete-confirm, which otherwise targets a shifted index after an assistant add/delete. UI call sites use the same wrapped trio.

src/assistant/tools/tableUtils.ts
lanthony42: Dropped - `FoundTable.columns` now types `name` as required and rowData indexes `col.name` directly.

src/assistant/tools/triggerTableAction.ts
lanthony42: Keeping prompt + snapshot as the defense here, matching clickElement: a text/container with no actions and no rules also dispatches an empty action chain and returns `ok: true`. The curation lives in the snapshot (tables expose `hasLogicRules`) and the prompt requires the flag, same as the clickable-element pattern. Enforcing at this boundary would make the table tool stricter than the click path.

PR #1655, opened 2026-06-10
feat: Assistant navigate tool for stepper navigation

lanthony42:
## Changes

- Added navigate tool for stepper navigation
- Improved navigation reporting for all actions

- [ ] Tested end to end

## Related pull requests

- https://github.com/feathery-org/ai-services/pull/480

### DIFF
--- a/src/assistant/tools/navigate.ts
+++ b/src/assistant/tools/navigate.ts
@@ -40,7 +40,7 @@ export async function dispatchNavigate(
       error: 'Form has not loaded yet.'
     };
   }
-  if (typeof stepKey !== 'string' || stepKey.length === 0) {
+  if (stepKey.length === 0) {
     return {
       ok: false,
       errorType: 'shape_mismatch',
@@ -60,7 +60,7 @@ export async function dispatchNavigate(
   }
 
   const client = state.assistantClient;
-  const fromStepKey = getLiveStepKey(state);
+  const fromStepKey = getLiveStepKey(state) ?? '';
   const errorsBefore = snapshotInlineErrors(state);
 
   try {
--- a/src/assistant/tools/navigate.ts
+++ b/src/assistant/tools/navigate.ts
@@ -1,6 +1,6 @@
 import internalState from '../../utils/internalState';
 import { ACTION_NEXT } from '../../utils/elementActions';
-import { collectStepperNavigation } from './panelRuntime';
+import { collectNavigableSteps } from './panelRuntime';
 import { getLiveStepKey, snapshotInlineErrors } from './utils';
 
 type NavigateErrorType =
@@ -48,7 +48,7 @@ export async function dispatchNavigate(
     };
   }
 
-  const target = collectStepperNavigation(state).find(
+  const target = collectNavigableSteps(state).find(
     (n) => n.stepKey === stepKey
   );
   if (!target) {
@@ -67,7 +67,11 @@ export async function dispatchNavigate(
     await client.runActions({
       actions: [{ type: ACTION_NEXT, next_step_key: stepKey }],
       element: target.element,
-      elementType: 'progress_bar'
+      elementType: target.elementType,
+      // A tab carries its own submit behavior, mirror the live tab click
+      ...(target.elementType === 'tab'
+        ? { submit: !!target.element.properties?.submit }
+        : {})
     });
   } catch (err) {
     return {
--- a/src/assistant/tools/navigate.ts
+++ b/src/assistant/tools/navigate.ts
@@ -1,0 +1,94 @@
+import internalState from '../../utils/internalState';
+import { ACTION_NEXT } from '../../utils/elementActions';
+import { collectStepperNavigation } from './panelRuntime';
+import { getLiveStepKey, snapshotInlineErrors } from './utils';
+
+type NavigateErrorType =
+  | 'step_not_allowed'
+  | 'no_form_state'
+  | 'shape_mismatch'
+  | 'dispatch_failed';
+
+type NavigateResult =
+  | {
+      ok: true;
+      navigated: { fromStepKey: string; toStepKey: string } | null;
+      fieldErrors?: Record<string, string>;
+    }
+  | {
+      ok: false;
+      errorType: NavigateErrorType;
+      error: string;
+    };
+
+export async function dispatchNavigate(
+  formUuid: string | undefined,
+  stepKey: string
+): Promise<NavigateResult> {
+  if (!formUuid) {
+    return {
+      ok: false,
+      errorType: 'no_form_state',
+      error: 'Form has not loaded yet.'
+    };
+  }
+  const state = internalState[formUuid];
+  if (!state || !state.currentStep || !state.assistantClient) {
+    return {
+      ok: false,
+      errorType: 'no_form_state',
+      error: 'Form has not loaded yet.'
+    };
+  }
+  if (typeof stepKey !== 'string' || stepKey.length === 0) {
+    return {
+      ok: false,
+      errorType: 'shape_mismatch',
+      error: 'stepKey is required.'
+    };
+  }
+
+  const target = collectStepperNavigation(state).find(
+    (n) => n.stepKey === stepKey
+  );
+  if (!target) {
+    return {
+      ok: false,
+      errorType: 'step_not_allowed',
+      error: `Step '${stepKey}' is not reachable through a navigation surface right now.`
+    };
+  }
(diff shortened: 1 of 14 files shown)


src/assistant/tools/panelRuntime.ts
```diff
@@ -131,6 +144,124 @@ const resolveText = (text: string, repeat?: number): string => {
 export const getCurrentStepKey = (formId: string): string | undefined =>
   internalState[formId]?.currentStep?.key;
 
+const isElementVisible = (state: any, el: any): boolean => {
+  const positionKey = el.position?.join(',') || 'root';
+  const flags = (state.visiblePositions ?? {})[positionKey];
+  return Array.isArray(flags) ? flags.some(Boolean) : true;
+};
+
+export const collectNavigableSteps = (
```
varygx, line 169: **Reachability is computed twice and will drift apart.** `collectNavigableSteps` (the dispatch gate) and `collectNavigationSurfaces` (the snapshot the model sees) each walk `progress_bars` + `tabs` and independently re-derive the same rule — `!active && (allowAll || completedStepKeys.has(stepKey))` for steppers, unconditional for tabs. They agree today, but they're the two halves of one contract: change the rule in one and not the other and the model is shown a step as `reachable: true` that `navigateToStep` then refuses with `step_not_allowed` (or the reverse — a reachable step it's never told about). Flattening the gate's navigable entries out of the same walk that builds the surfaces (carrying `element`/`elementType` on each reachable entry) keeps reachability defined in exactly one place.

src/assistant/tools/panelRuntime.ts
```diff
@@ -131,6 +144,124 @@ const resolveText = (text: string, repeat?: number): string => {
...
+};
+
+export const collectNavigationSurfaces = (
+  state: any
+): PanelRuntimeNavigationSurface[] => {
+  const step = state?.currentStep;
+  if (!step) return [];
+  const out: PanelRuntimeNavigationSurface[] = [];
+  const seenSurfaceIds = new Set<string>();
+  const completedStepKeys = getCompletedStepKeys();
```
varygx, line 224: **Stepper reachability reads a set the snapshot never hydrates.** `getCompletedStepKeys()` returns `initState.completedSteps`, which for prior-session completions is populated only as a side-effect of the stepper mounting — `ProgressBarElement` fires `loadCompletedSteps(client)` in a `useEffect` (`ProgressBarElement/index.tsx:50`) and that fetch is async. `getPanelRuntimeSnapshot` never triggers it. So on the first assistant turn after load, before `fetchCompletedSteps()` resolves, a non-`navigate_to_all_steps` stepper reports prior-session-completed steps as `reachable: false` and `navigateToStep` refuses them with `step_not_allowed` — the model tells the user "that step isn't reachable yet" when it is. It self-heals next turn, but the snapshot shouldn't depend on an unrelated element's render effect for correctness. Triggering `loadCompletedSteps(state.client)` from the snapshot path makes the first turn correct. (Tabs and `navigate_to_all_steps` steppers are unaffected — neither gates on completion.)

src/assistant/tools/navigate.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  const fieldErrors: Record<string, string> = {};
+  for (const key of Object.keys(errorsAfter)) {
+    if (errorsAfter[key] !== errorsBefore[key])
+      fieldErrors[key] = errorsAfter[key];
+  }
+
+  return {
+    ok: true,
+    navigated: toStepKey !== fromStepKey ? { fromStepKey, toStepKey } : null,
```
varygx, line 95: **`navigated: null` can't tell a blocked jump from a jump that never ran.** `dispatchNavigate` ignores `runActions`' return value and infers the outcome by re-reading the step key. But `runElementActions` early-returns without acting in two cases — a rapid re-click on the same element (`Form/index.tsx:2147`) and another button action already running (`Form/index.tsx:2152`). When either fires, no navigation and no field errors occur, so this returns `{ ok: true, navigated: null }` — byte-identical to the "validation blocked the jump" shape. The result-reading guidance in feathery-org/ai-services#480 tells the model `navigated: null` with no `fieldErrors` means validation blocked it, so on a no-op the model reports a block that never happened or loops. `clickElement` shares this shape, but `navigateToStep` is dispatched programmatically with no human gesture between calls, so the `isButtonActionRunning()` collision is far likelier here. Worth a distinct signal when `runActions` returns without effect, and documenting the no-op state in feathery-org/ai-services#480's "Reading the result".

src/assistant/tools/utils.ts
```diff
@@ -1,6 +1,9 @@
 import internalState from '../../utils/internalState';
 import { getPositionKey } from '../../utils/hideAndRepeats';
 
+export const getLiveStepKey = (state: any): string =>
```
varygx: **`getLiveStepKey`'s return type hides that it can be undefined.** The body is `state.latestStepName ?? state.currentStep?.key` — both operands are `string | undefined`, so the real return type is `string | undefined`, not `string`. The annotation matters because callers split on it: `navigate.ts:84`, `clickElement.ts:208`, and `triggerTableAction.ts:103` all write `getLiveStepKey(state) ?? fromStepKey`, whose `?? fromStepKey` is dead code under the declared `string` type but live at runtime — while `navigate.ts:63` reads it with no fallback. Annotating `string | undefined` makes the fallbacks honest and the asymmetry deliberate rather than accidental.

src/assistant/tools/navigate.ts
```diff
@@ -0,0 +1,98 @@
...
+  }
+  const state = internalState[formUuid];
+  if (!state || !state.currentStep || !state.assistantClient) {
+    return {
+      ok: false,
+      errorType: 'no_form_state',
+      error: 'Form has not loaded yet.'
+    };
+  }
+  if (typeof stepKey !== 'string' || stepKey.length === 0) {
```
varygx: nit: `typeof stepKey !== 'string'` can't be true here — `AssistantChat.tsx:60` already coerces the tool input to a string (`typeof input.stepKey === 'string' ? input.stepKey : ''`) before calling, and the empty-string case is already caught by the `collectNavigableSteps` lookup below returning not-found → `step_not_allowed`. The separate `shape_mismatch` branch guards a state the dispatch contract rules out; `clickElement` carries no equivalent guard on its `elementId`.

src/assistant/tools/panelRuntime.ts
```diff
@@ -131,6 +144,124 @@ const resolveText = (text: string, repeat?: number): string => {
...
+      if (!isStepperStepVisible(s)) return;
+      if (!allowAll && !completedStepKeys.has(stepKey)) return;
+      seen.add(stepKey);
+      out.push({ stepKey, element: el, elementType: 'progress_bar' });
+    });
+  });
+  // Tabs jump freely to any tab's step, no completed-step gating
+  (step.tabs ?? []).forEach((el: any) => {
+    if (!isElementVisible(state, el)) return;
+    (el.properties.entries ?? []).forEach((s: any) => {
```
varygx: nit: the tab loops reach `el.properties.entries` directly while the progress-bar branch just above guards `el?.properties?.stepper`. Rendered elements always carry `.properties`, so neither guard is needed — cleaner to drop the `?.` on the stepper branch for consistency than to read `properties` two ways in the same function.

src/utils/internalState.ts
```diff
@@ -66,7 +66,9 @@ export type UpdateDocusignEnvelopeParams = {
 export interface FormInternalState {
   language: string | undefined;
   currentStep: any;
+  latestStepName?: string;
```
varygx: nit: `latestStepName` is a shadow of the live step key that has to be hand-updated at every transition (four sites in `Form/index.tsx` plus `formContext.goToStep`), and `getLiveStepKey` reads it because `currentStep.key` lags a render behind on the hash-routing path. That's non-obvious — one line on the field noting it's the eagerly-updated step key for the dispatch-await window would keep the next person from deleting it as redundant with `currentStep`.

varygx (approved):

src/assistant/tools/navigate.ts
lanthony42: `runElementActions` return value doesn't clearly indicate no-op result, might indicate async action instead, changed prompting to account for this case instead

src/assistant/tools/navigate.ts
lanthony42: Removed `typeof` check, kept separate return message for `stepKey.length === 0`

PR #1656, opened 2026-06-11
feat: Table classes

trebitowski:
## Changes
* Adds classes to table elements so they can be targeted with custom css.
* Adds `data-feathery-field` attributes to cells, to allow for styling the table by column.
* Overrides emotion's default style cache so the emotion styles will be added to the document above custom css. Allowing for custom css with the same specificity to override the emotion styles.
   * This is a change in behavior, but impact should be low or none. The change is that now custom css that targets a single class `.classname {}`, will take precedence over our default emotion styles. We don't have any useful classes to target prior to this change. And if there is custom css, the user's intention is for that css to get applied, so it should be taking precedence over the default styles.

### Default - no custom css
image

### Style everything
image

### Dark mode, no corner radius
image

### Transparent
image

- [ ] Tested end to end

## Related pull requests

### DIFF
--- a/src/elements/basic/TableElement/Actions.tsx
+++ b/src/elements/basic/TableElement/Actions.tsx
@@ -1,4 +1,4 @@
-import { useState, useRef, useEffect } from 'react';
+import { useState, useRef, useEffect, Fragment } from 'react';
 import { createPortal } from 'react-dom';
 import { fieldValues } from '../../../utils/init';
 import {
@@ -11,6 +11,7 @@ import {
   menuIconStyle,
   actionIconButtonStyle
 } from './styles';
+import { TABLE_CLASS } from './classNames';
 import { Action, Column } from './types';
 import { featheryDoc } from '../../../utils/browser';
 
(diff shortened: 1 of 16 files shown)


src/elements/basic/TableElement/Actions.tsx
```diff
@@ -130,7 +131,7 @@ export function ActionButtons({
   return (
     <div ref={containerRef} css={actionContainerStyle}>
       {useOverflow ? (
-        <>
+        <Fragment>
```
joondotcode, line 134: Is Fragment necessary here? Can't you just do `<>`?

src/elements/basic/TableElement/classNames.ts
```diff
@@ -0,0 +1,32 @@
+// Public, user-targetable class names for the parts of the table element.
+//
+// These are a documented, stable API: form builders write custom CSS against
+// them (e.g. `.feathery-table-cell { ... }`).
+//
+// Treat renames/removals as breaking changes to customer CSS.
+export const TABLE_CLASS = {
+  container: 'feathery-table-container',
```
joondotcode, line 8: Can we alphabetize

joondotcode (approved):

PR #1658, opened 2026-06-12
fix: Enable iframe container auto scrolling

joondotcode:
## Changes

- Keep embedded iframe containers constrained to their configured box and let the iframe viewport scroll automatically when its content is taller.
- Add a focused StyledContainer test covering the iframe scroll behavior without changing generic container overflow behavior.

## Why

Customers can set a height for iframe containers, but setting the Feathery container overflow does not make the iframe content itself scroll. This keeps the fix scoped to native iframe scrolling for arbitrary iframe URLs, without introducing parent-side height measurement or a postMessage resize contract.

### DIFF
--- a/src/Form/grid/StyledContainer/index.spec.tsx
+++ b/src/Form/grid/StyledContainer/index.spec.tsx
@@ -1,0 +1,66 @@
+import { render } from '@testing-library/react';
+import { StyledContainer } from '.';
+
+const baseNode = {
+  id: 'container-1',
+  key: 'container-1',
+  type: 'container',
+  isElement: false,
+  parent: { styles: { height: 'fit', axis: 'column' } },
+  children: [],
+  properties: {},
+  styles: {
+    axis: 'column',
+    content_responsive: false,
+    height: 200,
+    height_unit: 'px',
+    width: 'fill',
+    width_unit: 'fill'
+  }
+};
+
+describe('StyledContainer iframe embeds', () => {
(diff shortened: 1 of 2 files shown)


joondotcode:
Before merge, can you confirm the expected iframe behavior?

There are two possible behaviors here:

1. The builder sets a fixed iframe box, for example 500px tall. If the iframe content is taller than 500px, the user scrolls inside that box. If the iframe content is shorter, the box still stays 500px tall.
2. The iframe box changes height to match the iframe content, for example growing to 900px for long content and shrinking to 300px for short content.

This PR handles the first behavior. It keeps the iframe inside the configured box and lets the iframe scroll internally.

It does not handle automatic height changes. For a third-party URL, we usually cannot read the iframe content height from the parent form because the browser blocks that across domains. To support automatic height, the embedded page would need to send us its height with something like `postMessage`, and we would need a separate resize contract for that.

If the expected behavior is fixed height plus internal iframe scrolling, this looks scoped correctly. If the expected behavior is automatic grow/shrink based on iframe content, that is a larger feature than this PR.

Manual checks I would run:
- [ ] Configure an iframe box with a fixed height, use iframe content taller than that height, and confirm the scroll happens inside the iframe box.
- [ ] Use iframe content shorter than the configured height and confirm the iframe box keeps the configured height without adding a parent-container scrollbar.
- [ ] Check the hosted form or embed path that will be used, not only the isolated component test.

trebitowski (approved):

PR #1659, opened 2026-06-12
feat: Add label styling support

ayeshah:
## Changes
Adds the ability to style field labels independently of the field. 
`applyFontStyles` now takes a prefix so labels read label_font_*/label_color/label_gap from their own style namespace, and `applyFieldStyles` wires up a 'fieldLabel' target.

- [ ] Tested end to end

## Related pull requests

- [Frontend changes
](https://github.com/feathery-org/feathery-frontend/pull/3417)
- [Backend](https://github.com/feathery-org/feathery-backend/pull/3323)

### DIFF
--- a/src/elements/__test__/styles.spec.js
+++ b/src/elements/__test__/styles.spec.js
@@ -1,4 +1,5 @@
 import ResponsiveStyles, { DEFAULT_MOBILE_BREAKPOINT } from '../styles';
+import { LABEL_TEXT_ALIGN_DEFAULT } from '../utils/labelStyleResolver';
 
 const TEST_COLOR_BACKGROUND = 'dddddd';
 const mockElement = {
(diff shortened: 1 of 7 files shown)


src/elements/fields/index.tsx
```diff
@@ -262,6 +269,14 @@ function applyFieldStyles(field: any, styles: any) {
   styles.applyFontStyles('fc', false, true);
   styles.applyFontStyles('field', false, ignoreSelectorColors);
 
+  // Label styling is independent of the field. When a label_* property is
+  // unset, nothing is applied and the label inherits the field's font from its
+  // container ('fc'), preserving prior behavior.
+  applyLabelFontStyles(styles);
```
joondotcode, line 281: Should we add the explicit handling for the `text_align` as well? Seems like it is handled as a customValue in the FE

src/elements/__test__/styles.spec.js
```diff
@@ -62,4 +62,113 @@ describe('responsiveStyles', () => {
...
+    const buildFieldLabelTarget = (styles) => {
+      const rs = new ResponsiveStyles(
+        { styles },
+        [],
+        false,
+        DEFAULT_MOBILE_BREAKPOINT
+      );
+      rs.addTargets('fieldLabel');
+      rs.applyFontStyles('fieldLabel', false, true, 'label_', true);
+      rs.apply('fieldLabel', 'label_margin_top', (a) =>
```
joondotcode: This helper says it mirros applyFieldStyles but label_margin_top is only here in the test, but I don't see this in any of the style key. Is this intended?

ayeshah: Yea this is no longer needed, will be removed. Instead of using margins I am using gap to specify the spacing between the label and field.

trebitowski (approved):

PR #1663, opened 2026-06-15
feat: Support transpose rendering in MatrixField

joondotcode:
When `servar.metadata.transpose` is true, the matrix field now renders questions as column headers and options as rows instead of the default layout. Value storage is unchanged — `fieldVal` remains keyed by question ID so existing submission data is unaffected.

The transposed path uses the same `data-question-id` attribute on each input so the parent's `onChange` handler works without modification.

Resolves Feathery Slack report: https://feathery.slack.com/archives/C09G9MGM4MA/p1781117624191829?thread_ts=1781117624.191829&cid=C09G9MGM4MA

### DIFF
--- a/src/elements/fields/MatrixField/index.tsx
+++ b/src/elements/fields/MatrixField/index.tsx
@@ -20,6 +20,7 @@ function MatrixField({
   const servar = element.servar;
   const allowMultiple = servar.metadata.multiple;
   const inputType = allowMultiple ? 'checkbox' : 'radio';
+  const transpose = servar.metadata.transpose ?? false;
   const containerRef = useRef(null);
   const { backgroundColor, borderRadius } =
     responsiveStyles.getTarget('sub-fc');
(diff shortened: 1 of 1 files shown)


joondotcode:
Before merge, I would check this:

Review findings:
- Non-blocking: Per-question highlight colors (q.highlight_color) are not applied in transposed mode — any matrix field that already uses row highlight coloring will visually lose those colors when Flip Matrix is turned on. The normal path uses backgroundColor per question row; the transposed path has no per-row coloring since options are plain strings. This is a pre-documented limitation (build result follow-up) and is intentional, but the team should confirm existing customers using highlight colors are aware before enabling transpose on their fields.
  File: `src/elements/fields/MatrixField/index.tsx`

Questions:
- In transposed mode, per-question highlight colors are not shown — any matrix row that currently has a highlight color in the normal layout will display without that color when flipped. Before enabling Flip Matrix on a form like Obermeyer's that may already have highlight colors configured, can you confirm whether any active customer matrices use row highlight colors and whether losing them in flipped mode is acceptable?
  Enabling Flip Matrix on a matrix with existing highlight colors silently removes the color styling with no warning in the builder.

Manual checks:
- [ ] Build a matrix with two questions and three options. Enable 'Flip Matrix'. In the live form, select a radio answer for each question row (one option per question), then submit. In the Responses tab, confirm the submission shows two entries keyed by question — not by option — and each value is the option that was selected. Then repeat with 'multiple answers' enabled (checkboxes) and select two options per question; confirm the response shows arrays of option values under each question key.
  Verifies that the transposed onChange wiring correctly maps selections to question IDs for both single and multi-select modes.
- [ ] Create a matrix field with a question that has a highlight color set (via the question's advanced settings). Load the form in normal orientation — confirm the row is highlighted. Enable 'Flip Matrix'. Confirm the form still loads and renders without errors, and note whether the highlight color is present (expected: not visible in transposed mode). This is a known limitation — confirm it does not cause a visual regression such as a broken layout or JavaScript error.
  Confirms the transposed path gracefully omits per-row coloring without breaking rendering.

Source context: https://feathery.slack.com/archives/C09G9MGM4MA/p1781117624191829?thread_ts=1781117624.191829&cid=C09G9MGM4MA
Automation run: https://github.com/feathery-org/feathery-frontend/actions/runs/27575769824

PR #1675, opened 2026-06-25
feat: Assistant voice mode

lanthony42:
## Changes

- Added voice mode: speak to the assistant and hear streamed spoken replies
- Revealed reply text in step with the audio, holding tool pills until their narration plays
- Showed the mic only when the form has voice enabled
- Anchored the chat to the bottom so new content pushes up instead of the view scrolling down

- [ ] Tested end to end

## Related pull requests

- https://github.com/feathery-org/feathery-backend/pull/3391
- https://github.com/feathery-org/ai-services/pull/499
- https://github.com/feathery-org/feathery-frontend/pull/3481

### DIFF
--- a/src/assistant/voice/useAssistantVoice.ts
+++ b/src/assistant/voice/useAssistantVoice.ts
@@ -1,0 +1,232 @@
+import {
+  MutableRefObject,
+  useCallback,
+  useEffect,
+  useRef,
+  useState
+} from 'react';
+import { v4 as uuidv4 } from 'uuid';
+
+import { AudioPlayer } from './audioPlayer';
+import { probeMicAvailable, VoiceMic } from './voiceMic';
+
+const nonWhitespace = (text: string): number => text.replace(/\s/g, '').length;
+
+export type VoiceState =
+  | 'idle'
+  | 'loading'
+  | 'listening'
+  | 'recording'
+  | 'transcribing'
+  | 'thinking'
+  | 'speaking'
+  | 'error';
+
+type ChatStatus = 'submitted' | 'streaming' | 'ready' | 'error';
+
+type VoiceDataPart = {
+  type: string;
+  data?: { text?: string; chunk?: string; done?: boolean };
+};
+
+type UseAssistantVoiceOptions = {
+  status: ChatStatus;
+  sendMessage: (message?: { text: string }) => void;
+  setMessages: (updater: (prev: any[]) => any[]) => void;
+  ensureThread: () => void;
+  voiceActiveRef: MutableRefObject<boolean>;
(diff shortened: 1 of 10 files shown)


src/assistant/voice/voiceMic.ts
```diff
@@ -0,0 +1,90 @@
+import { MicVAD, utils as vadUtils } from '@ricky0123/vad-web';
```
varygx: **The VAD runtime is eagerly bundled into every form load.** This top-level `import { MicVAD } from '@ricky0123/vad-web'` pulls in `onnxruntime-web` (plus protobufjs/flatbuffers/long), and the chain to it is static all the way up — `Form/index.tsx` statically imports `AssistantChat`, which imports `useAssistantVoice`, which imports this module. So the ONNX-runtime JS ships in the core SDK bundle for every form, including forms with the assistant or voice turned off. Loading the voice module lazily — a dynamic `import('./voiceMic')` inside `startVoice` — keeps onnxruntime-web out of the main bundle until a user actually enters voice mode.

src/assistant/voice/useAssistantVoice.ts
```diff
@@ -0,0 +1,225 @@
...
+    micRef.current?.destroy();
+    micRef.current = null;
+    playerRef.current?.skip();
+    playerRef.current = null;
+    pendingAudioRef.current = null;
+    voiceActiveRef.current = false;
+    setVoiceState('idle');
+  }, [voiceActiveRef, pendingAudioRef, setVoiceState]);
+
+  const skipSpeaking = useCallback(() => {
```
varygx, line 158: **"Tap to skip" only silences the current sentence, not the rest of the reply.** `skipSpeaking` calls `AudioPlayer.skip()` → `teardown()`, which clears the queue and resets the player, but there's no turn-level skipped flag — and the player is created once per voice session, not per turn, so `teardown()` just leaves it in a clean, ready state. While the reply is still streaming, later `data-segment` parts keep routing to `pushSegment` (the `voiceDataRef` handler in this file has no skip guard), which re-queues and plays the sentences that arrive after the tap. So tapping skip mid-reply stops the current clip and then the remaining sentences speak anyway (the button even stays "Tap to skip" throughout). A per-turn `skipped` flag that suppresses further `pushChunk`/`pushSegment`, set in `skipSpeaking` and cleared at the start of the next turn in `handleUtterance` (which already does the `skip()` + reset), would make skip mute the whole turn. Note this is playback-only: the server keeps running the agent and generating TTS for the remaining sentences regardless, so stopping that spend too would mean aborting the in-flight request, not just muting the player.

varygx (approved):

PR #1677, opened 2026-06-25
feat: DocuSign envelope options

troy-feathery:

- [ ] Tested end to end

## Related pull requests

### DIFF
--- a/src/utils/featheryClient/integrationClient.ts
+++ b/src/utils/featheryClient/integrationClient.ts
@@ -471,6 +471,7 @@ export default class IntegrationClient {
   sendDocusignEnvelope({
     documents,
     libraryDocuments,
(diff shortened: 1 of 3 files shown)


src/utils/featheryClient/integrationClient.ts
```diff
@@ -488,12 +494,29 @@ export default class IntegrationClient {
         form_key: this.formKey,
         documents,
         library_documents: libraryDocuments,
+        document_instances: documentInstances?.map((instance) => ({
```
bo-dun-1: a new prop for this is unnecessary, leverage the original documents prop. you can make it polymorphic and able to take in objects as well instead of straight UUID strings

bo-dun-1 (approved):

PR #1693, opened 2026-07-15
feat: document review editor for the Generate Documents action

trebitowski:
Adds the read-only review editor the `open_in_editor` envelope action opens: a
pdf.js viewer (lazy-loaded) with a page sidebar, thumbnails, keyboard paging and
a configurable toolbar, plus the client methods that generate for review and
finalize the reviewed envelopes.

- The toolbar's buttons come from the action's `editor_toolbar_actions`; an
  unconfigured editor still gets a Continue button that finalizes with `fill`.
- A shared poll loop backs generate and finalize; it resolves rather than hanging
  on a non-JSON body, and treats a dropped response as an error instead of
  success.
- `feathery.generateDocuments` can drive the same flow, so a logic rule reaches
  the editor and the envelope actions.

### DIFF
--- a/src/Form/index.tsx
+++ b/src/Form/index.tsx
@@ -1228,7 +1228,7 @@ function Form({
                 data.files,
                 replaceTextVariables(action.envelope_zip_name)
               );
-            } else if (envAction === 'save') {
+            } else if (envAction === 'save' && data.files) {
               let files = data.files;
               if (files.length === 1) files = files[0];
               const newValues = { [action.save_document_field_key]: files };
@@ -2898,7 +2898,7 @@ function Form({
               data.files,
               replaceTextVariables(action.envelope_zip_name)
             );
-          } else if (envAction === 'save') {
+          } else if (envAction === 'save' && data.files) {
             let files = data.files;
             if (files.length === 1) files = files[0];
             const newValues = { [action.save_document_field_key]: files };
--- a/src/Form/index.tsx
+++ b/src/Form/index.tsx
@@ -218,7 +218,7 @@ import {
 import { verifyAlloyId } from '../integrations/alloy';
 import { useFlinksConnect } from '../integrations/flinks';
 import { isNum } from '../utils/primitives';
-import { getSignUrl } from '../utils/document';
+import { editorContainerId, getSignUrl } from '../utils/document';
 import QuikFormViewer from '../elements/components/QuikFormViewer';
 import { createSchwabContact } from '../integrations/schwab';
 import { getLoginStep } from '../auth/utils';
@@ -249,6 +249,10 @@ import { hasDirtyDocxEditors } from '../elements/components/DocxEditor/docxDirty
 
 const UNSAVED_DOCX_MESSAGE =
   'You have unsaved changes in the document editor. If you leave now, your changes will be lost.';
+
+const DocumentViewer = React.lazy(
+  () => import('../elements/components/DocumentViewer')
+);
 
 export * from './grid/StyledContainer';
 export type { StyledContainerProps } from './grid/StyledContainer';
@@ -544,6 +548,7 @@ function Form({
 
   const [showQuikFormViewer, setShowQuikFormViewer] = useState(false);
   const [quikHTMLPayload, setQuikHTMLPayload] = useState('');
+  const [reviewViewerPayload, setReviewViewerPayload] = useState<any>(null);
   const { openFlinksConnect, flinksFrame } = useFlinksConnect();
 
   // When the active step changes, recalculate the dimensions of the new step
@@ -1200,6 +1205,84 @@ function Form({
         backNavMap,
         visiblePositions: newVisiblePositions,
         client,
+        // Runs the full Generate Documents flow for the
+        // `feathery.generateDocuments` logic-rule method, mirroring the
+        // ACTION_GENERATE_ENVELOPES handler (generate → optional review modal →
+        // envelope action) but resolving a promise instead of advancing the
+        // action flow.
+        generateEnvelopeFlow: async (
+          action: Record<string, any>,
+          signerEmail?: string
+        ) => {
+          // `actingAction` is the outcome to apply — the configured
+          // envelope_action on the direct path, or the toolbar button the
+          // filler pressed in the editor, where envelope_action is
+          // 'open_in_editor' and carries no outcome of its own.
+          const runEnvelopeAction = async (
+            data: any,
+            actingAction?: string
+          ) => {
+            const envAction = actingAction ?? action.envelope_action;
+            if (envAction === 'download' && data.files) {
+              await downloadAllFileUrls(
+                data.files,
+                replaceTextVariables(action.envelope_zip_name)
+              );
+            } else if (envAction === 'save') {
+              let files = data.files;
+              if (files.length === 1) files = files[0];
+              const newValues = { [action.save_document_field_key]: files };
+              updateFieldValues(newValues);
+              client.submitCustom(newValues);
+            } else if (!envAction || envAction === 'sign') {
+              // Feathery hosted eSign: open the signing page. The action-flow
+              // redirect/registerEvent variant is step-specific, so it's not
+              // run for logic-rule invocations.
+              openTab(getSignUrl(action.redirect));
+            }
+          };
+          const data = await client.generateEnvelopes(action, signerEmail);
+          if (!data) throw new Error('Document generation failed');
+          if (data.status === 'error') throw new Error(data.message);
+          if (action.envelope_action === 'open_in_editor') {
+            return await new Promise((resolve, reject) => {
+              // What finalize returned, so the method resolves with the real
+              // outcome ({files: [...]}) — the generate payload here only holds
+              // {documents, expires_at}.
+              let finalized: any;
+              setReviewViewerPayload({
+                payload: data,
+                action,
+                onFinalize: async ({ envelopes, envelopeAction }: any) => {
+                  const result = await client.finalizeEnvelopeReview(action, {
+                    envelopes,
+                    envelopeAction
+                  });
+                  if (!result) {
+                    return { status: 'error', message: 'Finalize failed' };
+                  }
+                  if (result.status === 'error') return result;
+                  await runEnvelopeAction(result, envelopeAction);
+                  finalized = result;
+                  return result;
+                },
+                onComplete: () => {
+                  setTimeout(() => setReviewViewerPayload(null), 500);
+                  resolve(finalized);
+                },
+                // Closing the viewer without continuing has to settle the
+                // promise, or an awaiting logic rule hangs for the life of the
+                // page.
+                onClose: () =>
+                  reject(
+                    new Error('Document review was closed before completing')
+                  )
+              });
+            });
+          }
+          await runEnvelopeAction(data);
+          return data;
+        },
         fields,
         products: Object.seal(
           getSimplifiedProducts(integrations?.stripe, updateFieldValues, client)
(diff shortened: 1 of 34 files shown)


src/utils/featheryClient/integrationClient.ts
```diff
@@ -454,21 +455,199 @@ export default class IntegrationClient {
...
+          console.warn(message);
+          resolve({ status: 'error', message });
+        }
+      };
+
+      const checkCompletion = async (): Promise<void> => {
+        attempts += 1;
+        let response;
+        try {
+          response = await this._fetch(pollUrl);
```
faizalsomani: This calls `this._fetch(pollUrl)` with the default `parseResponse = true`, and that turns every non-2xx poll response into a silent retry.

`_fetch` (line 145) forwards to `apiFetch(sdkKey, url, options, parseResponse)`, and with `parseResponse` true `client-utils` runs `checkResponseSuccess`, which throws for anything outside 200/201/202 - `FetchError('Internal server error')` on 500, `FetchError('Unknown error')` on 502/504, `FetchError("Can't find object")` on 404. `_fetch`'s catch only swallows conflict/auth/ignored-TypeError and rethrows the rest, so the throw lands in the bare `catch {}` on line 634 that is commented as handling transient network errors.

Consequences:

- The `resolve({ status: 'error', message: parseAPIError(data) })` on line 644 is unreachable. A non-ok poll response can never get past line 636.
- A 500 from `document/form/finalize/poll/` is retried every 2s for the full `FINALIZE_MAX_TIME`, then resolves `{ status: 'error', message: 'Document finalize took too long...' }`. The user waits 3 minutes and is told it was slow when it actually failed on the first attempt. Same for generation with `ENVELOPE_MAX_TIME`.
- A 403 is worse. `_fetch` routes it to `handleFormAuthenticationError` and returns undefined, which sets `initState.authenticationError`, so every subsequent `_fetch` short-circuits to undefined and the loop burns all 90 attempts before giving up.

`client-utils`' own `pollForCompletion` - the function this is modelled on - passes `false` explicitly: `apiFetch(sdkKey, pollUrl, { method: 'GET' }, false)`. I think the `false` just got dropped in the port. `this._fetch(pollUrl, { method: 'GET' }, false)` restores the intended behaviour and makes line 644 live.

src/Form/index.tsx
```diff
@@ -2706,6 +2707,54 @@ function Form({
...
+                }[];
+                attachments: Record<string, any>[];
+                envelopeAction: 'sign' | 'fill' | 'download' | 'save';
+              }) => {
+                const result = await client.finalizeDocumentReview(action, {
+                  envelopes,
+                  attachments,
+                  envelopeAction
+                });
+                if (result?.status === 'error') return result;
```
faizalsomani: `finalizeDocumentReview` returns `undefined` on a dropped request, and `?.` turns that into a success here.

In `integrationClient.ts` finalize does `const response = await this._fetch(url, options, false); if (!response) return;`. `_fetch` resolves undefined without throwing in four cases: `initState.authenticationError` already set, a 409 handled by `handleFormConflict`, a 403 handled by `handleFormAuthenticationError`, and an ignored TypeError - which covers a plain `Failed to fetch` / `Load failed` network blip, since `propagateNetworkErrors` is false here and those messages are in `TYPE_MESSAGES_TO_IGNORE`.

So on a network blip during finalize: `result` is `undefined`, `result?.status === 'error'` is false, and line 2745 runs `runEnvelopeAction(undefined)`. For the default sign action (`envelope_action` unset) that branch never touches `data` - it just calls `registerEvent` and sets `location.href` to the sign URL. Then `return result` hands `undefined` back to the viewer, where `continueReview` (`DocumentViewer/index.tsx:337`) sees `result?.status !== 'error'` and calls `onComplete()`, which runs `flowOnSuccess(i)` and advances the form.

Net effect: the finalize request never reached the server, and the filler is redirected to signing and the form moves on as if it had. Nothing is logged and nothing is shown.

For `envelope_action: 'save'` the same undefined hits line 2704 `let files = data.files; if (files.length === 1)` and throws a raw TypeError, which `continueReview` catches and renders to the filler as `Cannot read properties of undefined (reading 'files')`. Note line 2697 guards the download branch with `&& data.files` but the save branch has no equivalent.

Treating a missing result as a failure would cover both - `if (!result || result.status === 'error') return result ?? { status: 'error', message: '...' };`

faizalsomani (commented):
Reviewed `integrationClient.ts` and the `Form/index.tsx` wiring closely, plus the viewer's finalize handler. Two findings below, both on the error paths. I did not give the viewer internals (`DocumentCanvas`, `AnnotationLayerStyles`, the field layer, the sidebar) a proper line-by-line pass - that is roughly 1.5k lines of new rendering code and I would rather say so than pretend I covered it.

src/utils/featheryClient/integrationClient.ts
trebitowski: Correct, and fixed in 98bdde9c — the `false` did get dropped in the port. Now:

```ts
response = await this._fetch(pollUrl, { method: 'GET' }, false);
```

Your trace was exactly right, including the part I'd have missed: line 644's `resolve({ status: 'error', ... })` was dead code, so the 500 case was indistinguishable from a slow worker. The 403 path was the worse of the two — `handleFormAuthenticationError` setting `initState.authenticationError` makes every subsequent `_fetch` short-circuit to undefined, so the loop burned all 90 attempts and poisoned the rest of the session's requests.

Added a regression test that fails without the fix (`surfaces a failing poll response immediately instead of retrying to timeout`): a 500 from the poll now resolves the error after exactly one poll rather than after `FINALIZE_MAX_TIME`. Verified it fails on the old line and passes on the new one.

src/Form/index.tsx
trebitowski: Fixed in 98bdde9c, treating a missing result as failure as you suggested:

```ts
if (!result)
  return {
    status: 'error',
    message: 'Failed to finalize documents. Please try again.'
  };
if (result.status === 'error') return result;
await runEnvelopeAction(result);
```

The sign case you described was the bad one — redirecting the filler to signing and advancing the form after a request that never reached the server, with nothing logged and nothing shown.

Guarding at this single point covers both branches you called out, including the `save` path's raw `Cannot read properties of undefined (reading 'files')`, since `runEnvelopeAction` can no longer be reached with `undefined`. I deliberately did not add a second guard inside the viewer's `continueReview` — this is the only place `undefined` originates, so one guard here is the root-cause fix rather than one per consumer.

You're right that line 2697 guards `download` with `&& data.files` while `save` has no equivalent; that asymmetry predates this PR on the non-review path. Left alone since `data` can no longer be undefined at that point, but noting it's still there.

No test on this one: `Form/index.tsx` is ~3.4k lines and exercising the closure means rendering the whole form, which didn't seem worth it for a four-line early return. Say the word if you'd rather have the coverage.

src/Form/index.tsx
```diff
@@ -2749,6 +2754,41 @@ function Form({
         const envelopeId = `envelope-${i}`;
         updateEnvelopeGeneration(envelopeId, { status: 'incomplete' });
         await Promise.all([submitPromise, client.flushCustomFields()]);
+        // Shared with the review-step finalize response: runs the same
```
bo-dun-1: can we pull all of this button action logic out into another file? this is a lot of code to be bloating the form component with

src/Form/index.tsx
```diff
@@ -2777,35 +2818,55 @@ function Form({
...
-                replaceTextVariables(action.envelope_zip_name)
-              );
-            } else if (envAction === 'save') {
-              let files = data.files;
-              if (files.length === 1) files = files[0];
-              const newValues = { [action.save_document_field_key]: files };
-              updateFieldValues(newValues);
-              client.submitCustom(newValues);
-            }
+          } else if (action.review_documents) {
```
bo-dun-1: ditto on the code here

bo-dun-1: the whole generate envelope function probably needs to be pulled out

src/utils/featheryClient/integrationClient.ts
```diff
@@ -563,6 +587,170 @@ export default class IntegrationClient {
...
+      this.FINALIZE_MAX_TIME,
+      'Document finalize'
+    );
+  }
+
+  // Shared GET-poll loop for endpoints whose async path reports completion
+  // via `{ status: 'complete', ... }` (mirrors the poll pattern used by
+  // client-utils' generateFormDocuments / generateQuikEnvelopes, but routed
+  // through this._fetch so auth/conflict handling stays consistent).
+  private pollUntilComplete(
```
bo-dun-1, line 722: can we refactor all of our other polling endpoints (quik, document extraction) to also use this?

src/utils/featheryClient/integrationClient.ts
```diff
@@ -563,6 +587,170 @@ export default class IntegrationClient {
...
+      this.ENVELOPE_CHECK_INTERVAL,
+      this.ENVELOPE_MAX_TIME,
+      'Document generation'
+    );
+  }
+
+  FINALIZE_CHECK_INTERVAL = 2000;
+  FINALIZE_MAX_TIME = 3 * 60 * 1000;
+
+  async finalizeDocumentReview(
```
bo-dun-1: wouldn't this be finalizeEnvelopeReview

src/utils/featheryClient/integrationClient.ts
bo-dun-1: also feel like this method should be ordered elsewhere, not in the middle of just the document methods

src/utils/featheryClient/integrationClient.ts
```diff
@@ -720,6 +908,29 @@ export default class IntegrationClient {
     });
   }
 
+  async uploadDocumentAttachment(
```
bo-dun-1: uploadEnvelopeAttachment?

src/elements/components/DocumentViewer/index.tsx
```diff
@@ -0,0 +1,408 @@
...
+  // The toolbar exposes a single Continue action (label varies by
+  // `envelope_action`) that calls this to finalize the reviewed envelopes.
+  onFinalize?: (params: {
+    envelopes: { envelopeId: string; fieldOverrides: Record<string, any> }[];
+    attachments: { id: string; position: 'before' | 'after' }[];
+    envelopeAction: ReviewEnvelopeAction;
+  }) => Promise<{ status?: string; message?: string } | void>;
+}
+
+export default function DocumentViewer({
```
bo-dun-1, line 70: is the documentviewer code dynamically imported into the main library?

src/Form/index.tsx
trebitowski: Agreed, but I'd rather not do it in this PR — it's a refactor of structure that predates it. `runElementActions` is ~840 lines across **31** `ACTION_*` branches in a 3,403-line component; this PR added one branch to an already-overloaded chain.

Two concrete reasons to split it out:
- Three stacked PRs (#1704, #1705 and their backend/frontend counterparts) all touch this same branch. Refactoring underneath them means resolving it three more times.
- Any registry has to preserve some non-obvious control flow: the AI-extraction branch consumes following actions via `i++` and the chain uses `i--` to rewind, and `break` aborts the remaining chain on error. That's exactly the kind of thing worth doing deliberately rather than mid-review.

Left a `TODO (tyler):` at the branch in `b2c64f78` and wrote up a ticket covering the handler-module split, the interface the branches actually need, and those control-flow traps — starting with `ACTION_GENERATE_ENVELOPES` since it's the biggest and most self-contained.

Side benefit worth calling out: the finalize guard @faizalsomani caught on this PR has no direct test precisely because exercising that closure means rendering the whole form. Extracting handlers makes that testable.

src/utils/featheryClient/integrationClient.ts
trebitowski: Agreed, and there's a live bug hiding in the duplication — but same reasoning as the other thread, it reaches beyond this PR's surface into shipped flows.

`pollPersonaResponse` calls `this._fetch(pollUrl)` with the default `parseResponse = true`, so `checkResponseSuccess` throws on any non-2xx and its own `response?.status === 400` branch is **unreachable** — a hard failure becomes silent retries until `PERSONA_MAX_TIME`. That's the identical bug @faizalsomani found in the finalize poll here, and it's pre-existing on master.

So the consolidation is worth doing on its own merits, not just for dedupe. Ticket written up covering: auditing each flow (persona, Quik, AI extraction), migrating the ones whose contract is `{ status: 'complete' }`, fixing persona's `parseResponse` regardless of whether it migrates, and whether this belongs in `@feathery/client-utils` next to the existing `pollForCompletion` rather than duplicating the pattern in the SDK.

`TODO (tyler):` left at the helper in `b2c64f78`, noting the persona bug specifically.

src/utils/featheryClient/integrationClient.ts
trebitowski: Renamed to `finalizeEnvelopeReview` in `68031459` — you're right, it operates on `Envelope` rows, not Documents.

src/utils/featheryClient/integrationClient.ts
trebitowski: Renamed to `uploadEnvelopeAttachment` in `68031459` — same reasoning, it creates a document-less `Envelope` row.

src/elements/components/DocumentViewer/index.tsx
trebitowski: Yes — confirmed with a real build rather than just reading the source, since `React.lazy` only yields a separate chunk if the bundler emits one.

From a fresh `rollup -c`:

- Viewer lands in its own chunk, `dist/fthry_index.DI9J4p_P.js`, **86 KB**.
- `dist/index.js` contains **zero** occurrences of `Review Your Forms` / `annotationMode`, and zero of `pdfjs-dist`.
- The reference is a genuine deferred import: `function () { return import('./fthry_index.DI9J4p_P.js'); }`.
- The pdf.js CDN URL lives *inside* the viewer chunk, so even the jsDelivr fetch can't happen until the chunk loads.

So it's two levels of laziness — the 86 KB chunk isn't fetched until the Generate Documents action fires, and pdf.js only after that. Splitting is enabled in all three targets: rollup ESM/CJS via `dir` + `chunkFileNames`, UMD via `chunkFilename: '[name].[contenthash].js'`.

Unrelated but worth knowing: `yarn build` currently fails in a worktree ("This package doesn't seem to be present in your lockfile") — I had to invoke `./node_modules/.bin/rollup -c` directly.

arykhoda (approved):

PR #1699, opened 2026-07-21
feat: in-form DOCX document editor with save, download, and sign flows

ayeshah:
## Changes

In-form DOCX document editor: users can review and edit generated documents inside a form container before downloading or signing.

- **Editor suite** (`src/elements/components/DocxEditor/`): reusable `DocxEditor` (I/O-agnostic — bytes in via `source`, bytes out via `onSave`), Syncfusion loaded from the CDN at runtime (no bundle impact), and a custom toolbar split into focused modules (`DocxToolbar/`) — per-group components, portaled menus, table insertion, Feathery-themed accents.
- **Responsive toolbar**: tool groups collapse tail-first into a "More" menu when the container is narrow and return as space allows; fit is measured against a hidden copy of the full row and recomputed via ResizeObserver, so late container sizing in hosted forms can't strand the layout. Tool cluster stays centered against the page.
- **Container wiring**: `DocumentEditorContainer` resolves its document from the Generate Documents action targeting it (`view_draft_container`), loads the newest envelope, and reloads on the generate action's refresh event.
- **Save**: edits PATCH back to the envelope (`document/envelope/<id>/file/`). Disabled `keepalive` on the request — Chromium rejects keepalive bodies >64KB and the failure was silently swallowed, letting users lose work; an undefined response is now surfaced as a failure.
- **Download**: DOCX (client export of the exact saved bytes) or PDF (server-side conversion via the new envelope pdf endpoint — non-persisting, the draft stays editable). The Download control is a DOCX/PDF menu; hidden for save-to-field actions.
- **Sign**: finalizes the envelope first (docx→PDF + signature-field injection via the new finalize endpoint), restoring master's sign-ceremony contract while keeping drafts editable until the user commits; the editor then flips read-only for that envelope. Signer email is withheld at generation for editor-targeted actions (it would trigger generation-time PDF conversion and make the draft uneditable) and forwarded at finalize instead.
- **New action**: `Sign Documents` re-enters the sign flow without regenerating, preserving edits.

Merge gates: requires the backend PR below; a release containing this must publish before the builder/hosted-forms PRs merge (both consume `@feathery/react`).

Known follow-ups (from cross-repo review): sign-tab popup-blocker hardening, `updateTheme()` × form-schema cache interaction, surfacing save/finalize errors in the UI, multi-document sign finalization.

- [x] Tested end to end

## Related pull requests

- Backend endpoints: feathery-org/feathery-backend#3509
- Builder support: feathery-org/feathery-frontend#3627


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@feathery/react",
-  "version": "2.60.0",
+  "version": "2.52.1",
```
trebitowski: remove

trebitowski (approved):

trebitowski (approved):

PR #1705, opened 2026-07-23
feat: polymorphic documents array for Quik-sourced Generate Documents

trebitowski:
`documents` entries may now be source objects ({kind: 'quik'}) alongside plain
template ids, forwarded verbatim to the generate endpoint.

- A polymorphic entry takes the direct endpoint call, because the library helper
  interpolates ids into its poll URL and a source object would stringify to
  "[object Object]" and never match the backend's cache key.
- The headless path, which has no mounted <Form /> to run the editor flow,
  rejects source objects with an explanatory error instead of polling a key the
  backend never wrote until it times out.

### DIFF
--- a/src/utils/featheryClient/integrationClient.ts
+++ b/src/utils/featheryClient/integrationClient.ts
@@ -29,6 +29,11 @@ import {
 import { handleFormAuthenticationError, handleFormConflict } from './utils';
 import { editorContainerId, isDocusignSignAction } from '../document';
 
+// A configured Generate Documents entry in the ordered `documents` array: a
+// template UUID string, or the single polymorphic `{kind:'quik'}` source dict.
+// The SDK forwards these verbatim (action config -> request field).
+export type GenerateDocumentRef = string | { kind: string; [key: string]: any };
+
 export const TYPE_MESSAGES_TO_IGNORE = [
   // e.g. https://sentry.io/organizations/feathery-forms/issues/3571287943/
   'Failed to fetch',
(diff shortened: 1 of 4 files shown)


src/utils/featheryClient/integrationClient.ts
```diff
@@ -536,7 +541,20 @@ export default class IntegrationClient {
     if (!response.ok) return { status: 'error', message: parseAPIError(data) };
     if (!runAsync || data.documents) return data;
 
-    const pollUrl = `${getApiUrl()}document/form/generate/poll/?fid=${userId}&dids=${documentIds}`;
+    // Poll `dids` must match the backend's document_cache_keys: a template's
+    // UUID string, or the literal "quik" for the quik item. Interpolating the
+    // raw array would stringify a {kind:'quik'} entry to "[object Object]" and
+    // never match the cache.
+    const dids = documentIds
```
faizalsomani, line 684: This mapping is right, but it only covers `generateEnvelopesDirect`. The other branch of `generateEnvelopes` - line 487, the one every non-review, non-DocuSign action takes - still hands the raw polymorphic array to `client-utils`, which has the exact stringification bug described in the comment just above.

`client-utils`' `generateFormDocuments` builds its poll URL as ``` `...generate/poll/?fid=${userId}&dids=${documentIds}` ``` with no mapping. So for `documents: [{kind:'quik'}, '<uuid>']` it sends `dids=[object Object],<uuid>`.

Server side, `APIEnvelopeGeneratePollView` splits that on commas and hashes it, while the generate request cached under `document_cache_keys(...)`, which yields `['quik', '<uuid>']` (feathery-backend#3525, `apps/document/utils/document_items.py`). The hashes do not match, so the running-flag lookup misses, the payload lookup misses, and the poll raises `ValidationError("No document generation")`.

Concrete case: a Generate Documents action with the Quik source plus a template, `review_documents` off, `sign_method` unset - so Feathery sign, or download, or save. The documents generate fine on the server, but the SDK's first poll comes back 400 and the filler is shown `No document generation` while their documents sit in the cache under the correct key.

Routing any action whose `documents` array contains a non-string entry through `generateEnvelopesDirect` would cover it, or the same mapping needs to go into `client-utils`.

faizalsomani (commented):
Reviewed the incremental diff on top of #1704. One finding below on the `dids` handling. The findings I left on #1693 (`pollUntilComplete` parseResponse, and the undefined-finalize-treated-as-success path) carry into this branch unchanged.

src/utils/featheryClient/integrationClient.ts
trebitowski: Good catch — fixed. `generateEnvelopes` now also routes through `generateEnvelopesDirect` when the `documents` array contains any non-string entry, so the polymorphic case gets the mapped `dids` instead of `client-utils`' raw interpolation. Covered by a test for the plain (no review, no DocuSign) download path with `["doc1", {kind:"quik"}]`.

arykhoda (approved):

PR #1712, opened 2026-07-24
feat: add default value for button groups

mattstringer:
## Changes
This is one of three PRs to support adding a default value to button groups. A default value in this context is a button that is pre-selected within the form. Functionality has also been added for default multi-select, if multiple selections are enabled in the button group.

These specific changes handle single vs multi select button logic, and adds 4 test cases.

- [X] Cleaned up debug prints, comments, and unused code
- [X] Added / updated unit tests
- [X] Tested end to end if this affects other systems

## Related pull requests
Other PRs here that are related to this change:
-[backend PR](https://github.com/feathery-org/feathery-backend/pull/3532)
-[frontend PR](https://github.com/feathery-org/feathery-frontend/pull/3655)

### DIFF
--- a/src/utils/fieldHelperFunctions.ts
+++ b/src/utils/fieldHelperFunctions.ts
@@ -154,6 +154,14 @@ export function getDefaultFieldValue(field: any) {
     if (['multiselect', 'dropdown_multi'].includes(servar.type)) {
       return meta.default_value.split(',').map((val: string) => val.trim());
     }
+    if (servar.type === 'button_group') {
+      // For button_group, always return an array since it's an ARRAY_FIELD_TYPE
+      // Handle both single and multiple selection
+      if (meta.multiple) {
+        return meta.default_value.split(',').map((val: string) => val.trim());
+      }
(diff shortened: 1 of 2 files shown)


src/utils/fieldHelperFunctions.ts
```diff
@@ -154,6 +154,14 @@ export function getDefaultFieldValue(field: any) {
     if (['multiselect', 'dropdown_multi'].includes(servar.type)) {
       return meta.default_value.split(',').map((val: string) => val.trim());
     }
+    if (servar.type === 'button_group') {
```
benminor: This should be collapsed into the above if statement on line 154

benminor: `if (['multiselect', 'dropdown_multi', 'button_group']` etc.

bo-dun-1 (approved):

src/utils/fieldHelperFunctions.ts
mattstringer: fixed thank you

PR #1715, opened 2026-07-27
feat: Attach document ops and client-side logic rules to the SyncFusion editor

faizalsomani:
Attaches the Robin document-editing tool surface and client-side logic-rule execution to `DocumentEditorContainer` / `onEditorReady`, replacing the retired `docx_editor` servar path. Consolidates the previously-duplicated SyncFusion document-ops engine (`docxEditorBridge.ts` vs `feathery-frontend`'s `syncfusionDocumentOps.ts`) into a single implementation in `@feathery/react`, carrying its test suite.

**Architecture:** [Robin × DOCX Editor - Architecture](https://app.notion.com/p/3aa753ac481581d39494f2b0d64dfb59) (AI Team Wiki) - two diagrams and a short read on how the pieces fit. Worth skimming before reviewing this diff; the client-forwarded tool-call model is not obvious from the code alone.

## What this changes

- Emits an assistant document target for the in-form document so `ai-services`' `hasDocumentTarget` gate turns the doc tools on.
- Adds the `find_document_occurrences` handler to the tool dispatch.
- `runClientSideLogic` / `runLogicRuleById` no longer swallow rule exceptions as a false success, and gain the same `derivedUpdates` safety net the server path has.
- **A rule that only mutates a form field now reflects into the document.** Field values are snapshotted either side of a rule run and the difference surfaced as `derivedUpdates`, so a rule that returns nothing still propagates - previously it silently changed the field and reported success while the document was untouched.
- New-paragraph insertion inherits neighbour formatting by default, computed in the engine rather than left to the model to request. Formatting applied after a structural insert now resolves against the post-insert document instead of pre-insert indices.
- **Mid-table `insert_row` works.** Cell fills are deferred until the new row exists, rather than written against guessed anchors; only end-of-table appends worked before.
- A capability registry keeps the advertised op list and the real dispatch from drifting apart. Handlers are typed against their registry entries, so a mismatch does not compile - backed by a contract test per op, and per meaningful parameter variant, over a real `DocumentEditor`.
- Tracked-write proof compares the whole-document reject projection rather than the anchor index, so a paragraph-splitting insert is no longer misreported as an untracked write and rolled back.
- `OpError` survives the ES5 bundle, so structured error codes and their details reach the caller instead of collapsing to a bare `op_failed`.
- Fixes a document-index race: a progressively-loaded document could be certified fresh from its first, TOC-only content sighting, serving stale semantic-search results until the first edit.

## Base

Targets `feat/docx-editor` directly, not `master` - this only merges after that one. Reviewers see only our delta against Ayesha's branch; her commits are the base, not part of this change.

Teammate commits carried locally for end-to-end testing (Anthony's #1698, #1655, #1696, #1707, #1685, #1701) are **not** in this cut - all are already merged to `master` and drop out when the base catches up.

Verified end to end against a real 33-page Hilb proposal on a published form.


src/assistant/tools/assistantToolDispatch.ts
```diff
@@ -0,0 +1,475 @@
...
+
+// Compose document updates from a rule's changed-field details: one
+// { field, previous?, value, describes? } per field whose NEW value is usable
+// exact text. `previous` (from the pre-invoke snapshot) rides along when it
+// is usable search text; `describes` (via opts.describeField) names what the
+// text is for the semantic-search fallback. Fields already covered by the
+// rule's explicitly returned updates (opts.explicitUpdates) are skipped so
+// the same edit is never surfaced twice.
+export const composeDerivedRuleUpdates = (
+  details: ChangedFieldDetail[],
```
lanthony42: One change set in three shapes (changedFields, changedFieldDetails, derivedUpdates renamed), plus note duplicating prompts/assistant.ts:93. One list drops the dedup too.

src/assistant/tools/getFormFields.ts
```diff
@@ -0,0 +1,297 @@
...
+      : { value, truncated: false };
+  }
+  let serialised: string;
+  try {
+    serialised = JSON.stringify(value) ?? String(value);
+  } catch {
+    serialised = String(value);
+  }
+  return serialised.length > maxChars
+    ? { value: `${serialised.slice(0, maxChars)}…`, truncated: true }
```
lanthony42: File field values are File/Promise; returned by reference here, structuredClone throws in addToolOutput and the turn hangs. setFieldValue.ts:403 round-trips.

src/Form/logic.ts
```diff
@@ -583,12 +594,254 @@ export const runClientSideLogic = async (
...
+    });
+    return {
+      changedFields: changedFieldDetails.map((d) => d.key),
+      changedFieldDetails,
+      ...(derivedUpdates.length > 0 ? { derivedUpdates } : {}),
+      ...(changedFieldDetails.length > 0
+        ? { documentEdited: false as const, note: RULE_FIELDS_CHANGED_NOTE }
+        : {}),
+      returnValue
+    };
```
lanthony42: Same - returnValue goes out raw, so a File return hangs the turn.

src/assistant/tools/syncfusionDocumentOps.ts
```diff
@@ -0,0 +1,6221 @@
...
+  insert_text: ({ editor, op, block }) => {
+    const offset = insertionPoint(op, block);
+    selectRange(editor, block.anchor, offset, offset);
+    editor.editor.insertText(insertionText(op));
+  },
+  set_cell_text: ({ editor, op, block, byAnchor }) => {
+    // The engine, not the prompt, keeps model arithmetic out of numeric cells.
+    // This runs BEFORE any write, so a refused numeric write costs no revision.
+    const literalNumber = guardModelAuthoredNumber(op, block, byAnchor);
+    // Overwrite the (cell) block's content.
```
lanthony42, line 3957: Only call site, so the gate is set_cell_text-only - replace_text writes the number it just refused.

src/assistant/tools/syncfusionDocumentOps.ts
```diff
@@ -0,0 +1,6221 @@
...
+      'reference_not_found',
+      `The table at ${tableAnchor} has ${collected.columns} column${
+        collected.columns === 1 ? '' : 's'
+      } (0-${collected.columns - 1}); the anchor names column ${column}.`
+    );
+
+  const requestedStart = integerParam(op.startRow, 'startRow');
+  const requestedEnd = integerParam(op.endRow, 'endRow');
+  const wholeTable = requestedStart == null && requestedEnd == null;
+  const startRow = requestedStart ?? 0;
```
lanthony42: Numeric headers do produce a value - probed `2025 | 2026`, row 0 became `263`.

src/assistant/AssistantChat.tsx
```diff
@@ -294,16 +314,49 @@ const AssistantChat = ({
...
+
   const buildChatBody = (): Record<string, unknown> => {
+    // ai-services reads assistant scope exclusively from body.context;
+    // form_key stays top-level because backend form auth reads body params.
     const body: Record<string, unknown> = {};
     if (formKey) body.form_key = formKey;
+    const context: Record<string, unknown> = {};
     const targets = getTargets();
-    if (targets.length > 0) body.targets = targets;
+    if (targets.length > 0) context.targets = targets;
```
lanthony42, line 331: The mixin still reads targets off the top level, so {{field_key}} in admin instructions never resolves and saves to the thread literal.

src/assistant/tools/syncfusionDocumentOps.ts
```diff
@@ -0,0 +1,6221 @@
...
+ * `expect` against a genuinely empty block passed before and passes now, so no
+ * satisfiable expectation is being discarded - and the surviving guards
+ * (anchor resolution, the `find`-text check, and the tracked-write reject
+ * projection) are untouched.
+ */
+function expectGuardRefuses(expected: unknown, live: string): boolean {
+  if (expected == null) return false;
+  const wanted = String(expected);
+  if (wanted === '') return false;
+  return !expectTextMatches(wanted, live);
```
lanthony42: set_cell_text has no `find`, so expect was its only CAS - expect:'' now overwrites off a stale anchor.

src/assistant/tools/syncfusionDocumentOps.ts
```diff
@@ -0,0 +1,6221 @@
...
+        )
+      ]
+    );
+  if (expect == null) return;
+  // With a length pin the delivered prefix is authoritative for its own extent;
+  // without one the whole text must match exactly, as it always has.
+  const matches =
+    expectLength != null
+      ? range.text.startsWith(expect)
+      : range.text === expect;
```
lanthony42, line 2790: expectLength alone pins length only, not content.

src/elements/components/DocxEditor/DocumentEditorContainer.tsx
```diff
@@ -303,6 +307,14 @@ export default function DocumentEditorContainer({
     else openTab(url);
   }, [client, envelope, targetAction, terminalAction]);
 
+  // DocxEditor exposes its live SyncFusion instance at this exact lifecycle
+  // point. Register it for the assistant and clear it when the container
+  // unmounts so tools never mutate a stale editor.
+  const onEditorReady = useCallback((editor: any) => {
+    registerDocxEditor(undefined, editor);
+  }, []);
```
lanthony42: Undefined key here and in the unregister: editors share one key, either unmount clears it. Ayesha's picker spans every step - is multiple per form supported?

src/assistant/tools/syncfusionDocumentOps.ts
```diff
@@ -0,0 +1,6221 @@
...
+      throw new OpError(
+        'untracked_write',
+        `SyncFusion did not create a rejectable tracked ${structural} for ${op.op}.`
+      );
+    return;
+  }
+
+  if (priorRejectStream !== undefined) {
+    const nowRejectsTo = rejectProjectionStream(parseSfdt(editor.serialize()));
+    if (nowRejectsTo !== priorRejectStream)
```
lanthony42, line 4917: Serializes here and again in refresh(), per op - exhaustive replace is O(N x document) on the main thread.

src/Form/logic.ts
```diff
@@ -20,7 +20,15 @@ import {
...
-import { FormInternalState } from '../utils/internalState';
+import internalStateStore, { FormInternalState } from '../utils/internalState';
+import { getFormContext } from '../utils/formContext';
+import { getPrivateActions } from '../utils/sensitiveActions';
+import {
+  ChangedFieldDetail,
+  composeDerivedRuleUpdates,
+  DerivedRuleUpdate,
+  RULE_FIELDS_CHANGED_NOTE
+} from '../assistant/tools/assistantToolDispatch';
```
lanthony42, line 34: Form logic importing assistant tooling, and derivedUpdates runs with no document check - form-only chats get document instructions.

src/assistant/tools/assistantToolDispatch.ts
lanthony42: One change set in three shapes (changedFields, changedFieldDetails, derivedUpdates renamed), plus note duplicating prompts/assistant.ts:93. One list drops the dedup too.

src/assistant/tools/assistantToolDispatch.ts
```diff
@@ -0,0 +1,475 @@
...
+    return { handled: true, output };
+  }
+
+  return { handled: false };
+}
+
+// Build the callable_rules catalog from the form's logic rules - only
+// `trigger_event === 'tool'` rules, projected into the tool-facing shape.
+export const buildCallableRules = (logicRules: any[] = []): CallableRule[] =>
+  logicRules
```
lanthony42, line 269: No server_side filter, so server-side rules mount and throw on feathery.params when called. Filter them out for now?

src/assistant/tools/assistantToolDispatch.ts
```diff
@@ -0,0 +1,475 @@
...
+      toolName
+    );
+    return { handled: true, output };
+  }
+
+  // 2) Built-in docx bridge reads/edits.
+  if (
+    toolName === 'getDocumentInventory' ||
+    toolName === 'get_document_inventory'
+  ) {
```
lanthony42: ai registers camelCase only, so these are unreachable.

src/assistant/tools/assistantToolDispatch.ts
```diff
@@ -0,0 +1,475 @@
...
+  text: string;
+  isCollapsed: boolean;
+};
+
+export const SELECTION_TEXT_LIMIT = 500;
+
+// Normalize a host getSelection() result into the request payload shape: text
+// clamped to 500 characters, or null when there's nothing usable.
+export const normalizeSelection = (
+  raw: AssistantSelection | null | undefined
```
lanthony42: Seems unused - readSelection is the live one.

src/assistant/tools/numericCells.ts
```diff
@@ -0,0 +1,692 @@
...
+    const units = value.units * factor;
+    return safeUnits(units) ? { units, scale } : null;
+  }
+  const factor = pow10(value.scale - scale);
+  if (value.units % factor !== 0) return null;
+  return { units: value.units / factor, scale };
+}
+
+function toCommonScale(values: NumericValue[]): NumericValue[] | null {
+  const scale = values.reduce((max, v) => Math.max(max, v.scale), 0);
```
lanthony42: Seems unused.

src/assistant/AssistantChat.tsx
```diff
@@ -294,16 +314,49 @@ const AssistantChat = ({
...
 
     if (instanceId) {
       const panelRuntime = getPanelRuntimeSnapshot(instanceId);
-      if (panelRuntime) body.panel_runtime = panelRuntime;
+      if (panelRuntime) context.panel_runtime = panelRuntime;
+      context.selection = readDocxSelection(getDocxEditor(instanceId));
+      const callableRules = buildCallableRules(
+        internalState[instanceId]?.logicRules ?? []
```
lanthony42: getLogicRules already does this server-side; from the browser it pins the filter into an SDK version.

lanthony42 (changes requested):

src/assistant/tools/docxEditorRegistry.ts
```diff
@@ -0,0 +1,166 @@
...
+    editor,
+    order: candidates.get(resolvedInstanceId)?.order ?? nextOrder++
+  };
+
+  if (registration && registration.stepId !== next.stepId) {
+    // React mounts the incoming step before unmounting the outgoing one. A
+    // different step is therefore a handoff, not a simultaneous-editor error.
+    // Discard outgoing candidates so their late cleanup cannot resurrect or
+    // clear the incoming step.
+    candidates = new Map([[resolvedInstanceId, next]]);
```
lanthony42: A late re-register from the outgoing step wipes the incoming candidate, and its unmount then leaves `getDocxEditor()` undefined - `:342` refires when the envelope id settles.

src/assistant/capabilities/registry.ts
```diff
@@ -0,0 +1,813 @@
...
+
+// Entries are ordered exactly like ai-services' DOCUMENT_EDIT_OPS so a
+// name-by-name comparison of the two lists reads as a clean diff.
+//
+// The array is const-asserted so every entry survives as a literal type: the
+// typed handler tables in `../tools/syncfusionDocumentOps.ts` derive each
+// handler's parameter type from its entry here (S5). `satisfies` would be the
+// idiomatic spelling, but the repo pins TypeScript 4.7, so the entry-shape
+// check is the explicit assignability statement after the array instead.
+export const DOCUMENT_EDITOR_CAPABILITIES = [
```
lanthony42, line 51: Only `op` and `requiresAnchor` are read off this, and the declaration now ships names only, so the 49 summary / 41 example / 49 params blocks reach nothing.
Drop them along with the wire trim rather than leave prose that reads like it still describes the tools?

src/assistant/capabilities/declaration.ts
```diff
@@ -0,0 +1,29 @@
+import { AdvertisedDocumentOp, DOCUMENT_EDITOR_CAPABILITIES } from './registry';
+
+/**
+ * Machine-only facts about the form document executor. ai-services owns every
+ * description, example, payload rule, schema, and result shape the model sees.
```
lanthony42: Not quite true anymore - the 94 `OpError` messages and the receipt strings are react-authored, and the model reads them on every refusal and write.

lanthony42 (changes requested):
Second pass over the fixes. Three left, the registry one is cleanup.

src/assistant/tools/docxEditorRegistry.ts
```diff
@@ -0,0 +1,172 @@
...
+  stepId?: string;
+  documentId?: string;
+  envelopeId?: string;
+};
+
+const DEFAULT_STEP_ID = '__legacy_document_step__';
+let nextOrder = 0;
+let registration: DocxEditorRegistration | undefined;
+let candidates = new Map<EditorInstanceId, DocxEditorRegistration>();
+const supersededEditors = new Set<any>();
```
lanthony42: Only `_clearDocxEditors` clears this and it's test-only, so every superseded SyncFusion instance is retained for the page's life - `WeakSet`?

src/assistant/tools/docxEditorRegistry.ts
```diff
@@ -0,0 +1,172 @@
...
+
+export type DocxEditorContext = {
+  stepId?: string;
+  documentId?: string;
+  envelopeId?: string;
+};
+
+const DEFAULT_STEP_ID = '__legacy_document_step__';
+let nextOrder = 0;
+let registration: DocxEditorRegistration | undefined;
```
lanthony42: `internalState` is keyed by form uuid but this is a single global, so two `<Form>`s on a page share one editor. Intended?

AGENTS.md
```diff
@@ -0,0 +1,11 @@
+# feathery-react
```
lanthony42: Can this be its own PR? Repo-wide agent-tooling change across all three repos, unrelated to the Hilb work.

lanthony42 (changes requested):

src/assistant/tools/docxEditorRegistry.ts
faizalsomani: Fixed in dfec0e39 - production only calls has and add so it is now a WeakSet, with the test reset replacing the set because WeakSet has no clear.

src/assistant/capabilities/registry.ts
faizalsomani: Pruned in ab1e95c9 - summary, example, anchorKind and DOCUMENT_EDITOR_READS are gone, but params and requiresAnchor stay because they drive the compile-time OpParams type and partition live handlers.

AGENTS.md
faizalsomani: Removed in ca8c143a - AGENTS.md is gone and CLAUDE.md is restored byte-identical to master, so this PR no longer touches agent tooling at all.

src/assistant/tools/docxEditorRegistry.ts
faizalsomani: Deliberate - two Forms in one step is not a supported configuration, so there is one live Robin document target per page and getDocxEditor keeps the legacy argument without selecting by it; multiple editors across different steps of one Form remain supported.

src/utils/transportValue.ts
```diff
@@ -0,0 +1,101 @@
...
+            ...(typeof record.type === 'string' && record.type
+              ? { type: record.type }
+              : {}),
+            ...(typeof record.size === 'number' ? { size: record.size } : {})
+          };
+        }
+        if (typeof (record as any).then === 'function') {
+          return { kind: 'promise', present: true };
+        }
+        if (seen?.has(record)) return { kind: 'circular', present: true };
```
lanthony42: `JSON.stringify` already handles shared refs - this marks the second occurrence circular, so `.fill()`-padded repeat rows land as `{kind:'circular'}`.

lanthony42 (changes requested):

lanthony42 (approved):

PR #1725, opened 2026-07-30
feat: assistant-defined tracked-change accept groups via revision customData

ayeshah:
## Problem

When the assistant edits a document with track changes on, review was unintuitive:

1. **One replace = two cards** — SyncFusion has no "replaced" revision type; a tracked replace is a Deletion revision + an Insertion revision, and the pane can never merge cards of different types.
2. **Accepting one change accepted everything nearby** — SyncFusion's pane groups adjacent same-author/same-type revisions (`groupedView`), and the public `Revision.accept()` resolves the *whole* adjacency group. Our bridge additionally bound the entire change-set batch into a single all-or-nothing unit.
3. **No purpose-built review surface** — the native Changes pane lists raw revisions with author-colored text and no notion of the assistant's logical edits.

### Ops layer — assistant-defined accept groups

The assistant defines the accept/reject unit explicitly: each edit op may carry a `group` label, and every op sharing the label resolves together — never more, never less.

- **Durable tagging**: before each op, the engine sets `revisionSettings.customData` to a JSON tag (`{changeSetId, group}`); SyncFusion stamps it onto every revision the op creates. The tag round-trips through SFDT/DOCX. The host's prior customData is restored after the change set.
- **Per-group atomic binding**: created revisions are partitioned by tag and each partition bound as one accept/reject decision. Ungrouped ops share the change-set-wide unit. `rebindRevisionGroups(editor)` rebuilds bindings from persisted tags after reload (editor registration + start of every change set; idempotent).
- **No adjacency cascade**: members resolve through SyncFusion's single-revision `handleAcceptReject(isAccept, false)` instead of the public `accept()`/`reject()` (public path kept as a feature-detected fallback). `resolveRevisionIndividually` resolves one member alone; `resolveRevisionsAsOneUndo` wraps a multi-revision unit in the engine's complex history so **one ⌘Z restores the whole unit** (an undone replace comes back as a pending pair, not a dangling insert).
- **Group isolation**: SyncFusion coalesces adjacent same-author/same-type revisions into one object, which would fold a second group's content into the first group's revision. `installRevisionGroupIsolation` gates the engine's merge predicates (`isRevisionMatched`, `compareTwoRevisions`) on the tag. Type-less `isRevisionMatched` calls are **ownership** checks ("is this pending insertion my own?") and stay native — gating them broke reject-reversibility for a second write over a pending insertion (`untracked_write` on text frames; caught and fixed via the pre-existing `textFrameTrackedWrite` suite).
- **Replace folding**: a Deletion whose range links directly to an adjacent same-unit Insertion is ONE edit. `listRevisionGroups` folds the pair into a `Replace` item (`beforeText`/`text`), and `findReplaceCounterpart` (memoized, both directions) lets the renderer classify either half. Human (untagged) pairs match by author.
- **Reporting & validation**: `changeSet.groups` lists each accept unit (`id`, `opIndices`, `revisionCount`); malformed `group` values are refused at preflight (`invalid_group`). `group` documented as a reserved cross-op key in the capabilities registry.

### Renderer — GitHub-style inline changes (canvas patch)

SyncFusion draws on canvas, so this is a renderer patch, not CSS:

- Insertions: green wash, document's own font color. Deletions: red wash, **red glyphs with the engine's native single strikethrough** (routed through a scoped `checkRevisionType` swap so geometry is baseline-aware). No author colors, no underlines.
- **Boundary ring** around the active edit (the one selected inline or focused in the panel): merged per line across touching runs — a replace rings as one edit — flush with the highlight, 4px radius.
- Per-edit on-canvas geometry is recorded each render pass (`getRevisionRects` + `setAfterRenderCallback`) — staged API for margin gutter bars (follow-up).
- The native Changes pane is suppressed (`isUserClosed`) — the grouped panel below is THE review surface in this host.

### Review panel — `TrackedChangeGroups`

A 340px rail beside the editor:

- **One card per accept group**, titled by the humanized group id — plus **one card per human author** for manual (untagged) tracked edits, so user changes are reviewable in the same surface with author attribution on every chip.
- Cards expand to individual edits ("chips"): type badge (`Added` / `Removed` / `Replaced`), author, and −/+ mono diff rows. A replace is ONE chip requiring ONE approval that settles both halves.
- Resolution paths: focused-chip Accept/Reject, per-card `Accept N` / `Reject N`, and rail-head `Accept all` / `Reject all` — all through the non-cascading path as one undo unit. **Resolved edits leave the rail immediately**; undo brings them back via the contentChange refresh.
- **Navigation both ways**: clicking an inline tracked change expands + focuses + rings its chip (and reopens the panel); clicking a chip selects that exact revision in the document (`selectRevision` with `skipGroupSelect`, echo-suppressed so adjacent edits don't hijack focus).
- **Keyboard**: panel-scoped J/K or arrows step pending edits, A/R resolve the focused one; ⌘Z/Ctrl+Z (and ⇧⌘Z / Ctrl+Y) are forwarded to the editor history while the rail holds focus, and focus hands back to the editor when the last edit resolves.
- Collapses to a bookmark tab on the editor edge; hidden in read-only hosts; renders nothing when no tracked changes exist.

## Tests

- Real-SDK (live `DocumentEditor`) suites: group adjacency isolation, replace fold + counterpart memo both directions, one-approval + **single-undo restores the folded pair**, rebind rehydration/idempotency, and the pre-existing text-frame reversibility suite guarding the ownership-check scoping.
- Mock-editor ops suites: group-scoped resolution, `changeSet.groups` reporting, `invalid_group` preflight, host customData restoration, individual-resolve semantics.
- `TrackedChangeGroups` component suite: card/chip rendering (assistant + human author cards), replace chip, all resolution paths, resolved-edits-disappear, inline↔panel navigation, selection-echo skip guard, J/K/A/R flow, undo-chord forwarding, focus handoff, drawer behavior, contentChange refresh.
- Full assistant + DocxEditor suites pass, plus typecheck and lint (only pre-existing unrelated `DropdownMultiField` failures remain in the full repo run).


src/elements/components/DocxEditor/DocumentEditorContainer.tsx
```diff
@@ -335,6 +339,16 @@ export default function DocumentEditorContainer({
         documentId: activeDocumentId,
         envelopeId: envelope?.id
       });
+      // The review experience (render patch, pane suppression, rail) ships
+      // with DocxEditor itself; this site only rebuilds the accept-group
+      // bindings a RELOADED document lost (isolation here is an idempotent
+      // safety net). Never allowed to affect editor registration.
+      try {
+        rebindRevisionGroups(editor);
```
faizalsomani: Rebind after the source open completes: `onEditorReady` fires before `openAsync`, so persisted revisions are absent here and a reloaded document never regains its atomic group bindings.

trebitowski (approved):

PR #1726, opened 2026-07-30
feat: attribute assistant revisions to Robin

faizalsomani:
## Captain request

> "Whenever I make the edit, the track changes, says 'guest user'. Where is that coming from and can it be changed through us? And I would like it to say 'Robin' instead of 'guest user did it'."

## Root cause and change

SyncFusion 34.1.31 captures a revision's author when it creates the revision and falls back to:

```js
owner.currentUser ? this.owner.currentUser : "Guest user"
```

We did not set `currentUser` anywhere outside SyncFusion. `applyDocumentEdits` now saves the existing editor author, sets a named `Robin` author constant for the assistant batch, and restores the prior value in the same `finally` block that restores `enableTrackChanges`.

The assistant's registered `insert_comment` operation goes through this same batch, so assistant-created comments are also attributed to Robin.

## Scope

This is deliberately assistant-only. It does not set the author when the editor mounts, so manual typing is not relabeled as Robin. Manual edits retain the editor's existing author behavior before and after an assistant batch.

## Verification

- Real SyncFusion SDK: assistant revisions are authored by Robin; manual revisions before and after remain `Guest user`.
- Failure injection: the prior author is restored after a deliberately failed assistant write.
- Comment contract: assistant-created comments are authored by Robin.
- Mutation proofs: removing the batch assignment fails the revision and comment author guards; removing restoration fails both the failure-path and manual-edit guards.
- Focused tests: 154 passed across 2 suites.
- Full React suite: 1,335 passed across 82 suites, including text-frame and tracked-change coverage.
- ESLint: clean.
- TypeScript typecheck: clean.
- Protected-file diff (`AGENTS.md`, `CLAUDE.md`, `.claude`): empty.

### DIFF
--- a/src/assistant/tools/syncfusionDocumentOps.ts
+++ b/src/assistant/tools/syncfusionDocumentOps.ts
@@ -5903,8 +5903,6 @@ export function applyDocumentEdits(
   const announcement = describeChangeSetTouches(columnTouches);
   const priorTrackChanges = editor.enableTrackChanges;
   const priorCurrentUser = editor.currentUser;
-  editor.enableTrackChanges = true;
-  editor.currentUser = ASSISTANT_DOCUMENT_AUTHOR;
   let blocks: FlatBlock[] = [];
   let byAnchor = new Map<string, FlatBlock>();
   // "What the whole document would read if every revision were rejected",
@@ -6187,6 +6185,8 @@ export function applyDocumentEdits(
       result && !result.ok && !nonBlockingStoryWriteFailures.has(index)
   );
   try {
+    editor.enableTrackChanges = true;
+    editor.currentUser = ASSISTANT_DOCUMENT_AUTHOR;
     if (preflightFailed) {
       warnings.push(
         `change_set_preflight_failed: ${changeSetId}; no structural or formatting writes were attempted.`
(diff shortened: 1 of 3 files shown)


src/assistant/tools/syncfusionDocumentOps.ts
```diff
@@ -5900,7 +5902,9 @@ export function applyDocumentEdits(
   const columnTouches = collectColumnTouches(edits);
   const announcement = describeChangeSetTouches(columnTouches);
   const priorTrackChanges = editor.enableTrackChanges;
+  const priorCurrentUser = editor.currentUser;
   editor.enableTrackChanges = true;
+  editor.currentUser = ASSISTANT_DOCUMENT_AUTHOR;
```
lanthony42: This sits outside the `try`, so a throw from the unguarded `editor.serialize()` in preflight leaves `currentUser` as Robin and relabels manual edits.

lanthony42 (changes requested):

src/assistant/tools/syncfusionDocumentOps.ts
faizalsomani: Confirmed - the assignment is ~280 lines above the `try` and `refresh()` calls `editor.serialize()` in between, so a preflight throw would strand `currentUser` as Robin; fixing the scope and adding a test that injects the throw from that preflight serialize rather than from inside the batch.

lanthony42 (approved):

PR #1729, opened 2026-07-31
feat: connect respondent Box accounts at runtime

anthonyly-feathery:
## Summary
- Adds client-side support for the new `trigger_box_oauth` button action: opens a popup through Box's OAuth + folder-picker flow, then writes the selected folder path and connected account email back into their bound fields
- Generalizes the iOS-only pre-opened-popup workaround to cover the Box OAuth popup on all browsers, since the OAuth call is async and would otherwise break the user-gesture chain `window.open` needs
- Adds a `rawValue` override prop to `TextField` so a bound field's displayed value updates immediately once the OAuth flow completes

Paired with:
- feathery-backend: https://github.com/feathery-org/feathery-backend/pull/3564
- feathery-frontend: https://github.com/feathery-org/feathery-frontend/pull/3708

## Test plan
- [x] `yarn test:ci`, `yarn lint`, `yarn typecheck` all pass
- [x] Manually verified end-to-end against a local backend + hosted-forms-next: connect account, pick folder, submit, confirm delivery

### DIFF
--- a/src/elements/fields/TextField/index.tsx
+++ b/src/elements/fields/TextField/index.tsx
@@ -205,6 +205,7 @@ function TextField({
   setRef = () => {},
   inlineError,
   repeatIndex = null,
+  rawValue: providedRawValue,
   children
 }: any) {
   const [showAutocomplete, setShowAutocomplete] = useState(false);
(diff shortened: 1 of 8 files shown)


src/elements/fields/TextField/index.tsx
```diff
@@ -219,7 +220,9 @@ function TextField({
   const listItemRef = useRef<any[]>([]);
   const inputRef = useRef<{ element?: HTMLInputElement }>(null);
   const { value: fieldVal } = getFieldValue(element);
-  const rawValue = stringifyWithNull(fieldVal);
+  const rawValue = stringifyWithNull(
```
benminor, line 223: I don't understand this sorry can I get an explanation why it's needed? @claude can you explain this?

trebitowski:
Closing as already implemented

PR #1733, opened 2026-08-03
refactor: Clean up assistant auth and request handling

lanthony42:
## Changes

- Flattened the assistant request body, dropping the context envelope
- Send the document freshness hash on the envelope target, replacing `document_state` dirty tracking
- Added timeouts to natively handled assistant tools
- Dropped client-side result clipping, the server digests oversized outputs with full recall
- Moved the docx editing tools into their own subfolder

- [ ] Tested end to end

## Related pull requests

- https://github.com/feathery-org/feathery-backend/pull/3572
- https://github.com/feathery-org/ai-services/pull/650


src/assistant/tools/docx/documentIndex.ts
```diff
@@ -111,44 +97,12 @@ const stateFor = (target: DocumentIndexTarget): ScopeIndexState => {
...
-
-// The instant-staleness half of the contract: called synchronously from the
-// editor's contentChange, before any debounce.
-const markTargetDirty = (target: DocumentIndexTarget): void => {
-  const state = stateFor(target);
-  state.changeSeq++;
-  if (state.dirtySince === null) state.dirtySince = Date.now();
-};
+): string | undefined =>
+  scopeState.get(documentTargetKey(target))?.currentHash ?? undefined;
```
faizalsomani: Returning `undefined` before the first snapshot means the chat sends no `contentHash`, so ai-services skips the freshness gate entirely and answers from a previous session's index - the unconfirmed-session case that used to refuse; return the `pending:` sentinel here instead.

src/Form/logic.ts
```diff
@@ -620,43 +620,26 @@ export const runClientSideLogic = async (
...
-const LOGIC_RULE_FIELD_CHANGE_LIMIT = 100;
-const LOGIC_RULE_KEY_LIMIT = 160;
-const LOGIC_RULE_NAME_LIMIT = 400;
-const LOGIC_RULE_ERROR_LIMIT = 1_000;
-
-const boundedText = (value: unknown, limit: number): string =>
-  String(sanitizeTransportValue(String(value ?? ''), limit).value);
+// A rule is user-authored JavaScript and may return any runtime value.
+// Recursively sanitize every dynamic value before it reaches addToolOutput.
+// No client-side size clipping, the server digests oversized tool results
```
faizalsomani, line 625: queryOutput only digests after the body is already parsed, so an unclipped rule return value over Django's 5 MB `DATA_UPLOAD_MAX_MEMORY_SIZE` 413s every later request in that conversation and the thread is dead for good.

lanthony42: Known tradeoff, `panel_runtime` already unbounded and never faced a form nearly that big. Logic rule return is likely bounded by that

src/assistant/tools/docx/documentIndex.ts
lanthony42: Should be handled - if `undefined`, falls back to `pending:unindexed` in `onContentChange`

faizalsomani: That sentinel only fires on an edit: `documentIndex.ts:430` is inside `onContentChange`, so a fresh load with no edit leaves `currentHash` null until `syncSnapshot` writes it at `:221`, and ais-650 `readAccess.ts:68` then skips the gate entirely, which your `documentIndexFreshness.test.ts:117` asserts.

src/Form/logic.ts
faizalsomani: `panel_runtime` is recomputed per request from form state, while a rule return is arbitrary user JS that lands in the persisted history, and `AssistantChat.tsx:598` re-POSTs that whole history automatically on every tool round trip, so one oversized return 413s the thread for good rather than once.

src/assistant/tools/docx/documentIndex.ts
lanthony42: Used `unindexed` everywhere

src/Form/logic.ts
lanthony42: I'd argue this is on the client since they define the tool and are responsible for pathological output. Added overall truncation as small guard

src/assistant/tools/docx/documentIndex.ts
faizalsomani: Confirmed - the sentinel covers the fresh-load window now. Resolving.

src/Form/logic.ts
faizalsomani: The guard is what matters, responsibility argument aside - confirmed with the tests. Resolving.

faizalsomani (approved):

PR #1737, opened 2026-08-04
refactor: Clean up logic rule tool contracts and dispatch

lanthony42:
## Changes

- Resolved rule tool calls from the model-echoed `ruleId` instead of re-deriving the rule from the tool name
- Raised the rule execution budget to 240s and made timeout errors honest that the work may still complete
- Deleted the dead server-side branch from `runLogicRuleById`, tool rules are client-side only
- Removed the duplicate rule-error warning

- [ ] Tested end to end

## Related pull requests

- https://github.com/feathery-org/ai-services/pull/654

### DIFF
--- a/src/assistant/tools/assistantToolDispatch.ts
+++ b/src/assistant/tools/assistantToolDispatch.ts
@@ -1,20 +1,22 @@
 // Robin assistant tool dispatch for the docx-editor surface. This module is
 // deliberately SyncFusion-free: it never imports the editor. It only calls
-// handlers/objects the host hands in (a docx bridge, a custom-handler map, a
-// local rule execution allowlist, and a runLogicRuleById fn), so @feathery/react stays
-// decoupled from the editor implementation.
+// handlers/objects the host hands in (a docx bridge, a local rule execution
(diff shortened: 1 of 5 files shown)


src/assistant/tools/assistantToolDispatch.ts
```diff
@@ -205,26 +182,31 @@ export async function dispatchAssistantTool(
...
+    if (!run) {
+      return {
+        handled: true,
+        output: syntheticError(
+          'handler_unavailable',
+          'No live form is connected to run rules against.'
+        )
+      };
+    }
+    const ruleId = resolveRuleId(input, ctx.callableRules ?? []);
```
faizalsomani, line 197: Keep a legacy `rule_*` fallback that resolves by `toolName` and forwards the flat input, because this client otherwise returns `unknown_rule` for every call from the currently deployed server until #654 lands.

lanthony42: Deploy ordering issue which is low risk, no real consumers yet, I don't even think Hilb uses logic rule tools yet

faizalsomani (approved):

PR #1740, opened 2026-08-05
feat: live document tokens in the docx editor

trebitowski:
The browser half of live document tokens. A generated Word document opens in the Syncfusion editor with its values bound to the form's fields: editing a token writes through the form's own update path, dependents recalculate, repeated fields grow and shrink their table rows in both directions, and a formula token's result is written back to the field it names.

Pairs with feathery-backend #3589 (produces the content controls) and feathery-frontend #3758 (the per-document toggle).

## Shape

| Module | Responsibility |
| --- | --- |
| `plan.ts` | Parses formulas once, topologically orders them, maps dependents |
| `tokenCycle.ts` | The cycle: read fields, derive, write the document, commit on blur, push formula results back to fields |
| `controls.ts` | The Syncfusion boundary — every content-control call lives here |
| `rows.ts` | Table rows for repeated fields: grow, shrink, renumber |
| `tokenOverlay.ts` | An editor overlay that boxes each token control, aligned across zoom |
| `grammar.ts` / `format.ts` | Twins of the Python modules, sharing the `grammarCases.json` fixture (cross-repo hash) |

The form's field store is the single owner — the document is a view, nothing is duplicated. A formula token additionally mirrors its computed value into the field it names, so `{{ total }}` in a text element and the submission both see it. A field write aimed at a formula token is rejected: the field never feeds the formula, and every pass overwrites it with the derived value. Only formula tokens whose inputs trace back to a real field are written; a memory-only column invents no field.

## Focus while editing

Writing into the document steals focus to Syncfusion's hidden input. The host tracks the last focused form field and, after a field-driven reconcile, hands focus back — so editing a value and tabbing lands on the next field, never the document. A deliberate click into the document is left alone.

## Design notes worth reviewing

**Measured, not read from typings.** This editor's documented behaviour has been wrong repeatedly, so the findings are pinned as tests in `tests/rowMechanics.spec.ts` against a real `DocumentEditor`:

- `insertRow` does not clone content controls — a grown row arrives blank, so its tokens must be built.
- `insertContentControl('Text')` creates a control (the object form no-ops) and discards the selected text, inserting a placeholder.
- `deleteRow` leaves its controls in `contentControlCollection`, so a row count must come from the table widgets, never the token collection.
- The collection is in document order, so a newly created control is not the last entry.
- A control built in the browser survives serialise and reopen.

**Row index is an invariant.** A control's repeat index must equal its index in the field's array; every read and write goes through that number. Deleting a middle row prunes the dead controls and renumbers survivors.

**Overlay geometry across zoom.** Token boxes use Syncfusion's own page transform — `(rect.y − pageGap·(idx+1))·zoom + pageGap·(idx+1)` — where a naive `rect.y·zoom` drifted by `pageGap·(idx+1)·(zoom−1)`.

## Testing

328 tests across the `documentTokens` suite plus the editor-integration specs; `tsc` and eslint clean. Real-editor coverage for grow / shrink / mid-row delete / renumber, the zoom transform, and the formula-to-field write-back. Verified live in a browser against the running stack: adding a line item fills `description` / `qty` / `cost`, the derived amount, and the moved subtotal / tax / total.

## Rebased on latest master

Conflicts in `DocumentEditorContainer.tsx` and `useDocxEditor.tsx` (master's tracked-change review setup, dirty-state tracking, and the editor-registry move) were resolved alongside the token integration.

### DIFF
--- a/docs/document-tokens/assistant-integration-handoff.md
+++ b/docs/document-tokens/assistant-integration-handoff.md
@@ -1,0 +1,431 @@
+# Assistant ↔ document tokens — integration handoff
+
+**Audience:** the engineer wiring Feathery's AI assistant to the live-document-
+token system so it can create and manage tokens on a user's behalf.
+
+**Status:** the token *runtime* (reconcile cycle, formula graph, row sync, the
+Syncfusion write path) is built and covered by tests. A *value* API the
+assistant can call today exists (`setTokenValue`). A structural **add / edit /
+delete token** API for the assistant does **not** yet exist as a public surface
+— the primitives are here, the wrapper is not. This document gives you the
+mental model, the exact data model, the surface that exists, and what you'll
+need to add.
+
+Everything below is grounded in `src/documentTokens/`. File and symbol
+references are exact; line numbers are approximate anchors as of this writing.
+
+---
+
+## 1. Mental model
+
+```
+   field change ──┐
+   document edit ─┼──→ reconcile() ──→ recalc computed values ──→ write only
+   panel edit ────┤                     the tokens whose document text is stale
+   assistant op ──┘
+```
+
+Five things and their relationships:
+
+- **Field** — a form field, owned by the form engine. The single source of
+  truth for a field-backed token's value.
+- **Token** — an appearance in the document (a Syncfusion content control)
+  carrying a `TokenSpec`. Many appearances can share one value.
+- **Value** — the number or string behind a token. A field-backed token's value
+  lives in the field store; a memory token's lives in an in-memory store the
+  cycle owns.
+- **Formula** — an expression on a computed token, recomputed from other values
+  in dependency order.
+- **Format** — how a value is rendered to text (`currency`/`number`/`percent`/
+  `text`) plus optional display transforms (`UPPER`/`LOWER`/`TITLE`/`TRIM`).
+
+**The single-owner rule is the whole design.** There is exactly one owner per
+value. The document is a *view*; the cycle never keeps a second copy of a field
+value. Every propagation bug this replaced came from two stores disagreeing.
+
+### The reconcile cycle
+
+`reconcile()` (`tokenCycle.ts:371`) is the heart. It is **idempotent** — it
+derives everything from current inputs and writes only what the document does
+not already show — so calling it more often than necessary is free and can never
+lose an edit. One pass does, in order:
+
+1. **syncRows** (`tokenCycle.ts:549`) — bring the document's repeat rows in step
+   with the fields' arrays (grow/shrink). Gated so it never runs mid-keystroke.
+2. **derive** (`tokenCycle.ts:265`) — read every input value from its owner,
+   parse numbers, run `recalc` over the plan to compute formula tokens.
+3. **writeValues** (`controls.ts:504`) — for every token present in the
+   document, compare its expected text against what the control shows and write
+   only the differences, as one undo step.
+
+**Sync directions:**
+
+- **field → document:** a field change re-renders the host, which calls
+  `reconcile()`; derived values recompute and stale controls are rewritten.
+- **document → field:** a reader edits a token; on blur/Enter the cycle reads
+  the control's text and writes it to the owner (`setTokenValue`).
+- **document → field on undo:** during Syncfusion history replay the cycle
+  *adopts* the restored text into the fields (see §5, `adoptFromDocument`).
+
+**Seeding / precedence:** on the first read of a token this attach,
+`seedDefault` (`tokenCycle.ts:229`) writes the authored `default` into the owner
+**once** (`seededDefaults` set guards it). `DEFAULT_WINS = true`
+(`tokenCycle.ts:118`) means the default is written even over an existing value on
+open; after that first seed, the field wins and a cleared value stays cleared.
+
+---
+
+## 2. Data model
+
+### `TokenSpec` (`plan.ts:33`)
+
+The exact shape carried by every token, both in memory and on disk:
+
+```ts
+type TokenSpec = {
+  id: string;                 // the authored name; a field key when one matches
+  index?: number | null;      // row of a repeated field; absent/null for a scalar
+  source?: string | null;     // field key when field-backed; absent for memory tokens
+  formula?: string | null;    // expression for computed tokens; absent for inputs
+  display?: string | null;    // a display transform, e.g. "UPPER(note)"
+  reads?: string[] | null;    // field keys a formula reads that have no token of their own
+  instance?: string;          // address of THIS appearance (a token may appear many times)
+  default?: string | number;  // author fallback; only meaningful for a field-backed token
+  format?: TokenFormat;       // { kind: 'currency'|'number'|'percent'|'text'; decimals?: number }
+  validate?: TokenValidation; // { min?: number; max?: number; required?: boolean }
+};
+```
+
+Two derived keys you will use constantly (`plan.ts:66`, `plan.ts:72`):
+
+- **`valueKey(spec)`** identifies a **value**: `id` for a scalar, `` `${id}__${index}` ``
+  for a repeated row (e.g. `qty__2`). One value key, shared by every appearance.
+- **`instanceKey(spec)`** identifies an **appearance**: `spec.instance` if set,
+  else the value key. A token used twice in one row carries a `#N` suffix
+  (e.g. `qty__2#1`) so both controls stay separately addressable.
+
+### The on-disk content-control tag (`controls.ts`)
+
+A token is identified, read, and written entirely through its Syncfusion content
+control's **tag**. The tag is:
+
+```
+ftk:<compact JSON of the TokenSpec>
+```
+
+- `TAG_PREFIX = 'ftk:'` (`controls.ts:14`). Any control whose tag starts with
+  `ftk:` is ours; everything else is left untouched.
+- `encodeTag(spec)` = `` `ftk:${JSON.stringify(spec)}` `` (`controls.ts:93`).
+- `decodeTag(tag)` parses it back, returning `null` for a foreign or malformed
+  tag, or when the decoded object has no string `id` (`controls.ts:96`).
+
+A real tag (from `tokenSpecCases.json`, a repeated numeric field):
+
+```
+ftk:{"id":"qty","format":{"kind":"number"},"instance":"qty__2","index":2,"source":"qty"}
+```
+
+A derived scalar's second appearance:
+
+```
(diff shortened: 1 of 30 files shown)


trebitowski:
## Regression in the row-delete fix — do not merge

Reported: delete a row, then add one so the new row reuses the freed index -> the new row's tokens are unlinked and every cell shows the placeholder. Subtotal excludes the row.

Does **not** reproduce against a real `DocumentEditor` in jsdom (the regrow links and values correctly), so this is real-editor-specific. Notably, Add Item worked live *before* the delete fix landed — only the empty text token showed a placeholder — so this is a regression introduced by that fix, not a pre-existing gap.

**Prime suspect: `pruneDeleted` in `rows.ts`.** It splices Syncfusion's internal `contentControlCollection` in place. That same array is maintained by `insertContentControlInCollection` / `pushContentControlByOrder` when a control is created, so mutating it likely leaves the editor's bookkeeping inconsistent and controls created afterwards never register.

Pruning was added because `liveTokens` filtered by **address**, and renumbering a survivor onto a dead control's address made both disappear. The safer fix is to stop keying liveness on addresses at all — build the token list from the collection with control identity preserved, rather than from `exportContentControlData()` — and leave Syncfusion's array alone.

Until that is done, the mid-row delete path should be considered unsafe.

trebitowski:
## Root cause fixed; one downstream bug remains

**Fixed (61a752cf).** The regression came from `pruneDeleted` splicing Syncfusion's own `contentControlCollection`, which broke every control created afterwards. That whole approach is gone.

Liveness now lives in `controls.ts` — the Syncfusion boundary — and works by **identity**:

- `readTokens` walks the live control collection, decodes each tag, and drops any control whose row widget is no longer in its table (`table.childWidgets.indexOf(row) === -1`). A *stored* rowIndex goes stale after a structural edit, so identity is the only truthful test.
- Values come from `getResultContentControlText(control)` — read off the control, not looked up by address — with a positional fallback for hosts that cannot report it.
- Nothing mutates the editor's state.

This is what made pruning look necessary: a renumbered survivor can carry the same *address* as a deleted control, so any address-based filter loses both. Identity keeps them distinct. Confirmed on a real editor: after a mid-row delete the collection still holds the dead entries untouched, and the read returns exactly the survivors.

Also fixed while here: column position was taken from the stored `cell.cellIndex`, which goes stale the same way; it is now derived by identity too.

## Still broken, and marked in the suite

Growing a row **after** a deletion creates only the **first** cell's control. Once the first new control is tagged the caret is still inside it, and `insertContentControl` bails when the selection is inside an existing content control — so the second cell is skipped. The caret has to be moved clear before the next insert.

Left as an `it.skip` in `rows.spec.ts` with the cause written out, so the suite is green rather than falsely so and the bug is recorded where the next person will see it. **Delete-then-add is still not safe to ship.**

trebitowski:
## Second attempt at the grow-after-delete bug: three hypotheses eliminated, still unfixed

The stability fix in 61a752cf stands (identity-based liveness, no mutation of the editor's collection). The remaining bug — growing a row after a deletion creates only **one** cell's control — I could not fix. Recording what was ruled out so the next attempt does not repeat it:

| Hypothesis | Result |
| --- | --- |
| Caret left inside the control just built, so `insertContentControl` bails | **No.** Adding `moveToDocumentStart` between inserts changed nothing |
| Cell order — building left-to-right walks into the previous control | **No.** Reversing it merely swapped *which* cell fails: `amount__2` got created, `qty__2` did not |
| `group.table` going stale, since inserting a control relayouts the table | **No.** Re-resolving the table and row from the newly built control's placement changed nothing |

The invariant across all three: **whichever cell is processed second fails**, regardless of direction, caret position, or how the row is addressed. Both cells *can* be built individually, so it is not the cell or the spec.

That points somewhere I ran out of room to check — most likely the `existing` Set diff used to find the newly created control:

```ts
const existing = new Set(controlsOf(editor));
insertContentControl('Text');
const built = controlsOf(editor).find((c) => !existing.has(c));
```

If inserting a control **replaces** `contentControlCollection` with a new array rather than mutating it, then on the second pass no element is in `existing`, `find` returns whichever control is first in document order, and we retag an existing control instead of the new one. That would explain a token going missing while another appears with the wrong value. Worth logging the collection identity and length either side of the second insert before assuming anything.

Tree is at 61a752cf, 222 passing + 1 skipped, tsc and eslint clean. The failing case stays as an `it.skip` with the cause written out. **Delete-then-add remains unsafe to ship.**

trebitowski:
## Instrumentation added (4230ba3d) — and it immediately named the bug

Every failure path in `rows.ts` was a bare `continue`. A row that built half its controls was indistinguishable from one that built all of them, because the tokens that ARE there work — which is why this took three wrong hypotheses to chase and still is not fixed.

`growGroup` and `shrinkGroup` now take a `ProblemReporter` and speak up for: no template row, an unreachable row or column, the editor creating no control, a row that built fewer tokens than its template, and a shrink that did not reach its target. `syncRows` routes them to `console.warn('[feathery] document rows: …')`.

On the first run against the failing case:

```
[feathery] document rows: the editor created no control for amount at row 2 column 1
[feathery] document rows: row 2 built 1 of 2 tokens — the rest show a placeholder and stay unlinked
```

That is the whole bug in two lines, and it narrows it past all four earlier hypotheses: **the second `insertContentControl` in a row is a genuine no-op** — not a lookup failure, not stale widgets, not the caret, not cell order. I also tried finding the fresh control by "the one with no tag" instead of by set difference; same result, so the control is never created. That change was **reverted** — a foreign untagged content control in the document would have been retagged as ours, which is worse than the bug.

## What the next attempt should do

Measure why the second call refuses, in a row that already holds one control. Everything else is ruled out and recorded. Two things worth trying first: insert both controls **before** tagging either (a fresh control is a "plain" content control, and the documented bail is on the selection being inside one), and check whether the refusal depends on the two controls sharing a table row rather than a paragraph.

Tree at 4230ba3d: 222 passing + 1 skipped, tsc and eslint clean. Delete-then-add still unsafe.

trebitowski:
## Value invariants added to the fuzz harness (4b2055c0) — found and fixed a real bug

The harness asserted `shapeViolations` only: structure. It could not catch a token showing the wrong value, which is the failure a reader actually notices.

It now also checks, after every gesture, that each **derived** token's displayed text equals what it computes to. Derived is the unambiguous case — nobody types into one, so any disagreement is a defect rather than a mid-edit state.

It failed immediately, on two independent seeds:

```
seed 1  step 10 (clear a value): item_total__0 shows "", computes "$60.00"
seed 99 step  6 (clear a value): item_total__0 shows "", computes "$60.00"
```

**Cause.** `reconcile` exempts the token under the caret so a value is not rewritten under someone mid-word. That exemption was applied to computed tokens too — but nobody types into a derived token, its contents are locked, so exempting it just left whatever the caret did to it standing, including empty. A cleared derived token stayed blank until blur.

**Fix.** A computed token is never exempt. Small, and the fuzz confirms it across all seeds.

Worth noting this bug was reachable by ordinary gestures — clear a derived token's value and it stays blank — and had survived every earlier pass, including the browser session, because structural invariants are blind to values.

222 passing + 1 skipped, tsc and eslint clean.

trebitowski:
## Attempt on the grow-after-delete bug: partial progress, reverted

Acted on the measured hypothesis (the cached `group.table` being superseded by the delete). Re-derived the table from a live control on every step instead. Two genuine findings:

1. **The lookup was matching the DEAD control.** After a renumber a survivor carries the same address as the control it replaced, and the dead one comes first in the collection — so resolving "the row to copy" returned no row at all. Skipping detached controls fixed that, and `qty__2` then got created where previously nothing did.
2. **The second cell still creates no control**, even with a freshly derived table and a fresh-control lookup scoped to the target row. So the stale-table hypothesis is only part of it.

**Reverted**: the change regressed 9 previously passing tests — the re-derivation broke the ordinary grow path (no delete involved). Tree is back at 4b2055c0, 222 passing + 1 skipped.

### What is now known

- `insertContentControl` creates a control for BOTH cells in isolation (measured; count 4->5->6), with an empty selection outside any control, locks set or not.
- It creates only the first when driven from `growGroup` after a delete.
- The dead-control-shadows-the-survivor problem is real and worth fixing on its own, independent of this bug.
- Seven hypotheses eliminated: caret position, cell order, control locks, set-difference lookup, untagged lookup, stale table reference, dead-control shadowing.

The remaining difference between the working measurement and the failing path is **how the cell is addressed**: the measurement resolves the table from `pages[0].bodyWidgets[0].childWidgets[0]`, `growGroup` from the control's `ownerTable` chain. Those may not be the same widget. That is the next thing to check, and it is one comparison rather than another fix.

Delete-then-add remains unsafe. The `it.skip` carries the cause.

trebitowski:
## The five-minute measurement: three more hypotheses eliminated, cause narrowed to growGroup's own code

Ran the comparison. Results, all against a real `DocumentEditor`:

| Question | Answer |
| --- | --- |
| Is the table from `pages[0]...childWidgets[0]` the same object as the one from a control's `ownerTable` chain, after a delete? | **Yes** — same object, same row count, same cell counts |
| Does `insertRow` replace the table or row objects? | **No** — `captured === live`, and `captured.childWidgets[3] === live.childWidgets[3]` |
| Does the two-cell insert work *after* a delete when driven by hand? | **Yes** — control count 6 -> 7 -> 8, both cells |

So the delete is **not** implicated, the cached `group.table` is **not** stale, and the widget identities are fine. The earlier conclusion in my previous comment — that the delete supersedes the table reference — is **wrong**, and the `liveRowOf` change built on it regressed 9 tests and was reverted.

### What that leaves

The by-hand sequence and `growGroup` now differ in exactly one respect: **how the freshly created control is found.**

- by hand (works): `controlsOf(editor).find((c) => !c.contentControlProperties.tag)`
- `growGroup` (fails): set difference against a snapshot taken before the insert

`growGroup` reports "the editor created no control", which is the `!built` branch — so the set difference is finding nothing while the collection demonstrably grows. Either the collection array is replaced (making a Set of old references useless) or the insert is landing somewhere the snapshot already covered.

**The fix is almost certainly to replace that lookup**, scoped to the target row so a foreign untagged control can never be retagged as ours. I attempted it twice, but both times bundled with the `liveRowOf` change that caused the regressions, so neither attempt was a clean test of it. It needs to be tried on its own.

Eliminated so far: caret position, cell order, control locks, stale table reference, stale row/cell indexes, the delete itself, dead-control shadowing (real, fixed as part of the attempt), table-reference provenance, and `insertRow` replacing widgets.

Tree at 4b2055c0, 222 passing + 1 skipped, clean.

PR #1744, opened 2026-08-05
feat: atomic section authoring, delta indexing, and document-edit hardening

faizalsomani:
# PR #1744 — feat: atomic section authoring, delta indexing, and document-edit hardening

## What this PR achieves

Robin Assist can author complete, correctly formatted document content - new sections, subsections, and tables that visually match their siblings - as one reviewable tracked-change group, without the editor viewport jumping or the page appearing to refresh. Document indexing moves from full re-posts to deltas, cutting steady-state payloads from multi-MB to tens of KB. The custom tracked-change review experience is scoped strictly to assistant-enabled editable hosts; every other host keeps native Syncfusion behavior.

## How it does it

- **Atomic structured authoring**: new `applyDocumentBlocks`/`getSectionPattern` capabilities let the engine compose heading, prose, and populated tables in one change set; ordered change-set simulation and unified text verification run before any write, and source-format inheritance (fonts, table borders, shading, banding, widths, indents, margins, alignment) is derived from the sibling family in the document - nothing hardcoded.
- **Transactional appearance**: table/row/cell formatting participates in change groups and is restored on rejection, so accept/reject round-trips leave no untracked damage.
- **Delta document indexing**: content-hash diffs against the last confirmed snapshot with anchor remaps and a 409 full-snapshot rebase fallback.
- **Viewport stability**: resize goes through a view-preserving editor resize (no cursor homing, no container-level relayout), same-size ticks are skipped, and Chromium scroll anchoring is disabled on the editor scroller.
- **Blast-radius containment**: one predicate (`assistantEnabled && !readOnly`), derived once, gates the review rail, revision-group rebinding, and all review customizations; shared revision primitives live in a neutral `documentEditorPrimitives` module so shared components no longer import the assistant engine.
- **Cache prefix stability**: the client no longer rewrites already-completed tool results in settled history. That rewrite was changing the same tool call's bytes deep in the conversation and collapsing the provider's cached prefix; the service is now the single owner of reducing settled history, so the prefix only grows.
- **Incident-derived coverage**: captured live failures replay as regression suites against a real Syncfusion editor.

**Shared viewport changes (deliberate, all hosts).** Two viewport changes in this PR are deliberately shared across ALL `DocxEditor` hosts rather than gated behind the assistant predicate. Every host's geometry tick now goes through `resizeDocxEditor`, which resizes the inner editor inside `preserveDocumentViewDuring` (with a status-bar zoom refresh, replacing the container-level `resize()`), and skips ticks whose size did not change; separately, `overflowAnchor: 'none'` is set on Syncfusion's private scroller for every host. Both fix the same host-independent bug: on a resize or a tracked-edit page rebuild, Syncfusion's container-level relayout and Chromium's generic scroll anchoring each move the user's viewport (the "document jumps to the top / drifts on edit" reports), and a read-only viewer is yanked exactly the way an assistant host is. Gating these would keep a known viewport bug alive on non-assistant hosts, so the fix is intentionally host-wide; `useDocxEditor.spec.tsx` pins the non-assistant resize behavior (single editor resize, `FitPageWidth` refit, zoom refresh, no container resize, no cursor homing, review state untouched).

**Deliberate behavior removal.** Editable non-assistant hosts whose documents carry tracked changes - including external consumers of the public `DocxEditor`/`StyledContainer` exports (dashboard/builder) - revert from the #1725 grouped review rail back to Syncfusion's native review surface (auto-opening Changes pane, per-revision resolution) unless the host passes the assistant-enabled flag.

## Reordering as a structure change, not a rewrite

The assistant could read a document's structure but had no way to *rearrange* it, so the model
expressed structural intent as a sequence of content writes. Asked to move a section it retyped the
whole thing and left a duplicate behind when one op of its cleanup group failed; asked to swap two
subsections it improvised a three-way text shuffle through `__TMP_SWAP_HEADING_1__` /
`__TMP_SWAP_PARA_1__` placeholder tokens and failed on an `expect` truncated by one character. Both
are the same class: every content write authored to express a *structural* intent is a fresh chance
to mis-transcribe, mis-count an offset, or invent a value.

The engine now has **one relocation primitive** with **three advertised ops** over it, and none of
them carries a content field, so that class is unreachable through them:

- `move_section { anchor, targetAnchor, position }`
- `swap_sections { anchor, otherAnchor }` - the primitive twice, bottom-up, inside one op, so the
  exchange cannot half-apply
- `copy_section { anchor, targetAnchor, position }` - the primitive with its delete step skipped

The mechanism is SyncFusion's own and needed nothing new: select the range, take `selection.sfdt`,
paste at the target, delete the source. Because it runs inside `applyDocumentEdits` it inherits
forced track changes, single-group revision stamping, atomic rollback and viewport stability - so a
relocation is **one rail card, one accept, one reject, one undo**, and the moved content keeps its
tables, shading, column widths, `allowAutoFit`, images and subsections. Both anchors name section
*units*: moving a parent carries its subsections, moving a subsection leaves its parent behind, and
`position: 'after'` means after everything the target section covers.

SyncFusion has no producer for the `MoveTo`/`MoveFrom` revision types, so a move appears in the rail
as an insertion plus a deletion under one card - exactly what Word shows with "track moves" off. A
fidelity limit, not a correctness one.

## Section reads no longer truncate at the first subsection

`getDocumentInventory` with `scope: 'section'` stopped at the next heading of *any* level, so a
parent-section read returned only the blocks above its own first subsection - under-reporting most of
this proposal. The depth-aware rule (stop at the next heading at the same level or shallower) already
existed elsewhere in the file; it now has a single owner shared by both the inventory read and the
relocation ranges, so what a section read reports is exactly what a relocation moves. The
mark-inclusive range end - which must consume the section's trailing paragraph mark, and must not
consume a Word section break - likewise has one owner, shared with `delete_paragraph`.

## No write may carry placeholder text

One preflight pass over the post-expansion `edits` array refuses placeholder **shapes** -
`__DELIMITED_TOKENS__`, `{{...}}`, `<<...>>`, `%%...%%`, dollar-brace markers, and standalone
`PLACEHOLDER` / `TODO` / `TBD` / `LOREM IPSUM` / `XXX` - on every model-authored write field,
including composed section-table cells. Matching the shape rather than the tokens we happened to
observe refuses the class, not the one incident. It never runs on `find` or `expect`, which are reads,
so cleaning an existing placeholder up still works. A registry-enumerating test derives the guarded
field set from the capability registry itself and asserts every op is refused before any write, so an
op added later cannot bypass it.

## A resolved card no longer leaves a paragraph wearing the wrong style

SyncFusion has no Formatting revision type, so a paragraph's *style* is never part of what accepting
or rejecting a card resolves. That stays invisible until the resolution merges two paragraphs, and
then the survivor keeps the removed paragraph's format: the content comes back exactly right and a
paragraph is silently restyled, with no revision to explain it.

**This was a live defect on `insert_section`, which is already in this PR** - insert a section before
a body paragraph, reject the card, and that paragraph comes back as a heading. The relocation work
only made it easy to see, because moving a subsection above a top-level section puts two different
styles either side of one paste. It is also symmetric: accepting a tracked delete merges its last
mark into the following paragraph and restyles that one the same way, which is the worse half because
accept is the happy path.

Fixed once, as a paragraph-style inverse on the revision group itself, beside the appearance inverse
that already rides there - one owner, every op, no second mechanism. Two differences from that
neighbour, both forced by what the defect is: it runs *after* the content resolves, because the merge
that loses the style is the resolution; and it runs for accept as well as reject. Identity, not
position, decides what gets written - a restore applies only to a paragraph that still reads exactly
as it did when captured, so it cannot touch a paragraph the change set rewrote and cannot restyle
whatever occupies an anchor after the indices move. A paragraph already reading correctly is left
alone, so a style someone set deliberately is never overwritten.

## Four refusals, each naming what to do instead

Integrity lives in the engine, not in prompt guidance:

1. **The document's last section when the document ends in a table.** SyncFusion's `acceptAll` throws
   inside its own delete there, so the edit would apply and then break the review pane - refusing
   beats crashing. The message points at reaching the same order from the other side.
2. **A target inside the source's own range.** Checked on both the anchor the model named *and* the
   caret it resolves to, which are not the same test.
3. **A target inside a table cell.** A section is relocated between top-level blocks, never into a
   cell.
4. **A source holding a pending change by another author.** A move folds what it moves into its own
   card, so rejecting it would silently revert a human reviewer's edit. Robin's own pending edits fold
   in safely and reject restores the truth.

A copy deliberately gets neither source-side refusal: it deletes nothing, so there is no
document-tail delete to crash and nobody's pending edit is at risk. That the tail-table case does not
throw for a copy is asserted rather than assumed. Swapping a section with its own subsection is
refused as having nothing to exchange, and the relocation verifies its own re-resolved source is
outside the run it just pasted, so deleting the copy instead of the original is impossible rather
than merely unlikely.

## Verified live, not just green

Driven through the real Robin chat against the 22-page proposal: both captain scenarios (move a
section above another; swap two subsections), a cross-level move accepted and rejected,
`insert_section`'s reject, `copy_section` end to end, and the two cases jsdom could not settle - a
source containing a **page break** and a source containing an **image**. Both work; the jsdom hang was
a jsdom layout artifact, so the design report's rows 9 and 11 are supported. Every case produced
exactly one rail card. The swap prompt is the one that previously wrote placeholder tokens; the
document now contains none.

> Rejecting a card restores content, order, styles and formatting, and leaves both rendering and
> future-typing behaviour unchanged; on a real document the serialized bytes may still differ where
> SyncFusion normalizes inline runs or a paragraph mark, which was measured not to affect either.

That last clause was measured rather than assumed: typing at the end of the affected paragraph
produces the same character format before and after a move plus reject, and across the whole document
exactly one paragraph mark's serialized format changed, on a paragraph that has text - where new
typing inherits from the preceding run rather than the mark.

## Two corrections to the design report

- Row 6 is wrong for one sub-shape. A source range that ends at the *first block of the next Word
  section* puts the section break inside the range, and SyncFusion authors no rejectable revision for
  deleting one - the engine caught that itself as `untracked_write`. Ranges now stop at the Word
  section boundary instead, and reject is byte-identical.
- Accepting a move whose source was a Word section of its own collapses that now-empty section rather
  than stranding a blank page. Page setup on what remains is untouched: a relocation carries content,
  not page geometry.

## File-by-file changes

- `src/Form/grid/Container/index.tsx` — forwards the assistant-enabled host flag into styled field rendering so shared document editors can keep native behavior when the assistant is off.
- `src/Form/grid/StyledContainer/index.tsx` — passes the assistant-enabled host flag to `DocxEditor` so every custom review install is controlled by one explicit host predicate.
- `src/assistant/AssistantChat.tsx` — delegates outbound-history preparation, streamed-message coalescing, and scoped scrolling to focused helpers so transport and rendering policy no longer accumulate in the component.
- `src/assistant/capabilities/registry.ts` — declares the section-composition, table-appearance, grounded-cell-write, and document-read capability shapes so client dispatch advertises only supported editor operations.
- `src/assistant/capabilities/tests/registry.spec.ts` — asserts the expanded document capability catalog, parameter shapes, and read-operation list so catalog drift fails in tests.
- `src/assistant/chatScroll.spec.ts` — covers container-scoped chat scrolling so assistant output cannot move the editor or page viewport.
- `src/assistant/chatScroll.ts` — extracts the existing scroll-to-bottom assignment (`scrollTop = scrollHeight`, already container-scoped on master) into a named, tested helper; a relocation for reuse and coverage, not a behavior change.
- `src/assistant/messageHistory.spec.ts` — covers old-tool-output digesting, round-pinned selection, current-turn preservation, and unchanged pending tool parts so service-owned dangling-call repair cannot regress back into the client.
- `src/assistant/messageHistory.ts` — digests historical document-tool output and pins the request-time selection to its user round while leaving dangling-tool-call repair at the service boundary; the request preparers are module-private behind a `__testing` seam.
- `src/assistant/messageRendering.spec.ts` — covers snapshot-versus-append continuation coalescing so repeated streamed assistant parts stay fixed.
- `src/assistant/messageRendering.ts` — coalesces render-only assistant continuation snapshots without mutating transport history, preventing duplicated streamed content.
- `src/assistant/tools/assistantToolDispatch.ts` — routes structured block authoring and section-pattern reads through the existing editor bridge so the new tools share established timeout and handler boundaries.
- `src/assistant/tools/docx/documentIndex.ts` — computes serialized document-index deltas with changed blocks, removals, anchor remaps, pending-sync ordering, and safe full-snapshot fallbacks to reduce steady-state payloads without weakening consistency.
- `src/assistant/tools/docx/docxEditorBridge.ts` — exposes section-pattern derivation through the document bridge so tool dispatch does not reach into Syncfusion internals directly.
- `src/assistant/tools/docx/numericCells.ts` — centralizes zero-padded-integer detection so identifier-looking cell text is not reformatted as a number.
- `src/assistant/tools/docx/syncfusionDocumentOps.ts` — implements atomic section/table mutations, ordered change-set simulation, unified text verification, source-format inheritance, transactional appearance, indexing, and grouped review integration on top of the neutral revision primitives.
- `src/assistant/tools/docx/tableAppearance.ts` — models, reads, compares, applies, and restores cell, row, border, shading, and banding appearance so non-revision formatting participates transactionally in change groups.
- `src/assistant/tools/docx/tests/acceptGroupsWithTableFormat.spec.ts` — exercises grouped acceptance and rejection when table-format side effects are present so revision resolution and appearance replay stay atomic.
- `src/assistant/tools/docx/tests/cellMoneyFormat.spec.ts` — exercises source-grounded money formatting and zero-padded identifier preservation so numeric cell writes do not invent or corrupt presentation.
- `src/assistant/tools/docx/tests/documentIndex.spec.tsx` — covers delta computation, repeated-text remapping, serialization, behavioral fallback thresholds, accepted-change sync, the pinned server hash vector, and the 409 full-snapshot rebase so the client's delta obligations stay fixed.
- `src/assistant/tools/docx/tests/docxEditorBridge.spec.ts` — covers section-pattern bridge exposure so the document read operation is proven to reach the live editor adapter.
- `src/assistant/tools/docx/tests/fixtures/incidentPayloads.ts` — records minimized payloads from the 2026-08-03 live editing incidents, each documented by what went wrong, so regression suites replay the actual wire shapes instead of synthetic approximations.
- `src/assistant/tools/docx/tests/formatAfterInsertDrift.spec.ts` — covers inherited formatting after structural insertion so new content cannot silently drift from its source section.
- `src/assistant/tools/docx/tests/incidentReplays.spec.ts` — replays captured live editing incidents end to end so field-reproduced failures remain fixed across engine changes.
- `src/assistant/tools/docx/tests/opContracts.spec.ts` — asserts per-operation preconditions, result shapes, capability parity, and safety contracts so the engine cannot diverge from the advertised tool catalog.
- `src/assistant/tools/docx/tests/paragraphSplitOracle.spec.ts` — drives the exact selection-based and dependent-anchor paragraph-split wire strategies through real Syncfusion behavior so ordered preflight, newline normalization, grouping, and format inheritance are locked together.
- `src/assistant/tools/docx/tests/rowStaleRepro.spec.ts` — tracks the normalized change-set failure result for the stale-row reproduction so the original regression guard matches the current error contract.
- `src/assistant/tools/docx/tests/sectionComposer.spec.ts` — covers atomic heading, prose, and populated-table composition so a generated section lands as one reviewable change group with stable placement and formatting.
- `src/assistant/tools/docx/tests/sectionPattern.spec.ts` — covers sibling-family structure and style derivation so header rows, unrelated sections, and sparse examples cannot contaminate inferred patterns.
- `src/assistant/tools/docx/tests/selectionRewrite.spec.ts` — covers selection-sourced rewrites and formatting across paragraph splits so request-time selection integrity survives structural text edits.
- `src/assistant/tools/docx/tests/syncfusionDocumentOps.spec.ts` — provides the main integrated engine regression suite for mutation safety, ordered verification, grouping, appearance, inheritance, indexing, and view preservation.
- `src/assistant/tools/docx/tests/tableFormatOps.spec.ts` — covers table appearance reads, writes, copying, restriping, topology preservation, and restores so format operations round-trip without untracked damage.
- `src/assistant/tools/tests/assistantToolDispatch.spec.ts` — covers structured authoring routes, section-pattern reads, aliases, and timeout handling so every new dispatch path is explicit and bounded.
- `src/assistant/tools/tests/handleAssistantToolCall.spec.ts` — recognizes the expanded document tool names so the call-handler contract stays aligned with dispatch.
- `src/elements/components/DocxEditor/DocumentEditorContainer.spec.tsx` — proves revision-group rebinding occurs only for assistant-enabled editable hosts and that read-only or assistant-off hosts keep native behavior.
- `src/elements/components/DocxEditor/DocumentEditorContainer.tsx` — rebinds persisted review groups on document readiness only behind the review gate, removing redundant isolation installation from the shared container.
- `src/elements/components/DocxEditor/TrackedChangeGroups.spec.tsx` — covers rail grouping, stepping, scoped reveal, event containment, editor replacement, and atomic resolution so the custom review surface fails safely.
- `src/elements/components/DocxEditor/TrackedChangeGroups/GroupCard.tsx` — renders typed group actions against neutral revision contracts so review presentation stays independent of the assistant engine.
- `src/elements/components/DocxEditor/TrackedChangeGroups/index.tsx` — implements grouped review navigation and resolution with fault containment at exactly the React/editor event entry points and the rail error boundary; reads and resolution calls beneath them throw normally.
- `src/elements/components/DocxEditor/TrackedChangeGroups/types.ts` — defines rail group and action contracts using neutral revision primitives so shared UI has no static dependency on the large assistant mutation engine.
- `src/elements/components/DocxEditor/index.tsx` — mounts and contains the custom review rail only for assistant-enabled editable hosts so gated-off hosts remain on the fully native Syncfusion surface.
- `src/elements/components/DocxEditor/useDocxEditor.spec.tsx` — proves non-assistant host resize updates editor and status-bar chrome without cursor homing, container refresh, or assistant review mutation.
- `src/elements/components/DocxEditor/useDocxEditor.tsx` — gates renderer patching, native Changes-pane suppression, revision isolation, and resize interception behind the review predicate, preserves the view on resize, and defers review-gate editor recreation until any in-flight create/open settles.
- `src/utils/documentEditorPrimitives.ts` — owns neutral editor/revision types, view preservation, group tagging, listing, resolution, rebinding, and appearance replay so shared DocxEditor components no longer statically import the 12k-line assistant engine.
- `src/assistant/tools/docx/tests/relocateSection.spec.ts` — covers move, swap and copy over the shared relocation primitive: block-sequence position arithmetic over the measured paste delta, the four refusals, subsection depth, and byte-exact reject, including the live failure shape that motivated the fix.

## Validation and rollout

- Full local-stack validation against the production-shaped 94-page proposal: react 97 Jest suites / 1,603 tests, typecheck, lint, node+UMD builds; ai-services 8,680 tests (pure + real-Postgres db suites); backend end-to-end proxy regression. In-editor behavior (section authoring, table creation with full sibling-format inheritance, tracked-change review, viewport stability) verified live in the browser at the exact shipped commits, plus a round of end-to-end product testing on the assembled stack.
- An independent defensive review of the final branches ran before this PR left draft; every inline comment carries a reply naming the addressing commit or the reasoning for declining, and addressed threads are resolved.
- **Deploy order across the three linked PRs** (feathery-react #1744, ai-services #657, feathery-backend #3591): deploy ai-services and feathery-backend first (either order between them is safe), feathery-react last and promptly. No ordering crashes or corrupts data; the two skew windows degrade gracefully (document-index delta posts 400 and fall back to the last snapshot until the react deploy; an old dashboard bundle sees unknown-tool errors on the six new ops until it updates).


src/elements/components/DocxEditor/useDocxEditor.tsx
```diff
@@ -216,6 +598,17 @@ export function useDocxEditor({
         // Native right-click menu — insert/delete table rows & columns,
         // cut/copy/paste, etc. (the built-in toolbar is disabled).
         ed.enableContextMenu = true;
+        try {
+          // TrackedChangeGroups is the review surface: mark the native
+          // Changes pane user-closed (its ✕'s own switch) so its auto-open
+          // never fires and covers the panel.
+          ed.showRevisions = false;
```
faizalsomani: [question] This block (showRevisions=false, commentReviewPane.isUserClosed=true, installRevisionGroupIsolation, installRevisionHighlightRendering) runs at create for EVERY DocxEditor mount, but TrackedChangeGroups only mounts when !readOnly. A readOnly host showing a document with pending revisions loses the native Changes pane and gets the custom rendering with no review surface at all. Non-assistant hosts also get the isRevisionMatched/compareTwoRevisions merge-gating unconditionally. Should installation be gated on the same condition that mounts the rail (or on the tracked-change-groups feature), and what is the intended readOnly-with-revisions experience?

src/elements/components/DocxEditor/useDocxEditor.tsx
```diff
@@ -1,6 +1,11 @@
 import { useCallback, useEffect, useRef, useState } from 'react';
 import { featheryDoc, featheryWindow } from '../../../utils/browser';
 import { dynamicImport } from '../../../integrations/utils';
+import {
+  findReplaceCounterpart,
+  installRevisionGroupIsolation,
```
faizalsomani, line 6: [shrink] Shared editor infrastructure (useDocxEditor, DocumentEditorContainer, TrackedChangeGroups) now statically imports from src/assistant/tools/docx/syncfusionDocumentOps - a ~12k-line assistant edit-engine module - so every form bundle with a docx field carries the dependency even with the assistant disabled, and the shared component layer now points into src/assistant. Extract the revision-group primitives these components need (installRevisionGroupIsolation, rebindRevisionGroups, listRevisionGroups, resolve*AsOneUndo, findReplaceCounterpart, preserveDocumentViewDuring) into a neutral shared docx module that both the editor components and the assistant engine import.

src/elements/components/DocxEditor/useDocxEditor.tsx
```diff
@@ -308,7 +701,34 @@ export function useDocxEditor({
     return editor.saveAsBlob('Docx');
   }, [editor]);
 
-  const resize = useCallback(() => containerInstRef.current?.resize?.(), []);
+  const resizeEditor = useCallback(
```
faizalsomani, line 972: [question] resizeEditor swaps DocumentEditorContainer#resize for editor.resize() on every host, and the ResizeObserver now skips same-size ticks. editor.resize() relays out the page but does not run the container's refreshLayout, which also owned toolbar/status-bar/pane geometry - has a plain (non-assistant) host been checked for stale container chrome after a window resize, and is there any host that relied on a same-size resize() call to recover layout (e.g. after late font load)?

faizalsomani (commented):
Defensive review - inline comments below.

src/elements/components/DocxEditor/DocumentEditorContainer.tsx
```diff
@@ -335,9 +339,34 @@ export default function DocumentEditorContainer({
         documentId: activeDocumentId,
         envelopeId: envelope?.id
       });
+      // Group rebinding deliberately does NOT happen here: this fires from
+      // SyncFusion's `created` callback, before the source document is opened,
+      // so there are no persisted revisions to bind yet. See onDocumentReady.
+      // Isolation is document-independent, so installing it once at create is
+      // both correct and cheapest.
+      try {
+        installRevisionGroupIsolation(editor);
```
faizalsomani: [shrink] Redundant install: useDocxEditor already calls installRevisionGroupIsolation(ed) in its own created path for every editor this container mounts, and the installer is idempotent-guarded. Two call sites for one invariant is the pattern principle 8 warns about - drop this one and let the common primitive own it.

src/elements/components/DocxEditor/index.tsx
```diff
@@ -260,45 +290,59 @@ function DocxEditor({
...
+                lineHeight: '22px'
+              }
+            }}
+          />
+          {loading && !error && <div css={overlay}>Loading document…</div>}
+          {error && <div css={{ ...overlay, color: '#dc2626' }}>{error}</div>}
+        </div>
+        {/* Grouped review cards for assistant-authored tracked changes.
+            Read-only hosts cannot resolve revisions, so no panel there. */}
+        {editor && !readOnly && (
```
faizalsomani: [question] The rail mounts for every non-readOnly DocxEditor and listRevisionGroups also surfaces one card per HUMAN author, so a plain form user typing with track changes on - no assistant anywhere - now gets a 340px review rail. Is that the intended product scope for this release, or should the rail be gated to hosts where the assistant/tracked-change-groups feature is active?

src/assistant/tools/docx/PARAGRAPH_SPLIT_AUDIT.md
```diff
@@ -0,0 +1,96 @@
+# Paragraph-split failure audit
```
faizalsomani: [shrink] This is a bisect/investigation narrative (references origin/master..hilb-fresh and local oracle commands), not product documentation. It belongs in the PR description or an issue; committed under src/ it ships an internal audit trail with the SDK. Drop it from the diff.

src/elements/components/DocxEditor/TrackedChangeGroups/index.tsx
```diff
@@ -0,0 +1,464 @@
...
+// inspectable via the browser's verbose/debug level so a RECURRING failure
+// that isn't teardown — a real regression — stays visible to developers.
+const debugSwallowed = (error: unknown) =>
+  console.debug('Feathery: tracked-changes rail editor call failed.', error);
+
+// Any raw call into the EJ2 instance can throw once the editor is mid-destroy
+// (step navigation tears the document down under a still-mounted rail; EJ2's
+// observer internals then hit `Object.keys(null)`). A dead editor must read as
+// a no-op, never as a crash that unmounts the form step.
+const quietly = (fn: () => void) => {
```
faizalsomani: [shrink] quietly/debugSwallowed wraps nearly every editor call in this file even though RailErrorBoundary is the declared outer boundary. Event-handler entry points genuinely need their own guard (React boundaries do not catch them), but the per-call wraps inside already-guarded paths (settleRevisions, resolveGroups, refresh's inner catches, the remaining-count reads) are the many-small-boundaries pattern the captain rejected. Keep guards at the event-handler entries, let everything beneath them throw to the boundary.

src/assistant/messageHistory.ts
```diff
@@ -0,0 +1,224 @@
...
+    const currentTurn = i > lastUserIndex;
+    let partsChanged = false;
+    const parts = [...message.parts];
+
+    for (let j = parts.length - 1; j >= 0; j--) {
+      const part = parts[j] as MessagePart & UnknownRecord;
+      if (
+        message.role === 'assistant' &&
+        !currentTurn &&
+        isDanglingToolPart(part)
```
faizalsomani: [question] The client now rewrites dangling historical tool parts to output-error before sending, and the ai-services PR adds interrupted-turn repair at the service boundary for the same shape. Two mechanisms for one concern - which side owns dangling-tool repair? If the server boundary is authoritative, this client rewrite can be dropped; if the client owns it, the server repair should not also rewrite. Document the choice either way.

src/elements/components/DocxEditor/useDocxEditor.tsx
faizalsomani: Addressed in `db020a6466b23e25bc0c3f3f3fd999034883344c`: all shared-surface review installs now use the same `assistantEnabled && !readOnly` predicate as the rail, with tests proving read-only and assistant-off hosts retain native Changes-pane, rendering, and revision-merging behavior.

src/elements/components/DocxEditor/useDocxEditor.tsx
faizalsomani: Addressed in `1e9cdf8fd5c69137550356ec15b9395a88cad697`: the shared revision/group/view primitives moved to `src/utils/documentEditorPrimitives.ts`, and DocxEditor components now import that neutral module while the assistant engine imports and re-exports it for compatibility.

src/elements/components/DocxEditor/useDocxEditor.tsx
faizalsomani: Addressed in `8c512f2bc9a5ea8f32c5338985fddbe72e84f902`, `310173ea9851a5bf4df3be21f6d5473d6970cf33`, and `374cf5610ae5ac180779e50cbd727721b900bf7a`: actual geometry changes use the non-homing editor resize path, unchanged observer ticks are intentionally ignored, and the non-assistant regression proves FitPageWidth plus status-bar chrome refresh without container refresh, cursor homing, or review-state mutation.

src/elements/components/DocxEditor/DocumentEditorContainer.tsx
faizalsomani: Addressed in `4fbcd7ba98fb65be14f03a3e5c95c2e24c06ef16`: the redundant `DocumentEditorContainer` installation call and its duplicate expectations were removed, leaving the common `useDocxEditor` creation path as the single owner.

src/elements/components/DocxEditor/index.tsx
faizalsomani: Addressed in `db020a6466b23e25bc0c3f3f3fd999034883344c`: the rail now mounts only for assistant-enabled editable hosts, so ordinary form typing and every read-only editor stay on the native review surface.

src/assistant/tools/docx/PARAGRAPH_SPLIT_AUDIT.md
faizalsomani: Addressed by the effective cleanup in `b591e357cf75cdf49cb124af7ed57fa90979ae0e`: `PARAGRAPH_SPLIT_AUDIT.md` is absent from the final PR diff, with the investigation content retained only in firstmate's PR-description handoff note.

src/elements/components/DocxEditor/TrackedChangeGroups/index.tsx
faizalsomani: Addressed in `9aa25327e02c32d30f280ed9c7f24cc2828490e5`: the general `quietly` wrapper was replaced by guards at React/EJ2 event entries and listener, history, focus, and resolution calls beneath them now throw normally; narrowly targeted empty-state fallbacks remain only for EJ2 objects that can disappear during teardown.

src/assistant/messageHistory.ts
faizalsomani: Addressed in `a3d7cba8631e9575f2edf24bc8606d2511372c2a`: the service boundary is authoritative for interrupted-tool repair, so the client no longer rewrites dangling parts and tests require pending histories to remain byte-identical while digesting and selection pinning stay client-owned.

src/elements/components/DocxEditor/TrackedChangeGroups/index.tsx
faizalsomani: Follow-up: the code now matches this reply as of `87918a66`. The nine swallowing catch sites that remained below the event-entry guards (revisionCount, refresh, the two inner onSelectionChange catches, settleRevisions, both remaining-count reads, resolveGroups, focusChip) are gone; reads and the mutating resolution calls (`resolveRevisionsAsOneUndo`, `resolveLiveRevisionGroupsAsOneUndo`) throw to the single `handleEditorEvent` entry boundary, and render/effect faults go to `RailErrorBoundary`. The only remaining guards are the `isDestroyed` pre-checks for EJ2 teardown races (each with a one-line constraint comment) and a try/finally in focusChip that exists to lift the selection-echo suppression, not to contain.

src/assistant/tools/docx/syncfusionDocumentOps.ts
```diff
@@ -4080,36 +6097,117 @@ export const ANCHORED_OP_HANDLERS: {
...
+        ]
+      );
+    }
+    const blocks = Array.from(byAnchor.values());
+    const index = blocks.findIndex(
+      (candidate) => candidate.anchor === block.anchor
+    );
+    const next = index >= 0 ? blocks[index + 1] : undefined;
+    if (next && next.kind !== 'table_cell')
+      editor.selection.select(`${block.anchor};0`, `${next.anchor};0`);
```
lanthony42: BLOCKER: when `next` is in another section this spans the section break, and the delete is untracked so rollback has no revision to reject.

src/assistant/tools/docx/syncfusionDocumentOps.ts
```diff
@@ -4208,13 +6306,55 @@ export const ANCHORED_OP_HANDLERS: {
...
+      insertionAnchor = next.anchor;
+    }
+    selectRange(editor, insertionAnchor, 0, 0);
     callEditor(
       editor,
       'insertTable',
       positiveCount(op.rows),
       positiveCount(op.columns)
     );
+    if (Array.isArray(op.initialCells)) {
```
lanthony42, line 7848: BLOCKER: these writes never cross `guardModelAuthoredNumber`, and `compileSectionComposer` routes every composed table through here.

src/assistant/tools/docx/syncfusionDocumentOps.ts
```diff
@@ -4363,499 +6616,1676 @@ function intendedBlockText(
...
-  // must include the paragraph mark (see selectParagraph above).
-  selectBlock(editor, target);
-  const cf = editor.selection.characterFormat;
-  for (const [prop, value] of Object.entries(inherited.characterFormat ?? {})) {
-    if (!isMeaningfulInheritedFormatValue(prop, value)) continue;
-    writeFormatPropIfDifferent(
-      cf,
-      prop,
-      normalizeInheritedCharValue(prop, value)
+function liveTableLayout(
```
lanthony42, line 9198: BLOCKER: this reads materialized widget values rather than what the SFDT stated, so replaying it as the restore writes `allowAutoFit:false` in.

src/elements/components/DocxEditor/TrackedChangeGroups/index.tsx
```diff
@@ -199,62 +194,57 @@ function TrackedChangeGroups({ editor, hidden, onHiddenChange }: Props) {
...
-      remaining = 0;
-    }
-    if (remaining) refocusPanel();
-    else quietly(() => editor?.focusIn?.());
+    if (listRevisionGroups(editor).length) refocusPanel();
+    else editor?.focusIn?.();
+  };
+
+  const resolveGroups = (groupViews: GroupView[], isAccept: boolean) => {
+    resolveLiveRevisionGroupsAsOneUndo(editor, groupViews, isAccept);
```
lanthony42, line 275: BLOCKER: this goes through `robinResolveSelf`, so the group's `onReject` never fires - reject clears the content and leaves the shading behind.

src/assistant/tools/docx/documentIndex.ts
```diff
@@ -210,13 +386,69 @@ const readSnapshot = (editor: any): IndexSnapshot | null => {
...
+      blocks,
+      headers,
+      contentHash: digest
+    });
+  }
+
+  const { failed, storedBlocks } = result;
+  // The server reporting fewer blocks than were sent (or failed embeds)
+  // means the index does NOT hold this document. Preserve the last confirmed
+  // base so a later retry can finish the same CAS delta.
```
lanthony42: BLOCKER: right for delta, wrong for full - that path already mutated chunks, so keeping the base lets the next delta certify a stale one.

src/assistant/tools/docx/syncfusionDocumentOps.ts
```diff
@@ -4080,36 +6097,117 @@ export const ANCHORED_OP_HANDLERS: {
...
+      );
+    }
+    const blocks = Array.from(byAnchor.values());
+    const index = blocks.findIndex(
+      (candidate) => candidate.anchor === block.anchor
+    );
+    const next = index >= 0 ? blocks[index + 1] : undefined;
+    if (next && next.kind !== 'table_cell')
+      editor.selection.select(`${block.anchor};0`, `${next.anchor};0`);
+    else selectParagraph(editor, block);
```
lanthony42: On the document's last paragraph there is no paragraph mark left to consume, so this reports `ok:true` over an unchanged document.

src/assistant/tools/docx/syncfusionDocumentOps.ts
```diff
@@ -6341,16 +10763,1265 @@ function detectUnannouncedChain(
...
+        tableLabel,
+        'needs at least one string in `columnHeaders`.'
+      );
+    if (
+      !Array.isArray(table.rows) ||
+      table.rows.some(
+        (row) =>
+          !Array.isArray(row) ||
+          row.length !== table.columnHeaders.length ||
+          row.some((cell) => typeof cell !== 'string')
```
lanthony42, line 13755: Cells only get a type check while `sectionText` rejects newlines for titles and paragraphs, so a newline here splits the cell into two anchors.

src/elements/components/DocxEditor/TrackedChangeGroups/index.tsx
```diff
@@ -149,41 +138,47 @@ function TrackedChangeGroups({ editor, hidden, onHiddenChange }: Props) {
   // (a DIFFERENT document opened in place) also rebuilds immediately.
   useEffect(() => {
     if (!editor) return;
+    // EJ2 teardown race: a destroyed instance throws on any touch, so read it as "no editor".
+    if (editor.isDestroyed) {
+      setGroups([]);
+      return;
+    }
     refresh();
```
lanthony42: A throw here reaches `RailErrorBoundary`, which has no reset, so one mid-destroy read removes the rail for the rest of the session.

src/elements/components/DocxEditor/useDocxEditor.tsx
```diff
@@ -635,7 +685,9 @@ export function useDocxEditor({
     };
     // `source` / `isReadOnly` intentionally omitted — open and readOnly are
     // handled by sibling effects so we never tear down mid-fetch.
-  }, [resolvedLicenseKey, serviceUrl, headersKey]);
+    // `reviewGate` (not `reviewChanges`) so a gate flip waits out any
+    // in-flight load before recreating - see its declaration above.
+  }, [resolvedLicenseKey, serviceUrl, headersKey, reviewGate]);
```
lanthony42, line 877: `reviewGate` in these deps means a readOnly flip now recreates the editor, and the reopen re-fetches the pre-save `sourceUrl`.

lanthony42 (changes requested):

src/elements/components/DocxEditor/TrackedChangeGroups/index.tsx
faizalsomani: Commit `5bd8cbeb`.

Correct, and it made the restore above worth nothing from the UI: the rail resolves every card - chip, group and rail-wide - through the non-cascading path, so the appearance inverse was only ever reached from the native cascading `reject()`.
It now fires on that route too, once per group, as the last undecided member is rejected - so it still runs before that member's content resolves, while the row and cell indices it names are valid.
A group with a member accepted keeps its appearance: part of the change survived and still needs it.
Proven end to end by rejecting through `resolveLiveRevisionGroupsAsOneUndo` and asserting row count, cell appearance and stated layout all return to their pre-write values.

src/assistant/tools/docx/syncfusionDocumentOps.ts
faizalsomani: Commit `35064130`.

Correct. A paragraph mark sits BETWEEN two paragraphs, so the range needs a paragraph on both sides in THIS section; extending into the next section spanned the break, and a deleted section break is not something SyncFusion authors a rejectable revision for, so a group rollback had nothing to put back.
The neighbour is now taken only from this section, and `delete_paragraph` joins `TRACKED_STRUCTURAL_OPS`, so the engine's existing post-write assertion refuses any future path here that mutates without a rejectable deletion card rather than reporting success.

src/assistant/tools/docx/syncfusionDocumentOps.ts
faizalsomani: Commit `35064130`.

Correct - the same fix, because both cases are "this path does not know where the document ends".
At the end of a section, the document's last paragraph included, the old fallback selected the paragraph's own content, which on an already-empty paragraph deleted nothing while the op still reported success.
It now consumes the PRECEDING mark - the same range a person removes a trailing empty paragraph with - so accepting really removes the paragraph and rejecting restores the document byte for byte; where neither neighbour is a body paragraph of this section there is no mark to take and nothing is written.
The tracked-structural assertion above is the backstop: an unchanged document can no longer come back `ok`.

src/elements/components/DocxEditor/TrackedChangeGroups/index.tsx
faizalsomani: Commit `8e987ea1`.

Correct, and it is fallout from our own last round. We swung too far: removing the swallowing catches was right, latching the rail off for the session was not - the pending changes stay in the document with nothing left to see them with.
Two changes, one per kind of failure. An instance on its way out is an expected state, so the mount effect's read asks the instance whether it is gone and treats that as "no editor"; a failure on a LIVE editor is a programming fault and still reaches the boundary.
And the boundary now retries, briefly and at most twice, and is remounted with its retries restored when the editor instance changes or the document is reopened - a fault that reproduces re-latches on the retry, so there is no loop and no flicker.

src/elements/components/DocxEditor/useDocxEditor.tsx
faizalsomani: Commit `8e987ea1`.

Correct, and reachable in the product: the host derives the gate from `readOnly` (`assistantEnabled && !readOnly`), so finalizing an envelope for signature flips it mid-session, and the reopen re-fetched the pre-save url - unreviewed assistant edits and unsaved typing replaced by older bytes, with no signal.
The recreation now carries the live document with it: the outgoing instance is serialized in the create effect's cleanup and the new one opens exactly that.
Only when there is something to lose - unsaved edits or pending revisions - so a pristine document still reopens from its source, which stays the most faithful path for it; and the stash is keyed to the source it came from, so a regenerate or a bumped `openNonce` always wins over it.
The gate flip still recreates the instance, because a gated-off editor must not retain instance-scoped renderer patches - that part of the last round's fix was right.

src/assistant/tools/docx/syncfusionDocumentOps.ts
faizalsomani: Commits `ddb9416f`, `814e4303`. Right, and it was the only cell-write path in the vocabulary with no provenance requirement at all: `insert_table` writes its whole `initialCells` matrix inside one op, so the per-cell gate - which reads one op's text against the one cell it targets - never saw those figures.

src/assistant/tools/docx/syncfusionDocumentOps.ts
faizalsomani: Commit `5bd8cbeb`. Correct. `liveTableLayout` always answers with a concrete `allowAutoFit` and, once the table has been laid out, with rendered `cellWidth` values the document never stated, so replaying it as the restore wrote `allowAutoFit:false` and a point-width grid INTO a table that had neither - a reject adding untracked damage instead of removing it.

src/assistant/tools/docx/documentIndex.ts
faizalsomani: Commit `dd67b729`. Correct, and the difference is what the service has already done by the time it reports the failure. Traced through the whole path rather than the client line alone: `indexDelta` returns before `applyGeneratedDocDelta` when an embed fails, so a refused delta mutated no chunk and left the marker at the base this client holds - which is why keeping it is right there.

src/assistant/tools/docx/syncfusionDocumentOps.ts
faizalsomani: Commit `c7e1d63b`. Correct. I read the rule rather than assuming it: the ai-services schema types section text as `z.string().min(1)` and cells as plain strings and carries no newline rule at all, so this is the engine's own rule and the engine is the only place it can hold.

src/utils/documentEditorPrimitives.ts
```diff
@@ -0,0 +1,1382 @@
...
+      if (state.resolved || resolvedAlone.has(index)) return;
+      // The non-cascading path the review rail resolves every card through:
+      // per-chip, per-card and rail-wide all arrive here, member by member. The
+      // group's appearance inverse has to fire on this route too, or rejecting a
+      // card clears the content and leaves the shading behind. It fires as the
+      // LAST undecided member is rejected - by then the whole group is a reject,
+      // and the restore still runs before that member's content resolves.
+      const lastUndecided = resolvedAlone.size === members.length - 1;
+      if (!isAccept && !acceptedAlone.size && lastUndecided)
+        restoreAppearance();
```
lanthony42: Two groups writing the same cell, rejected in document order (what the rail shows), leave an assistant fill in the document after both cards are rejected.

src/utils/documentEditorPrimitives.ts
```diff
@@ -0,0 +1,1382 @@
...
+      resolvedAlone.add(index);
+      if (isAccept) acceptedAlone.add(index);
+      resolveSingleRevision(members[index], isAccept);
+      // After the last member's content resolves: this is the route the rail's
+      // card buttons take, so the style inverse has to fire here too or a
+      // resolved card leaves a restyled paragraph behind. It runs for accept as
+      // well as reject, because both merge a paragraph mark away.
+      if (lastUndecided) restoreParagraphStyles();
```
lanthony42: Accepting one chip then rejecting the rest fuses two headings and empties the rail, so no card is left to undo it.

src/utils/documentEditorPrimitives.ts
```diff
@@ -0,0 +1,1382 @@
...
+      .join('');
+  const identify = (block: any) => paragraphIdentityText(readText(block));
+  for (const restore of restores) {
+    const [sectionIndex, blockIndex] = restore.anchor.split(';').map(Number);
+    const atAnchor = sections[sectionIndex]?.[blockIndex];
+    let target: { section: number; block: number } | undefined;
+    if (atAnchor && identify(atAnchor) === restore.text)
+      target = { section: sectionIndex, block: blockIndex };
+    else {
+      // The anchor moved. Fall back to the one paragraph that still reads the
```
lanthony42, line 1552: The appearance restore above still selects by positional `cellAnchor`, which is the failure this fallback exists to avoid.

src/assistant/tools/docx/tests/acceptGroupsWithTableFormat.spec.ts
```diff
@@ -0,0 +1,990 @@
...
+    } finally {
+      destroyEditor(ed);
+    }
+  });
+
+  // The review rail never calls the native cascading reject: every card - chip,
+  // group and rail-wide - resolves member by member through robinResolveSelf, so
+  // the group's appearance inverse was never reached from the UI at all. Reject
+  // cleared the content and left the shading behind.
+  it('restores appearance when the rail rejects the group member by member', () => {
```
lanthony42, line 760: This group has one revision-bearing op, so `lastUndecided` is true on the first iteration - add a `set_cell_text` and it goes red.

src/assistant/messageHistory.ts
```diff
@@ -0,0 +1,206 @@
...
+  messages,
+  body,
+  trigger,
+  messageId
+}: PrepareRequestOptions): { body: Record<string, unknown> } {
+  return {
+    body: {
+      ...body,
+      id,
+      messages: prepareAssistantMessagesForRequest(messages),
```
lanthony42: `e024cc1f` reverted `24765b33` here - net diff for this file across the round is empty, and the spec came back too so CI stayed green.

lanthony42 (changes requested):

src/assistant/messageHistory.ts
faizalsomani: You're right, and thanks for catching it. A later commit was assembled by diffing a branch that predated `24765b33` and applying it whole, so it carried a revert of all three files - and the spec reverted with the source, which is why CI stayed green. Restored in `6364d99f`, content identical to `24765b33`. No test can catch this since the test disappears too, so I added a pre-push check that walks the round's commits and flags any whose diff is no longer visible at the tip. It flags this one on the old tip and passes on the fixed one. It lives in our staging tooling rather than the repo - happy to move it if you'd rather see it here.

src/utils/documentEditorPrimitives.ts
faizalsomani: Confirmed and fixed in `a594b9ff`. The snapshots are an undo stack, but the rail hands out cards in document order, so the older card's restore went back first and the newer one wrote the older card's fill on top. The stack is now owned by the document: a rejected card hands its snapshot down to the newest sibling still holding that cell and writes only when it's last; an accepted card cancels the older snapshots under it. Worth flagging how this one was verified, since it changed the fix. Four unit tests were red before and green after, and the product was still broken: they all built both cards inside one change set, and two cards from two separate turns hit the same defect. The browser run is what found it - Robin won't emit two groups in one turn, so the live path forced the cross-turn shape. Live now: baseline fill restored exactly after rejecting both cards, neighbouring rows untouched.

src/utils/documentEditorPrimitives.ts
faizalsomani: Confirmed - it welds "National Capabilities" onto the following Heading 1 and empties the rail. The cause was the chip being one SyncFusion revision: a paragraph's text and its paragraph mark are separate revisions, so accepting one and rejecting the other removes the boundary that made it a paragraph. In `1c365b43` a chip is a paragraph's worth of change and every resolve path settles its revisions together. The empty chips those marks used to produce are gone from the rail too. Checked live on your exact sequence: inserted a subsection, accepted one chip, rejected the rest - the two headings stay two headings, each keeping its style.

src/utils/documentEditorPrimitives.ts
faizalsomani: Agreed - it was the odd one out next to `replayParagraphStyles`. In `1c365b43` each snapshot binds to its cell's paragraph widget and asks the SDK where that widget sits when it replays, so a row that moved is still found and a row that went away restores nothing. Binding at bind time rather than at first resolve matters: a sibling card rejected first has already moved the rows. I also folded the two copies of that replay into one - the ops-module copy is now `rollbackAppearanceWrites` and only does mid-apply rollback.

src/assistant/tools/docx/tests/acceptGroupsWithTableFormat.spec.ts
faizalsomani: Right that it couldn't fail - one revision-bearing op means the only member is also the last. Fixed in `1c365b43`: two revision-bearing ops plus an explicit assertion that the card has more than one member. One correction on the prediction: adding a `set_cell_text` alone doesn't turn it red - I checked against the unpatched base. What exposes the drift is a snapshot whose row moves before it replays, so that's a separate test now (red on base, green after). I also removed the counter this is really about: a card is finished when the document holds none of its revisions, which can't be miscounted. Siblings in that spec don't share the flaw - the partial-accept one already asserts `items.length > 1`.

src/assistant/tools/docx/tests/splitTable.spec.ts
```diff
@@ -0,0 +1,1274 @@
...
+  for (const change of changes)
+    out[change.revisionType] = (out[change.revisionType] ?? 0) + 1;
+  return out;
+};
+
+/** Row texts of one table, in order. */
+const rowsOf = (editor: DocumentEditor, index: number): string[] =>
+  tablesOf(editor)[index]?.rows.map((row) => row.text) ?? [];
+
+describe('split_table: the selective split, rows anywhere in the table', () => {
```
lanthony42, line 330: BLOCKER: no fixture has two adjacent tables - `markInclusiveRangeEnd` ends at `next.anchor;0`, so a neighbouring table is captured and pasted back mangled, `ok:true`.

src/assistant/tools/docx/tests/splitTable.spec.ts
```diff
@@ -0,0 +1,1274 @@
...
+      expect(rowsOf(viaBoundary, 0)).toEqual(rowsOf(viaList, 0));
+      expect(rowsOf(viaBoundary, 1)).toEqual(rowsOf(viaList, 1));
+    } finally {
+      destroyEditor(viaBoundary);
+      destroyEditor(viaList);
+    }
+  });
+});
+
+describe('split_table: refusals derived from the document', () => {
```
lanthony42, line 951: BLOCKER: `detectBatchedSplits` only covers one change set - a second split while the first is still pending fuses paragraphs and leaves an orphan header-only table, 3/3 targets.

src/assistant/tools/docx/tests/sectionFamilyInheritance.spec.ts
```diff
@@ -0,0 +1,1063 @@
...
+  };
+
+  it('has coverage for every op the engine calls content-creating', () => {
+    expect(Object.keys(recipes).sort()).toEqual(
+      Array.from(CONTENT_CREATING_OPS).sort()
+    );
+  });
+
+  it.each(Array.from(CONTENT_CREATING_OPS))(
+    '%s inherits the document look rather than the editor default',
```
lanthony42, line 1119: BLOCKER: `composedParagraphDonor` asks for `'section_heading'` hardcoded and nulls the op's `styleName`, so `apply_style {styleName:'Body Text'}` on a paragraph the same change set created yields `sectionBanner` 16pt bold.

src/assistant/tools/docx/tests/documentTailTableDelete.spec.ts
```diff
@@ -0,0 +1,394 @@
...
+
+/** The last data row index of the table at `tableAnchor`. */
+const lastRowOf = (editor: DocumentEditor, tableAnchor: string): number =>
+  Math.max(
+    ...flattenSfdt(JSON.parse(editor.serialize()))
+      .filter((block) => block.anchor.startsWith(`${tableAnchor};`))
+      .map((block) => Number(block.anchor.split(';')[2]))
+  );
+
+describe('the SyncFusion defect this guard exists for', () => {
```
lanthony42, line 195: BLOCKER: `delete_table` (`:7773`) has no guard, so a tail table still throws `Cannot read properties of undefined (reading 'childWidgets')` inside `acceptAll` and breaks the rail.

src/assistant/tools/docx/tests/relocateSection.spec.ts
```diff
@@ -0,0 +1,1340 @@
...
+    }
+  });
+});
+
+// The rail card's own Accept/Reject buttons do NOT call revisions.rejectAll():
+// they call resolveLiveRevisionGroupsAsOneUndo over the card's group, which is a
+// different path with a different failure mode (it is what Anthony's 3729780254
+// was about). Live, the browser tab died the first time that button was clicked
+// on a relocation, and the retry then completed cleanly - so the death was the
+// machine, not the code. This settles it deterministically instead: the path the
```
lanthony42, line 1468: BLOCKER: it was the code - with a header present, `move_section` returns `ok:true` then accept and rail reject both never return. `f6695b0c` is tail-table deletion and doesn't touch it.

src/assistant/tools/docx/tests/relocateSection.spec.ts
```diff
@@ -0,0 +1,1340 @@
...
+      );
+      expect(result.results[0]).toMatchObject({ ok: true });
+      editor.revisions.rejectAll();
+      expect(editor.serialize()).toBe(before);
+    } finally {
+      destroyEditor(editor);
+    }
+  });
+
+  it('moves the last section of a document that ends in a paragraph', () => {
```
lanthony42, line 479: BLOCKER: `resolveRelocationTarget:6804` still returns `${tail.anchor};${tail.length}`, so all four tail shapes fuse on accept - the tail paragraph came back as `Heading 2` reading `...Friday.National Capabilities`.

src/assistant/tools/docx/tests/opContracts.spec.ts
```diff
@@ -1199,6 +1570,28 @@ describe('op contracts: every advertised op works over its real route', () => {
     expect([...Object.keys(CONTRACTS)].sort()).toEqual([...registered].sort());
   });
 
+  it('every registered content-creating op crosses a tracked mutation contract', () => {
```
lanthony42, line 1573: BLOCKER: `split_table` is in neither tracked set so this returns immediately for it, and the relocation ops only compare the reject-side text projection - the accept side is unverified, which is why all of the above report `ok:true`.

src/assistant/tools/docx/tests/relocateSection.spec.ts
```diff
@@ -0,0 +1,1340 @@
...
+      // SyncFusion retiring a section that no longer has content, on accept.
+      const sfdt = JSON.parse(editor.serialize());
+      const sections: any[] = sfdt.sec ?? sfdt.sections;
+      expect(
+        sections.map((section) => (section.b ?? section.blocks).length)
+      ).not.toContain(0);
+      // Page setup is intact on what remains - a relocation carries content, not
+      // page geometry.
+      for (const section of sections)
+        expect((section.secpr ?? section.sectionFormat).pw).toBe(612);
```
lanthony42: All three fixture sections declare `pw: 612`, so this can't fail - moving a section that is its own Word section drops its landscape page and margins and goes 3 sections to 2.

lanthony42 (changes requested):

src/assistant/tools/docx/tests/splitTable.spec.ts
faizalsomani: Right, and thank you - this was live, not just an untested shape. `markInclusiveRangeEnd` now crosses into `next` only when `next` is a body paragraph of the same Word section. A table's cells are its blocks, so the block after a table is a `table_cell` the section test waved through, and ending at its offset 0 selected the whole neighbour. Fixed in 49b3e065, with cases for author-made and split-made adjacency, accept and reject.

src/assistant/tools/docx/tests/splitTable.spec.ts
faizalsomani: Same root cause as the adjacent-table one, and the same one condition fixes it - a split is how two tables become adjacent, so the second split reached it. Reproduced on the unpatched base first: fused paragraphs plus a spurious extra table; after 49b3e065, no fusion and reject byte-exact. Two splits in one change set stay refused separately, by `detectBatchedSplits`.

src/assistant/tools/docx/tests/relocateSection.spec.ts
faizalsomani: You're right that it hangs, but it's jsdom, not the product. jest-canvas-mock gives a non-empty header widget a NaN height, that reaches viewer.clientArea, and Layout.shiftWidgetsForPara then splits the same paragraph forever (5772->5846 pages in 7s). Control: replace_text hangs identically, no relocation involved. Real Chrome, same header, real move_section: accept 116ms, reject 169ms byte-exact. 53b75dc1 fixes the harness (layoutType Continuous). No product guard - every real DOCX has a header.

src/assistant/tools/docx/tests/documentTailTableDelete.spec.ts
faizalsomani: Confirmed. delete_table returns ok:true and acceptAll then throws on childWidgets; rail reject is clean, so it's accept-only, exactly like delete_row. 8004b332 routes it through the same documentTailTableLastRow predicate, with a registry-enumerating test so a future deleting op can't bypass it. Worth flagging: delete_table is unguarded on master too, so this one is live today rather than something this PR introduced.

src/assistant/tools/docx/tests/relocateSection.spec.ts
faizalsomani: Two things, and you're right on the first. All three sections declare 612 so the assertion can't bite; b6178df5 gives the moved section landscape and its own margins and asserts each survivor keeps its own geometry. The collapse to 2 is deliberate and stated in the PR body - a relocation carries content, not page geometry - because a section break authors no rejectable revision, which is the untracked_write case. Happy to make geometry travel instead if you'd prefer that.

src/assistant/tools/docx/tests/sectionFamilyInheritance.spec.ts
faizalsomani: You're right, and it's worse than the style name. Measured on this file's fixture: apply_style {styleName:'Normal'} on a paragraph the change set created came back subsectionBanner 14pt bold - the heading donor - and reported no disagreement, because `requested` read the block's current style rather than the op's. The composer already uses per-role donors; this one path did not. Both halves fixed in c2d01150.

src/assistant/tools/docx/tests/relocateSection.spec.ts
faizalsomani: Confirmed - the caret lands at end-of-text rather than past the mark, so accept fuses and strands an empty paragraph behind, and it also makes split_table's reject non-byte-exact when the target is the last block. Two notes: length + 1 alone does NOT fix it (measured - the paste caret is clamped), so a8e556b4 gives the payload a real paragraph start to land in; and 49b3e065 does not touch this line, so you were right that they are separate.

src/assistant/tools/docx/tests/opContracts.spec.ts
faizalsomani: Both halves right, and the second half proved itself. d77a8d6d adds split_table to the tracked set, with copy_section as precedent. Adding it immediately turned a split test red - and the failure was real, landing exactly on the tail-fusing defect above. So the missing membership is what hid it. The relocation and split suites now assert the accepted document, not only the reject projection. That surfaced two further rail-path defects, which I have filed rather than buried.

lanthony42 (approved):

PR #1754, opened 2026-08-07
fix: load Google fonts with display=swap so text isn't blank on cold cache

trebitowski:
## Changes

Customer report: on SDK forms, text is fully invisible on a cold cache until the custom fonts finish downloading (seen on 2.52.0).

**Root cause.** `_loadFormPackages` loaded Google fonts through `webfontloader`:

```js
WebFont.load({ google: { families: res.fonts } })
```

`webfontloader`'s google module has no way to pass `display` through to Google's CSS endpoint — `node_modules/webfontloader/src/modules/google/fontapiurlbuilder.js` only ever builds `?family=…&subset=…&text=…`. Google's default response omits `font-display`, so the faces fall back to `font-display: auto` and the browser applies its default block period, blanking text until the font arrives.

**Fix.** New `src/utils/fonts.ts` requests the stylesheet directly with `&display=swap`, so text paints immediately in the fallback face and swaps in when the webfont lands. Both SDK call sites now route through it:

- `FeatheryClient._loadFormPackages` — the path in the report
- `SignatureField` — had the identical bug for `La Belle Aurore`

Since nothing reads `global.webfontloaderPromise` anymore, the `init()` preload, the ambient type, and the `webfontloader` / `@types/webfontloader` dependencies are removed — one less chunk fetched on every SDK init.

**Not addressed (deliberate):** the reporter also self-hosts these families via `next/font` and asked about skipping already-loaded faces. A `document.fonts.check()` guard wouldn't help them — `next/font` registers a generated family name (`__Inter_e8ce0c`), not `Inter`, so the check returns false and the SDK would fetch anyway. With `display=swap` the duplicate fetch is no longer render-blocking, so it costs bandwidth rather than paint. Worth a real opt-out prop if bandwidth becomes the complaint.

**Same pattern elsewhere:** `feathery-frontend/src/utils/themes.ts:597` has the same `WebFont.load` call. Left alone here — it's the builder, not customer-facing, and it depends on WebFont's `active` callback for `isFontLoaded.storeFonts`.

- [ ] Tested end to end
- [x] Included screenshots or walkthrough video of change if impacts UX — n/a, no visual change beyond text no longer being invisible during load

`yarn typecheck` passes; `fonts.spec.ts` (3 new tests covering `display=swap`, multi-word/multi-family URL encoding, and dedupe) and `featheryClient.spec.ts` (31 tests) pass. ESLint clean on the changed files.

## Related pull requests

None

### DIFF
--- a/src/utils/fonts.ts
+++ b/src/utils/fonts.ts
@@ -1,0 +1,29 @@
+import { featheryDoc, runningInClient } from './browser';
+
+const loadedGoogleFonts = new Set<string>();
+
+/**
+ * Loads Google fonts via a stylesheet link rather than webfontloader, which
+ * has no way to pass `display` through to the CSS API. Without it Google's
+ * response omits font-display, so the browser falls back to the default block
+ * period and text renders invisible on a cold cache until the font arrives.
+ *
+ * @param families webfontloader-style specs, e.g. 'Inter:400,400italic,700'
+ */
+export function loadGoogleFonts(families: string[]) {
+  if (!runningInClient()) return;
+
+  const newFamilies = families.filter(
+    (family) => family && !loadedGoogleFonts.has(family)
+  );
+  if (!newFamilies.length) return;
+  newFamilies.forEach((family) => loadedGoogleFonts.add(family));
+
+  const link = featheryDoc().createElement('link');
+  link.rel = 'stylesheet';
+  // Google's v1 CSS API: families joined by '|', spaces as '+'
+  link.href = `https://fonts.googleapis.com/css?family=${newFamilies
+    .map((family) => family.replace(/ /g, '+'))
+    .join('%7C')}&display=swap`;
+  featheryDoc().head.appendChild(link);
+}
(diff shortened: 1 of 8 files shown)


src/utils/fonts.ts
```diff
@@ -0,0 +1,29 @@
...
+ * @param families webfontloader-style specs, e.g. 'Inter:400,400italic,700'
+ */
+export function loadGoogleFonts(families: string[]) {
+  if (!runningInClient()) return;
+
+  const newFamilies = families.filter(
+    (family) => family && !loadedGoogleFonts.has(family)
+  );
+  if (!newFamilies.length) return;
+  newFamilies.forEach((family) => loadedGoogleFonts.add(family));
```
joondotcode, line 20: Can we clear each family from `loadedGoogleFonts` if the stylesheet fails to load? Right now a transient request failure leaves it marked as loaded, so cached form loads and signature remounts skip the font for the rest of the page session.

joondotcode (approved):

src/utils/fonts.ts
trebitowski: Good catch — confirmed: the set was populated before the stylesheet loaded, so one transient failure skipped the family for the rest of the page session (a regression vs webfontloader, which re-inserted a link per call). Fixed in 7db08511: `link.onerror` now unmarks the families and removes the dead link, so the next `loadGoogleFonts` call retries. Covered by a new spec case.

PR #1762, opened 2026-08-11
feat: render image alt text from alt_text with aria_label fallback

FranklinGao123:
## Why

Form authors have no way to put alt text on images — the only input is "Aria Label," and this SDK hardcodes `alt=''` on every rendered ``, so broken images show a blank box, crawlers/exports/email clients see a "decorative" image, and authors auditing for WCAG find no alt text support. The rendering lives here, so the fix starts here: the builder's new Alt Text field (feathery-org/feathery-frontend#3795) is inert until this ships, is published, and the frontend's pin is bumped.

## Changes

`ImageElement` hardcodes `alt=''` and puts the accessible name on `aria-label`. The label does announce (presentational-role conflict resolution means `aria-label` suppresses the implied decorative role), but there's no way to get a real `alt` attribute, so:

- a broken image renders a blank box instead of its description — `alt` is the browser's fallback text, `aria-label` isn't;
- the markup self-contradicts: `alt=''` asserts decorative, `aria-label` asserts the opposite, and anything reading HTML rather than the accessibility tree (crawlers, exports, email clients) sees a decorative image.

This renders the accessible name through a real `alt`:

- **`src/utils/accessibility.ts`** (new) — `getImageAltText(properties)` → `alt_text || aria_label || ''`. Truthy `||`, not `??`: both properties default to `''`, so nullish coalescing would never reach the legacy fallback. It's a pure module rather than living in `utils/image`, which suites `jest.mock` wholesale for its File/blob dependencies — that placement broke 22 tests before the move.
- **`src/elements/basic/ImageElement.tsx`** — ``, `aria-label` dropped from the `` branch: an `` carrying both would have `aria-label` silently win over `alt`, which would make the builder's new Alt Text field appear to do nothing. The PDF `<embed>` branch keeps the name on `aria-label` (with the same fallback) since `alt` isn't a valid embed attribute, and loses its bogus `alt=''`. `{...elementProps}` stays last so host overrides still win.
- **`src/assistant/tools/panelRuntime.ts`** — the page collector uses the same helper, so the assistant sees the text a screen reader would. It previously read only `props.alt_text` (which nothing wrote) and missed every legacy `aria_label`.

### Backwards compatibility

The fallback means every existing form with an `aria_label` keeps its accessible name with no author action — same text, now on the correct attribute, plus broken-image fallback. `alt_text` wins when both are set; neither set → `alt=''` (genuinely decorative). The builder currently writes both keys with the same value (see the frontend PR), so this renders identically before and after this change ships.

No version bump in this diff — release is the manual `Release & Publish` workflow dispatch; suggest **minor**.

- [x] Tested end to end

`yarn test ImageElement`: 20/20 (14 existing + 6 new covering all four resolution branches and the embed case). Full suite: 1867/1869 — the 2 failures are a timing-flaky docx suite that fails identically on the base commit (verified in a clean worktree). `yarn lint` and `yarn typecheck` clean. Also verified live: builder canvas and hosted draft form both render `alt` from the same stored properties.

## Related pull requests

- feathery-org/feathery-frontend#3795 — the builder's Alt Text field writing `alt_text`
- feathery-org/feathery-backend#3622 — defaults, Robin agent schema, inheritance pruning

### DIFF
--- a/src/assistant/tools/panelRuntime.ts
+++ b/src/assistant/tools/panelRuntime.ts
@@ -20,6 +20,7 @@ import {
 } from '../../utils/stepper';
 import { findClickableAncestorSubgrids, getTableCapabilities } from './utils';
 import { sanitizeTransportValue } from '../../utils/transportValue';
+import { getImageAltText } from '../../utils/accessibility';
 
 export type PanelRuntimeFieldEntry = {
   key: string;
@@ -614,7 +615,7 @@ export const getPanelRuntimeSnapshot = (
   (step.images ?? []).forEach((el: any) => {
     const props = el?.properties ?? {};
     const altRaw =
-      (typeof props.alt_text === 'string' && props.alt_text) ||
+      getImageAltText(props) ||
       (typeof props.caption === 'string' && props.caption) ||
       '';
     if (!altRaw) return;
--- a/src/elements/basic/ImageElement.tsx
+++ b/src/elements/basic/ImageElement.tsx
@@ -1,6 +1,7 @@
 import { useEffect, useMemo, useState } from 'react';
 import { fieldValues } from '../../utils/init';
 import { getRenderData } from '../../utils/image';
+import { getImageAltText } from '../../utils/accessibility';
 
 export const PLACEHOLDER_IMAGE =
   'https://feathery.s3.us-west-1.amazonaws.com/theme-image-preview.png';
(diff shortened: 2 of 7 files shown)


FranklinGao123:
Superseded — this branch bundled two unrelated features, and both have landed as focused PRs:

- Image alt text → #1765 (`feat/image-alt-text-only`)
- Themeable inner padding for input-box fields → #1760 (`feat/field-inner-padding`), which is a superset of the inner-padding work here (this branch had 17 lines of `elements/styles.ts` and 147 of `innerPadding.spec.ts`; #1760 has 507 and 598, plus the dropdown fields)

Nothing here is unique to this branch. Closing to keep the two features independently reviewable and to avoid conflicts on the files they share.

PR #1767, opened 2026-08-12
fix: stop an uploaded font from hijacking a same-named catalog font

FranklinGao123:
## Changes

An org-uploaded font was registered as `new FontFace(family, ...)` under its raw family name. Because CSS font matching is by name, an upload called `Arial` then won for **every** stack naming Arial, restyling published-form text that never chose the upload — and the choice was unrecoverable, since a theme stores only the bare family string.

Uploads now register under `Uploaded--<family>`, and `transformFontFamilies` — the single choke point every `font_family` style value passes through — rewrites a value naming an upload to that aliased family. The registry it reads is populated by `_loadFormPackages` before any step renders, from the same `uploaded_fonts` payload that drives registration, so the two halves cannot drift.

Key properties:

- **Persisted values are unchanged.** Stored `font_family` stays the bare name. This is deliberate: there is no SDK-version negotiation (every build requests `panel/v20`), so a namespaced persisted value would emit invalid CSS on older pinned builds. Old builds keep today's behavior until upgraded.
- **Whole-value matching only, never per comma-segment.** Picking an upload persists its bare name, whereas a catalog stack like `Arial, sans-serif` merely *mentions* the family — rewriting segments would hand every such stack back to the upload, which is the bug being removed.
- Names that cannot be safely quoted into CSS (quotes, backslashes, control characters) are never aliased and keep fully legacy behavior.
- The prefix is vendor-neutral on purpose: it is visible in DevTools computed styles on customer forms.

Also fixes a pre-existing bug in `transformFontFamilies` (separate commit): when a stack's family contained an unquoted space it returned the **entire** input quoted, so `My Font, sans-serif` became `'My Font, sans-serif', sans-serif` — one bogus quoted family that browsers skip, falling through to the fallback instead of the named font.

### Behavior change worth calling out

Raw CSS in `custom_css` is deliberately **not** rewritten. A customer rule naming `Arial` now resolves to the real Arial rather than their uploaded look-alike. Orgs affected are those with an upload named exactly like a catalog family; the intent is that only text which explicitly selected the upload gets it.

### Tests

- `transformFontFamilies`: passthrough with an empty registry, bare-name rewrite, catalog stacks left alone, case-insensitive matching, idempotence, comma-containing upload names, unsafe names untouched, plus a regression test for the quoting bug.
- `_loadFormPackages`: registers the aliased family and populates the rewrite registry.
- Integration test across both halves: after loading an `uploaded_fonts` payload, `ResponsiveStyles.applyFontFamily` (the function every rendered element goes through) emits `'Uploaded--Arial'` for text that selected the upload and leaves `Arial, sans-serif` alone.

Note: two tests in `integrationClient.spec.js` (document-signing `signers`) fail on this branch and **also fail on `master`** — pre-existing and unrelated to fonts.

- [ ] Tested end to end — unit + cross-module integration tested; **not yet run against a real hosted form with a locally built bundle**, worth one manual check on staging before release
- [x] Included screenshots or walkthrough video of change if impacts UX — n/a, no visual change except fonts resolving correctly

image
image
image

## Related pull requests

- feathery-org/feathery-frontend#3770 — the dashboard half: same alias in the builder canvas, plus the picker/grouping work. Verified there that a loaded upload no longer captures the bundled Arial.
- feathery-org/feathery-backend#3628 — reserves the `Uploaded--` prefix at upload time, and stops requesting a Google stylesheet for a family the org has uploaded.
- feathery-org/feathery-react#1754 — also modifies `_loadFormPackages` and adds a `src/utils/fonts.ts`. This **used to conflict**; the loading block here has since moved into `utils/uploadedFonts.ts`, which mirrors how #1754 extracts Google-font loading and leaves only a two-line diff in `featheryClient/index.ts`. A trial merge against #1754 is now **clean**, so the two can land in either order.

### DIFF
--- a/src/elements/__test__/styles.spec.js
+++ b/src/elements/__test__/styles.spec.js
@@ -192,4 +192,26 @@ describe('responsiveStyles', () => {
       expect(actual.textAlign).toBe('center');
     });
   });
+
+  describe('transformFontFamilies', () => {
+    const transform = (families) =>
+      new ResponsiveStyles(
+        mockElement,
+        [],
+        false,
+        DEFAULT_MOBILE_BREAKPOINT
+      ).transformFontFamilies(families);
+
+    it('passes single unspaced families through unquoted', () => {
+      expect(transform('Arial')).toBe('Arial');
+    });
+
+    it('quotes only the spaced family in a stack, not the whole value', () => {
+      expect(transform('My Font, sans-serif')).toBe("'My Font', sans-serif");
+    });
+
+    it('leaves already-quoted families alone, normalizing to single quotes', () => {
+      expect(transform('"Open Sans", serif')).toBe("'Open Sans', serif");
+    });
+  });
 });
--- a/src/elements/__test__/styles.spec.js
+++ b/src/elements/__test__/styles.spec.js
@@ -1,5 +1,9 @@
 import ResponsiveStyles, { DEFAULT_MOBILE_BREAKPOINT } from '../styles';
 import { LABEL_TEXT_ALIGN_DEFAULT } from '../utils/labelStyleResolver';
+import {
+  registerUploadedFonts,
+  resetUploadedFonts
+} from '../../utils/uploadedFonts';
 
 const TEST_COLOR_BACKGROUND = 'dddddd';
 const mockElement = {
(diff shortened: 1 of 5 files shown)


FranklinGao123:
Dashboard counterpart repointed: the combined feathery-org/feathery-frontend#3770 was split, and the half that pairs with this PR is now **feathery-org/feathery-frontend#3824** (the `Uploaded--` alias + render-data rewrite). The picker previews/grouping work moved to feathery-org/feathery-frontend#3823, which #3824 is stacked on but which this PR does not depend on.

trebitowski:
Not needed

PR #1772, opened 2026-08-14
feat: document bindings for the DocxEditor

ayeshah:
## Changes

Adds **document bindings** to the DocxEditor: a system that links Syncfusion Word content controls to live-computed values, formulas, and structured tables, so both the assistant and end users can edit bound `.docx` documents safely with tracked changes. Feature is opt-in and gated behind a compile-time switch.

### Binding engine — pure core (`DocxEditor/bindings/core/`)
Framework-free TypeScript core, fully unit-tested in isolation from the editor:
- **`engine.ts`** — the rules engine that evaluates bindings and produces document updates.
- **`sfdtAdapter.ts` / `sfdtTypes.ts`** — read/write Syncfusion SFDT document JSON.
- **`tagDsl.ts`** — the binding-tag DSL parsed from content controls; separates a binding's **value** from its **default**.
- **`formula.ts` / `decimal.ts` / `valueTypes.ts`** — formula evaluation, exact decimal math, and typed values (money, numbers, etc.).
- **`templateImport.ts`** — imports a template's tokens into bindings.
- **`commandRevisions.ts`** — maps binding commands to tracked-change revisions.

### Live editor integration (`DocxEditor/bindings/`)
Wires the pure core to a running Syncfusion editor:
- **`controller.ts` / `editorAdapter.ts` / `attachBindings.ts` / `useDocxBindings.ts`** — attach the engine to a live document and keep bindings and the document in sync.
- **`nativeStructuralAdapter.ts` / `reconcileRegistry.ts` / `rowCommandWatch.ts`** — anchor-driven structural edits (add/duplicate/delete table rows) and reconciliation.
- **`keystrokeGuard.ts` / `contentControlSafety.ts` / `controlGeometry.ts`** — keep the caret inside its content control, protect control contents, and track control geometry.
- **`commitTriggers.ts` / `persistence.ts`** — decide when to commit and persist binding state.

### Assistant integration (`assistant/tools/docx/`)
- Routes assistant docx ops through the binding engine so assistant edits respect bindings.
- Adds a **`duplicate_table`** operation.
- Isolates assistant revision tracking: assistant writes to bound documents are tracked, while **user typing stays untracked** — including inside content controls and after an assistant edit.
- Makes bound-document structural edits **anchor-driven** and gates bound-document writes on real binding tags.

### DocxEditor element wiring
- Opt-in document bindings on the DocxEditor element; assistant and bindings share **one** document instance.
- `FORCE_DOCUMENT_BINDINGS` compile-time switch to force bindings on.

### Notable fixes
- Keep Docx undo/redo on a single Syncfusion history stack; don't re-adopt rows during undo/redo.
- Judge an adoptable row by its **content**, not its position; keep adopting after all bound rows are deleted; fill a new row the moment it's created.
- Stop a colourless content control crashing the render; stamp control colours after layout.
- Restore both scroll axes; wait for `documentChange` before settling the loading state.
- Keep engine errors identifiable after the ES5 downlevel.

### Tests & harness
- Extensive unit and integration specs across the pure core, the live-editor adapters, and the assistant docx ops (bound cell writes, bound row ops, duplicate-table, coexistence, content-control round-trips, caret/undo behaviour, and more).
- Browser harness under `harness/bindings/` that drives the binding engine against a real editor.

- [x] Tested end to end

## Related pull requests

- #1771 — feat(assistant): route docx ops through the binding engine and add duplicate_table (merged into this branch)


src/elements/components/DocxEditor/bindings/core/engine.ts
```diff
@@ -0,0 +1,715 @@
...
+    const node = nodes.get(id);
+    if (!node || broken.has(id)) continue;
+    try {
+      const raw = evalAst(node.ast, node);
+      if (Array.isArray(raw))
+        throw new FormulaError('formula produced a column, not a value');
+      const fieldType = node.occ.def.fieldType as { scale?: number };
+      // Round to the type's scale BEFORE storing, so a dependent formula sums
+      // already-rounded values rather than compounding fractions.
+      results.set(id, D.roundTo(raw, fieldType.scale ?? 0));
```
ayeshah: **[High] Percent-typed formula results are rounded to 0 decimals, zeroing most percentages.**

A formula tagged `type=percent` (e.g. `[[name=eff_rate|type=percent|expr=sub(gross,disc)]]`) is numeric, so it evaluates. Its canonical value is a fraction (`0.085` for 8.5%). The result is stored via `D.roundTo(raw, fieldType.scale ?? 0)`, but the percent `FieldType` has no `scale` field, so scale defaults to `0` and `roundTo('0.085', 0) === '0'`. Any percent formula under 50% renders as **0%**. Only integer/currency/decimal carry a `scale`; percent silently loses all precision.

src/elements/components/DocxEditor/bindings/editorAdapter.ts
```diff
@@ -0,0 +1,359 @@
...
+    updateValues(writes: EngineWrite[]): boolean {
+      const helper = editor.documentHelper;
+      const editorModule = editor.editorModule;
+      if (!helper || !editorModule || !editorModule.updateContentControl)
+        return false;
+      pruneDetachedContentControls(editor);
+      const collection = helper.contentControlCollection;
+      if (!Array.isArray(collection)) return false;
+      // Empty text would be replaced by the editor's placeholder string.
+      if (writes.some((write) => !write.text)) return false;
```
ayeshah: **[High] A single empty-text write makes `updateValues` abort the whole patch batch, dropping every other write.**

`if (writes.some(write => !write.text)) return false;` bails on the entire batch if ANY write has empty text. Consider a document-level text field with two occurrences: the user clears it in one place. The engine fans out `''` to the second occurrence (`write.text === ''`), so `updateValues` returns false — the second occurrence is never cleared, AND any other pending field/formula writes in the same reconcile are silently discarded. The controller then records the reconcile as committed (see the `commit()` finding).

src/elements/components/DocxEditor/bindings/controller.ts
```diff
@@ -0,0 +1,547 @@
...
+      }
+    ];
+    this.onChange({ controller: this, event: 'error' });
+  }
+
+  private commit(
+    result: ApplyRulesResult,
+    { apply, markDirty, event }: CommitOptions
+  ): void {
+    this.workingSfdt = result.sfdt;
```
ayeshah: **[High] `commit()` advances `workingSfdt`/`values`/`index` before applying to the editor, so a failed live patch leaves controller state describing content the editor never received.**

`commit()` sets `this.workingSfdt/values/index/rowTemplates = result.*` unconditionally (lines 485-489), then tries to apply. When `apply` is `'patch'` or `'structural'` and `updateValues`/`applyStructuralMutations` returns false (and `apply` is not `'open'`), it only appends a `native-mutation-failed` diagnostic — it does NOT roll back state or fall back to `open()`. The editor still shows the old content while `controller.values` holds the new canonical values. The next flush serializes the editor's stale text and reconciles against those wrong `prevValues`, so change-detection can treat the engine's intended value as a user 'revert' and silently discard the edit.

src/elements/components/DocxEditor/bindings/core/commandRevisions.ts
```diff
@@ -0,0 +1,330 @@
...
+    const superseded = revisionIdsIn(oldNode);
+    const originalText = superseded.size
+      ? textWithoutRevisions(oldNode, revisionIdsOfType(before, 'Insertion'))
+      : previous.text;
+    if (superseded.size)
+      revisions = revisions.filter((revision: any) => {
+        const id = revision?.revisionId ?? revision?.revisionID;
+        return id == null || !superseded.has(String(id));
+      });
+    next = setAt(next, occurrence.path, {
```
ayeshah: **[Medium, lower confidence] Tracked-change authoring writes an `inlines` array onto the occurrence node, which is malformed if that node is a block/cell-level content control (which holds `blocks`, not `inlines`).**

The text-change loop does `setAt(next, occurrence.path, { ...newNode, inlines: [del, ins] })`. `scanBindings` can record an occurrence whose node is a block-CC wrapper or a cell that is itself a content control. For such an occurrence, this adds an `inlines` field it should not have while its `blocks` remain, producing SFDT the editor may not render/serialize correctly for an assistant tracked edit. (Depends on a bound value being carried by a block/cell-level CC rather than an inline CC.)

ayeshah (commented):
Automated code review — 4 findings in the binding logic (most severe first). Findings 2 and 3 share a root cause: a failed patch-apply is not reconciled with the controller's already-advanced state. The pure layer (decimal/formula/tagDsl/valueTypes), caret geometry, row adoption, and template import all held up with no correctness issues.

faizalsomani:
Thanks Ayesha - I went through all four against the source rather than taking the report at face value. **All four are real.** Fixes are in progress now and will land as a single commit on this branch.

Verdict and scope on each:

**1. Percent formula results rounded to zero decimals - confirmed, High.**
`FieldType` really does define `percent` with no `scale`, so `fieldType.scale ?? 0` collapses to `0` and `roundTo('0.085', 0)` gives `'0'`. Every percent-typed formula under 50% renders `0%`.

One refinement to the scope: percent *inputs* are unaffected. `parseDisplay`'s percent branch just multiplies by `0.01` with no rounding, so only percent-typed **formulas** (`expr=`) are broken. The fix has to make a percent formula render identically to the equivalent percent input for the same value, without touching currency/decimal/integer.

**2. One empty-text write aborting the whole batch - confirmed, High.**
`writes.some(write => !write.text)` discards every write in the reconcile, not just the empty one. Your two-occurrence clearing case is exactly right.

The guard's *intent* is legitimate - the editor substitutes its placeholder for empty text, which your comment on that line already says. So we're fixing the granularity, not removing the guard: handle empty text per write and let the rest of the batch through.

**3. `commit()` advancing state before the editor is updated - confirmed, and the most serious of the four.**
State is assigned unconditionally at the top of `commit()`, and a failed `updateValues` / `applyStructuralMutations` only appends a `native-mutation-failed` diagnostic. The controller then holds values the editor never received, and as you say, the next flush can read the engine's intended value as a user revert and drop the edit. That's silent loss of a user's edit with no error surfaced, which is why we're treating it as the priority.

Worth flagging the constraint we're fixing it under, since it's yours and it's deliberate: your header notes that a failed live patch must never reload because `open()` destroys `editorHistoryModule`. We're honouring that - so the fix is a rollback to the pre-commit snapshot plus a real surfaced error, not a fallback to `open()`. That constraint has also been reinforced from our side: the commit we just landed removed the last `open()` from the assistant's op path, because reloading was producing no tracked change, discarding pending revisions, and stranding global tracking on.

**4. `inlines` written onto a block/cell-level content control - confirmed, and this one is ours.**
`commandRevisions.ts` is from the tracked-changes work, so this is our defect, not yours. You flagged it as lower confidence - it holds. In `sfdtAdapter.ts` a table **cell** carrying its own `contentControlProperties` that parses as a field or formula is recorded as an occurrence with the cell as its node, and a cell holds `blocks`. Spreading it and adding `inlines` produces a node with both.

It hasn't shown up in live testing because our test template carries bound values in inline content controls inside cells rather than cell-level ones. We're fixing it to descend to the correct inline-bearing node, and where a shape genuinely can't be authored as a tracked change it will refuse rather than emit malformed SFDT - an untracked or corrupt write is the exact failure mode the last commit was closing.

Each of the four gets a regression test that fails before the fix and passes after.

Also worth surfacing from the last commit, since it changes a rule rather than just fixing a bug: it deliberately replaces the previous "a refused batch must leave the tracking flag untouched" behaviour. That rule was part of how global tracking got stranded on, which meant the user's own typing started being recorded as tracked changes under `Guest user`. Every assistant batch now leaves tracking off on both the editor and its container. The reasoning is in the commit message.

src/elements/components/DocxEditor/bindings/core/engine.ts
faizalsomani: Confirmed and fixed in c81919e0 - percent formulas now keep the canonical fraction the way percent inputs already did, so only formula results were affected and typed percentages were always fine.

src/elements/components/DocxEditor/bindings/editorAdapter.ts
faizalsomani: Confirmed and fixed in c81919e0 - your placeholder guard stays, but it is now applied per write instead of aborting the batch, so clearing one occurrence no longer discards every other pending write.

src/elements/components/DocxEditor/bindings/controller.ts
faizalsomani: Confirmed and fixed in c81919e0 - commit() now rolls the controller back to its pre-commit snapshot and surfaces a real error instead of advancing state, keeping your rule that a failed live patch must never reload and destroy editor history.

src/elements/components/DocxEditor/bindings/core/commandRevisions.ts
faizalsomani: Confirmed, and this one was ours rather than yours - fixed in c81919e0 to descend to the inline-bearing node instead of writing inlines onto a block or cell wrapper, and to refuse outright rather than emit malformed SFDT when no such node exists.

faizalsomani (approved):

PR #1774, opened 2026-08-17
feat: make global docx bindings deterministic

faizalsomani:
## What this PR achieves

A `global=true` binding is now one document-wide identity that survives duplication, and duplicating content that carries bindings actually works. Change a global value once and every occurrence moves together as one reviewable card. Duplicate a bound table or a section and the copy becomes its own independent thing - fresh identities, fresh row ids, formulas rewritten to follow - while any global binding inside it stays shared with the original.

Copying a section that contained a bound table used to **lose content silently**. That is fixed here, and the mechanism that fixed it is the substance of this PR.

## Why it exists

Two failures, both reachable from ordinary use.

**Same-named bindings were written implicitly.** Two independently identified bindings that happened to share a label were treated as one thing, so a write to "the tax rate" silently changed a value the user never named. There was no way to say "these two are deliberately the same" and no way for the engine to ask which one was meant.

**Copying a section lost blocks.** Asked to copy a section containing a bound table, the copy came back missing its heading with the following block truncated - and the operation reported failure while having already mutated the document. Isolated precisely: the trigger is a table wrapped in a **block-level content control** sitting among sibling blocks. Remove only that wrapper and the same copy succeeds; remove content controls from any other block and it still fails. It is independent of target position and of the binding reconcile, and it predates this branch.

## How it does it

- **`global=true` in the tag DSL.** An explicit declaration that several controls share one document-wide identity. Nothing is inferred from matching labels or anchors - equal names are not permission to write them together.
- **One identity rule for every clone.** `cloneIdentityRewrite` is the single place that decides what a copy is called: a global binding keeps its name so the copy joins that identity, a non-global one forks to a fresh unique name carrying the value the user can see, table-scoped bindings are re-scoped through fresh table and row ids, and `del=keep` transfers. Both the table duplicate and the section copy call it, so they cannot drift apart.
- **Copies are built from the document's own SFDT, not captured by SyncFusion.** `rawBlocksInRange` Ã^CÃÂ¢^F^R `clonedWithoutRevisions` Ã^CÃÂ¢^F^R identity rewrite. Building the payload ourselves is what makes it possible to give a copy its own identities before it is pasted, rather than repairing it afterwards.
- **The payload is pasted in wrapper-safe segments.** `copyPasteSegments` splits a clone so each block-level-content-control-wrapped table is pasted alone and runs of other blocks paste together, with a separator paragraph only where the insertion would leave two tables adjacent - Word renders adjacent tables as one. This is the fix for the content loss.
- **The copy is read back and verified.** `assertPastedRangeMatches` compares what landed against what was meant to land and throws on mismatch, rolling the whole change set back. An untested document shape now refuses rather than corrupts.
- **Ambiguity instead of an implicit write.** Where one identity resolves to several independent instances, the engine returns a no-write ambiguity payload listing every instance with its value and location. The user chooses all or one, and the confirmed choice is applied as exactly one operation validated against a live-document fingerprint, so a stale confirmation is refused rather than applied to a document that moved.
- **`duplicate_table` uses the same three primitives**, and the payload-string path it replaced is deleted outright.
- **Conflicting writes to one global identity are refused atomically**, before anything is written, rather than resolved by ordering.

## Files changed and why

| File | Why |
| --- | --- |
| `bindings/core/tagDsl.ts` | `global=true` in the tag grammar, parsed and serialized |
| `assistant/tools/docx/bindingWriteContract.ts` | the typed frontend/ai-services wire contract for identity, ambiguity occurrences and all-or-one resolution - this repo owns these shapes |
| `assistant/tools/docx/syncfusionDocumentOps.ts` | the engine work: identity rule, SFDT clone, segmented paste, read-back assertion, ambiguity payload, `duplicate_table` migration, dead path deleted |
| `assistant/capabilities/registry.ts` | declares the ops and params ai-services mirrors |
| `bindings/core/sfdtAdapter.ts` | binding discovery reads the new identity |
| `assistant/tools/docx/tests/copyFidelity.spec.ts` | new: pins copy and duplicate behaviour across the content matrix - plain paragraphs, bound prose, global and non-global, N tables, mixtures |
| `boundDuplicateTable.spec.ts`, `boundDocumentCoexistence.spec.ts`, `multiTable.spec.ts`, `tagDsl.spec.ts`, `valueTypes.spec.ts`, `registry.spec.ts`, `costsFixture.ts` | coverage for global identity, fork-on-copy, formula rewriting and the wire shape |

## Known limitation, addressed in the next PR

**Duplicating a bare paragraph that carries a content control does not carry the control over.** The paragraph is copied, but the copy is inert text rather than a bound value.

The reason is that table duplication and paragraph duplication are still different code paths, and only the table path understands bindings. This PR makes the table and section paths share one mechanism; the bare-paragraph case is the remaining gap.

The next PR closes it by removing the distinction entirely: one range operation that takes anchors and duplicates whatever lies between them - paragraphs, tables, or any mixture - with the same identity rule applied uniformly. The shape of the content stops being something the caller has to name.

## Validation

- Full docx suite green: 44 suites, 1138 tests. `yarn typecheck` (`tsconfig.typecheck.json`) clean; `yarn build` clean - the build is the strictest gate and catches errors the other two miss.
- Behaviour verified by reading back the resulting binding tags, not by trusting the op result: non-global copies fork (`project.name` Ã^CÃÂ¢^F^R `project.name_2`, `costs` Ã^CÃÂ¢^F^R `costs_copy` with fresh row ids, `mul(costs_subtotal_2, tax_rate_2)`), global copies share (`tax_rate` unchanged at three occurrences, the copied formula still referencing it), and a formula referencing a binding outside the copied range correctly keeps pointing outside.
- Three adversarial cases tested rather than reasoned about: multi-segment paste past the last block, a paragraph copy whose following block is a table, and two bound tables in one range. None regressed.
- Manually exercised in the browser against a real bound template: section copy with a wrapped table, table duplication, and global propagation across two tables.

## Review round

Four issues Ayesha raised in review, each with a test that fails without the fix:

- **Expanded vs raw block indices** - `clonedRangeBlocks` took anchors carrying expanded block indices into `rawBlocksInRange`, which indexes raw section blocks, so a wrapper holding several children shifted the copied range and the validation compared against that wrong clone. Expanded addresses are now translated to raw indices before cloning.
- **Duplication copying a container's siblings** - `duplicate_table` cloned `container.block`, which can be a content control wrapping the table together with a caption or note, and the read-back only ever verified the table. One rule now decides whether a container carries only its table, used by both duplication routes.
- **Unreserved generated ids** - copy ids were checked against source ids but not against each other, so `costs-us` and `costs_us` both allocated `costs_us_copy` and two copies merged under one id. Ids are reserved as they are allocated.
- **A section copy leaving its binding family** - `documentFieldIdentity` recognised only the table-prefixed form, so `project.name_2` and its source read as unrelated fields and the ambiguity flow never fired for the case that creates two instances. The numeric copy suffix is now stripped alongside the table prefix.

## One more defect, found while testing those cases

A paragraph can hold any number of inline content controls, so a template writes prose like `Effective [policy.date], quoted at [tax_rate] tax.` The reader collected every binding range and then reported only the first, leaving the second bound value invisible to the model and therefore unwritable: asked to change the tax rate, the only op left was a plain text replace, which the engine refuses because it would write across a control it cannot see - and the refusal named `policy.date` for a tax edit, because routing resolved the block's first tag rather than the binding the write targeted.

The complete list was already computed in `bindingRanges`; only the projection and the router discarded it. Inventory entries now carry `bindings` as a list, and a write names which one it means with `field` - identity rather than position, since offsets inside bound blocks are already untrusted. Disambiguation is scoped to bound writes, so formatting ops are unaffected. The matching `field` parameter and prompt guidance ship in ai-services #694.

## Review-round validation

Verified live in a bound template built for these cases (5 layout sections, 2 tables with deliberately colliding ids, global and non-global fields), reading back the document rather than trusting the op results:

- Duplicating a table whose wrapper also holds a caption: tables 2 ÃÂ¢^F^R 3, caption still at one occurrence.
- Duplicating both colliding-id tables: `costs-us`, `costs_us_copy`, `costs_us`, `costs_us_copy_2` - four distinct ids.
- Copying a section carrying both kinds of binding: `client.name_2` and `client.contact_2` forked while `tax_rate` stayed one global identity across all three occurrences; the ambiguity prompt offered all-or-one, and choosing all updated 3 of 3.
- Changing the tax rate: one `set_cell_text` with `field:"tax_rate"` updated the global identity across both occurrences as tracked insertions, leaving `policy.date` in the same sentence untouched.
- Appending to that two-control sentence: text landed at the paragraph end with both bindings intact.

## A binding in a header was never a binding at all

Found while building a header case for the conservation work, then confirmed at the source: `convertTemplateTokens` walked `sections[].blocks` and never `headersFooters`. A template that puts a binding in its header - which a proposal does, because it names its client at the top of every page - had that token survive as literal text, so the client read the tag language itself, brackets and default value included, on every page of their deliverable.

The damage was not only cosmetic. With no content control there was no binding occurrence in the header, so nothing fanned a new value into it and a document could display two different values for one field. And `assertNoHeaderOccurrence` - the guard that exists precisely to refuse a bound write whose value also lives in a header - had no header occurrence to find, so it was dead on the path most templates take. A guard that cannot fire reads as protection while protecting nothing.

One conversion rule now applies to every story in a section rather than the body alone.

**Behaviour change reviewers should weigh.** That guard is now live: a bound write to a value that also appears in a header refuses with `bound_value_also_in_header`, where it previously succeeded and left the header showing the old value. Both post-fix behaviours are more honest than before - the header now shows the bound value rather than raw DSL, and a write that cannot be completed everywhere says so instead of half-landing - but it is a real limit on templates that bind one value in both body and header. The guard's own rationale is that a value the user corrects by hand is recoverable, while a deliverable silently reading "P-1000P-2000" is not.

---

## Write-integrity fixes

While porting the op engine I went looking for writes that change a document in
ways the reviewer cannot see or undo. Seven landed here. They are separable from
the port and small, so they go in this PR rather than waiting for it.

The property they defend: **every change this engine makes must be visible as a
tracked change and reversible by rejecting it.** A write that slips past that is
worse than a failed write, because the document looks fine and nobody knows to
look.

### What each one does, and whether it is a refusal or a behaviour change

Stated plainly, because "seven fixes" hides a real capability trade.

| # | Change | Kind |
| --- | --- | --- |
| 1 | `insert_hyperlink` / `insert_page_number` no longer destroy the text they are anchored to | **behaviour change** - they now insert at a caret instead of consuming a block selection |
| 2 | `insert_section_break` selects an explicit caret rather than a block | **behaviour change**, latent defect, same shape as #1 |
| 3 | A page break aimed inside a table cell is refused before it is written | **refusal** |
| 4 | Duplicating the document's tail table is refused before pasting | **refusal** |
| 5 | `split_table` on a table carrying bindings is refused | **refusal** - the largest trade; see below |
| 6 | Overwriting text that carries somebody else's pending tracked change is refused | **refusal** |
| 7 | An empty `replace_text` actually deletes instead of verifying as a no-op | **behaviour change** |

Four refusals and three behaviour changes. The refusals all write nothing and
say so; each one's test asserts the document is byte-unchanged afterwards,
because a refusal that changed the document is a lie this engine has told before.

### The capability trade in #5, measured

`split_table` on a bound table used to succeed. Measured over the shape corpus,
it succeeded on 17/17 shapes and destroyed binding tags on 13 of them. It now
refuses on those 13 and works on 4.

Demonstrated end to end in the browser, through the real assistant chat, on a
bound corpus document - same request, same answer to the assistant's own
clarifying question, engine the only variable:

| | before | after |
| --- | --- | --- |
| assistant says | "Done - the first table is now split..." | "I couldn't split that table because it contains linked values, and splitting it would break them. Nothing changed." |
| tables | 2 -> 3 | 2 -> 2 |
| **binding tags** | **11 -> 1** | **11 -> 11** |
| revisions afterwards | 0 | 0 (nothing was written) |

Two things make the old behaviour worse than an ordinary bad edit: while the
change is still pending all 11 tags are present, so a reviewer reading the
tracked changes sees nothing wrong - the destruction lands on ACCEPT. And
afterwards there are no revisions left, so nothing can be rejected; the bound
values keep rendering their last computed figure and never recompute.

The capability comes back with the `split_table` recomposition
(copy rows + delete rows, under the relocation law), which is the first slice PR.
The refusal is the interim state, not the destination.

### Tests

Five of the seven shipped without a tracked test. All seven now have one, and
each guard is proven TWICE: the defect refuses and leaves the document
byte-unchanged, AND the legitimate case next door still succeeds. A guard that
over-refuses is a regression wearing a safety jacket, and only the second half
catches that.

- `writeIntegrityGuards.spec.ts` - the seven guards, 7 tests
- `additiveOpsKeepText.spec.ts` - four cases covering the ops fixed in #1/#2,
  hand-listed. NOT registry-derived on this branch: the `contentEffect`-driven
  version, which would cover a future additive op without anyone remembering,
  lives on the port branch. Worth carrying back here later; stated so nobody
  reads this as stronger coverage than it is.
- `foreignRevisionRange.spec.ts` - the #6 guard's range narrowing, 5 tests
- `revisionAuthorship.spec.ts` - every revision this engine creates is the
  assistant's, and no pre-existing foreign revision changes hands, 5 tests

### Two more defects found while writing those tests

**Fixed here:** the #6 guard could never fire on a table cell.
`expandBlockContentControls` yields `{block, insideControl}` pairs and the anchor
resolver never unwrapped them; for a body paragraph the caller reached `.block`
by accident while walking for revision ids, but for a table anchor row
resolution collapsed to undefined and the guard found nothing. `set_cell_text`
was unguarded.

**Found, not fixed:** `liveText` is a corrupted hybrid when a block carries a
pending deletion - the deletion-included text truncated to the length of the
deletion-excluded text. Measured: "Alpha beta gamma delta." with "beta " pending
-deleted yields `"Alpha beta gamma d"`, which matches neither projection, so
every index resolved in it is wrong. This is pre-existing and belongs to whoever
owns that projection; it is NOT fixed here. The #6 guard is made safe against it
by discarding the range and falling back to block granularity whenever the block
carries a foreign pending deletion - it over-refuses deliberately rather than
deciding what is safe to overwrite from known-wrong offsets.

### Exact contract, narrower than it sounds

The no-op short-circuit runs before the #6 guard, so: **the guard protects every
write that CHANGES the document**, not every write. A write that changes nothing
cannot harm a foreign revision, and guarding no-ops would manufacture
over-refusals.

### Known open, not fixed here

- **Banding does not restart on the new table after a split.** Measured
  identically on both engines, so it is pre-existing and unchanged by this PR:
  the new table keeps the parent's parity, so its first body row is shaded where
  it should be clear. An appearance concern; it belongs to the `split_table`
  recomposition's acceptance rows.
- **A split leaves the two tables with no separating paragraph**, which Word
  renders as one fused table. Now REPRODUCED DETERMINISTICALLY with a failing
  test: the new table lands at a block index adjacent to the original and no
  paragraph is created between them. This corrects an earlier characterisation of
  it as "split-point dependent" - it is not; the split that appeared to leave a
  gap simply targeted a position further from the table.
- **`liveText` corruption** above.
- **The abandon-to-block-wide policy is inconsistent for a nested run.**
  `structuralRevisionIdsAtAnchor` iterates only top-level runs, so a tracked
  FORMATTING revision on a run nested inside a content control is judged by
  range instead of abandoning the narrowing the way an unnested one does. NO
  PROTECTION IS LOST - both the range walk and the block-wide path see that
  revision - it is only the policy that differs between the two cases. Noted
  rather than fixed tonight, because changing it is a behaviour change on a path
  the tests do not yet cover, and this PR is not the place to make one
  unmeasured.
- **`move_section` refuses and writes anyway.** Already a known item, not a new
  discovery - it is tracked as "move_section anchored at a NON-HEADING destroys
  binding tags through a refusal". What the boundary sweep added while gating
  this series is the MEASUREMENT, and confirmation on the committed engine that
  it is pre-existing and untouched by this PR. On `mirrored-bindings`, at two
  different boundaries, `move_section` returns the `untracked_write` refusal AND
  changes the document: 176 -> 470 chars, 7 -> 6 top blocks, 0 -> 3 revisions,
  one binding tag destroyed; and 176 -> 596 chars, 0 -> 43 revisions, two tags
  destroyed. That is the same defect class this PR fixes - a refusal that is not
  a refusal - on an op outside its scope. It should be the next one fixed.
  Note the sweep that caught it is untracked and gitignored (`*.local.spec.ts`),
  so CI never runs it; that is its own problem and is why this went unseen.
- **The #6 range narrowing is proven at the engine boundary by tests, not through
  the chat.** A browser attempt to write words carrying no revision, in a line
  carrying one, was still refused at line granularity - the assistant's chosen op
  for that phrasing does not reach the guard with a narrow range. No user-visible
  improvement is claimed for it.

## Design note: why five split_table tests changed

Two captain instructions bear on this, and the later one narrows the earlier.

**Earlier:** "Here it should NOT create a new subsection heading, just a new
table." That is about a HEADING - no title, no new section. The test written
from it extended the rule to forbid ANY paragraph.

**Later, 2026-08-27:** "if tables are split then there has to be like a space
between them right they should not be just attached to each other."

Word renders two flush tables as ONE table, so a split that leaves them adjacent
is undone at render time - the document the person sees is not the document the
engine wrote. The existing suite did not merely tolerate that; it encoded it,
in a test named "a second split of a table the first split left adjacent stays
clean" whose own comment reads "the two tables are now neighbours with nothing
between them".

The resolution: a split now leaves ONE EMPTY paragraph, and only where the two
tables would otherwise land flush. An empty separator is neither a title nor
content; it is what makes two tables be two tables. So the acceptance criterion
is now "no new heading and no content paragraph", and the test asserts the added
paragraph IS empty rather than merely counting it.

The mechanism was not invented for this. `copyPasteSegments` already inserted a
separator wherever a CLONED table would land flush against another - the rule
was right and its home was wrong, living in one caller's helper rather than at
the seam every relocation pastes through. It now sits at that seam, and the
separator is prepended to the PAYLOAD so it arrives in the same paste: one card,
one undo entry, and rejecting the split takes the separator with it. A separator
written as a separate edit would have survived the rejection of the split it
exists to serve.

Four other tests changed only because block indices after a split shift by one.
Their comments record that, so the next reader sees a consequence rather than a
mystery.

### DIFF
--- a/src/assistant/tools/docx/syncfusionDocumentOps.ts
+++ b/src/assistant/tools/docx/syncfusionDocumentOps.ts
@@ -1168,20 +1168,29 @@ function getRows(block: any): any[] | undefined {
  * Anchors are `section;block` for body paragraphs and
  * `section;block;row;cell;para` inside a table.
  */
-function revisionIdsAtAnchor(sfdt: any, anchor: string): Set<string> {
-  const found = new Set<string>();
+function resolveAnchoredNode(sfdt: any, anchor: string): any {
   const parts = String(anchor).split(';');
   const sections: any[] = pick(sfdt, 'sections', 'sec') ?? [];
   const section = sections[Number(parts[0])];
-  if (!section) return found;
+  if (!section) return undefined;
   // Anchors count EXPANDED blocks: a block-level content control holding N
   // blocks contributes N entries. Raw indices diverge from anchors the moment
   // such a control holds anything but exactly one block - which is precisely
   // what a bound document has - so resolve the same way `tableContainerAt` does.
-  let block: any = expandBlockContentControls(getBlocks(section))[
+  //
+  // UNWRAP. expandBlockContentControls yields `{ block, insideControl }` pairs,
+  // not raw blocks. The previous caller never unwrapped and got away with it
+  // only because it walked every value looking for revision ids, so it reached
+  // `.block` by accident. A TABLE anchor did not get away with it: getRows on
+  // the wrapper finds nothing, cell resolution collapses to undefined, and the
+  // foreign-revision guard silently found NOTHING for every table-cell anchor -
+  // set_cell_text was unguarded. Unwrapping here fixes that hole too.
+  const entry = expandBlockContentControls(getBlocks(section))[
     Number(parts[1])
   ];
-  if (!block) return found;
+  if (!entry) return undefined;
+  let block: any = entry.block;
+  if (!block) return undefined;
   if (parts.length >= 5) {
     const rows = getRows(block) ?? getRows(getBlocks(block).find((b: any) => getRows(b)));
     const row = rows?.[Number(parts[2])];
@@ -1189,6 +1198,14 @@ function revisionIdsAtAnchor(sfdt: any, anchor: string): Set<string> {
     const cell = Array.isArray(cells) ? cells[Number(parts[3])] : undefined;
     block = getBlocks(cell)[Number(parts[4])] ?? cell;
   }
+  return block;
+}
+
+/** Every revision id anywhere in an anchored block, at any depth. */
+function revisionIdsAtAnchor(sfdt: any, anchor: string): Set<string> {
+  const found = new Set<string>();
+  const block = resolveAnchoredNode(sfdt, anchor);
+  if (!block) return found;
   const collect = (node: any): void => {
     if (!node || typeof node !== 'object') return;
     if (Array.isArray(node)) return node.forEach(collect);
@@ -1197,6 +1214,142 @@ function revisionIdsAtAnchor(sfdt: any, anchor: string): Set<string> {
     for (const value of Object.values(node)) collect(value);
   };
   collect(block);
+  return found;
+}
+
+/**
+ * Where each revision sits, in the SAME offset space `find` resolves against.
+ *
+ * Deliberately the structural twin of `bindingRangesOf` below: same walk, same
+ * skip rule, same offset counter. The two must agree, because both map a
+ * character range onto the live text that `inlineText` projects - and an offset
+ * walk that drifts from the one the caller used is a guard that fires on the
+ * wrong text.
+ *
+ * ONE DELIBERATE DIFFERENCE. `bindingRangesOf` drops an inline whose revisions
+ * are all deletions, because deleted text is not part of the live projection.
+ * This must NOT drop it: a pending deletion is the single most dangerous thing
+ * in the paragraph, since SyncFusion re-authors it under the current user
+ * rather than leaving it alone. It occupies no live width, so it is recorded as
+ * a ZERO-WIDTH marker sitting at the current offset.
+ */
+function revisionRangesOf(
+  inlines: any[],
+  deletedIds: Set<string>
+): Array<{ id: string; start: number; end: number }> {
+  const ranges: Array<{ id: string; start: number; end: number }> = [];
+  let offset = 0;
+  const walk = (items: any[]): void => {
+    for (const inline of items) {
+      if (inline == null) continue;
+      // Two places, not one. A run carries its content revisions directly, and
+      // a TRACKED FORMATTING change is recorded on the run's own
+      // characterFormat. Reading only the first missed the second entirely, so
+      // a foreign tracked formatting change on a run was narrowed straight past
+      // - the same lost-protection shape as the paragraph mark.
+      const raw = pick(inline, 'revisionIds', 'rids');
+      const formatRaw = pick(
+        pick(inline, 'characterFormat', 'cf'),
+        'revisionIds',
+        'rids'
+      );
+      const ids = [
+        ...(Array.isArray(raw) ? raw.map(String) : []),
+        ...(Array.isArray(formatRaw) ? formatRaw.map(String) : [])
+      ];
+      const start = offset;
+      if (ids.length > 0 && ids.every((id) => deletedIds.has(id))) {
+        // CURRENTLY UNREACHABLE for a FOREIGN pending deletion: the guard
+        // detects that case earlier and abandons the narrowing altogether, so
+        // this branch never decides anything for one. It is kept, and kept
+        // correct, because it still runs for the assistant's OWN pending
+        // deletions (which are not in `preExisting` and so are filtered out
+        // downstream), and because it is what the branch must do the day the
+        // liveText projection is fixed and narrowing over deletions becomes
+        // safe again. Deleting it would quietly remove the offset arithmetic
+        // that restoration depends on.
+        for (const id of ids) ranges.push({ id, start, end: start });
+        continue;
+      }
+      const nested = getInlines(inline);
+      if (nested.length) {
+        walk(nested);
+        for (const id of ids) ranges.push({ id, start, end: offset });
+        continue;
+      }
+      const text = pick(inline, 'text', 'tlp');
+      if (typeof text === 'string') offset += text.length;
+      for (const id of ids) ranges.push({ id, start, end: offset });
+    }
+  };
+  walk(inlines);
+  return ranges;
+}
+
+/**
+ * Does a revision's span touch the half-open range [start, end) being written?
+ *
+ * A zero-width marker (a pending deletion) counts when it sits anywhere from
+ * the start boundary to the end boundary INCLUSIVE. That is deliberately
+ * generous: a deletion sitting exactly where the write begins or ends is inside
+ * the text SyncFusion's selected-range delete will walk over, and a guard whose
+ * job is to prevent an unrecoverable re-authoring should err toward refusing.
+ */
+function revisionTouchesRange(
+  revision: { start: number; end: number },
+  start: number,
+  end: number
+): boolean {
+  return revision.start === revision.end
+    ? revision.start >= start && revision.start <= end
+    : revision.start < end && revision.end > start;
+}
+
+/**
+ * Revision ids the character-range walk CANNOT address, at an anchor.
+ *
+ * Two of them, and both were invisible to the narrowing until review caught it:
+ *
+ *   PARAGRAPH MARK  `characterFormat.revisionIds` on the block itself. Deleting
+ *                   a paragraph mark is how a paragraph gets merged into the
+ *                   next one. It is not in any run, so walking inlines never
+ *                   sees it.
+ *   ROW FORMAT      `rowFormat.revisionIds` on a table row - a whole inserted or
+ *                   deleted row. Worse than the paragraph mark: it hangs off the
+ *                   ROW, above the cell's paragraph, so even the block-wide
+ *                   recursion from a cell anchor cannot reach it.
+ *
+ * Neither has a character range, so there is nothing to intersect a write
+ * against - which is exactly why the answer is to stop narrowing rather than to
+ * invent an offset for them.
+ */
+function structuralRevisionIdsAtAnchor(sfdt: any, anchor: string): Set<string> {
+  const found = new Set<string>();
+  const add = (ids: unknown): void => {
+    if (Array.isArray(ids)) for (const id of ids) found.add(String(id));
+  };
+  const node = resolveAnchoredNode(sfdt, anchor);
+  add(paragraphMarkRevisionIds(node));
+  // A run's own characterFormat revision ids - a tracked FORMATTING change.
+  // Included here as well as in the range walk: a formatting revision covers a
+  // run whose boundaries need not line up with the text being written, so the
+  // safe reading is to treat it as structural and stop narrowing.
+  for (const inline of getInlines(node))
+    add(pick(pick(inline, 'characterFormat', 'cf'), 'revisionIds', 'rids'));
+  const parts = String(anchor).split(';');
+  if (parts.length >= 5) {
+    const sections: any[] = pick(sfdt, 'sections', 'sec') ?? [];
+    const section = sections[Number(parts[0])];
+    const entry = section
+      ? expandBlockContentControls(getBlocks(section))[Number(parts[1])]
+      : undefined;
+    const table = entry?.block;
+    const rows = table
+      ? getRows(table) ??
+        getRows(getBlocks(table).find((b: any) => getRows(b)))
+      : undefined;
+    add(rowRevisionIds(rows?.[Number(parts[2])]));
+  }
   return found;
 }
 
@@ -1220,7 +1373,13 @@ const PRE_EXISTING_REVISIONS_KEY = '__fmPreExistingRevisionIds';
 function assertNoForeignPendingRevisions(
   editor: LiveEditor,
   block: FlatBlock,
-  op: EditOp
+  op: EditOp,
+  // The character range this op will actually overwrite, in the same offset
+  // space `find` was resolved in. Omit it when the op rewrites the WHOLE block
+  // (set_cell_text, change_case, a replace_text with no `find`) - then every
+  // revision in the block genuinely is in range and block granularity is not a
+  // widening, it is the truth.
+  range?: { start: number; end: number }
 ): void {
   const preExisting: Set<string> | undefined = (editor as any)[
     PRE_EXISTING_REVISIONS_KEY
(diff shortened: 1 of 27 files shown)


src/assistant/tools/docx/syncfusionDocumentOps.ts
```diff
@@ -7531,6 +7551,187 @@ function resolveRelocationTarget(
   };
 }
 
+/** A clone of the range's raw blocks, resolved to what the user currently sees. */
+function clonedRangeBlocks(sfdt: any, range: BlockRange): any[] {
+  return clonedWithoutRevisions(sfdt, rawBlocksInRange(sfdt, range));
```
ayeshah, line 8224: These anchors use expanded block indices, but `rawBlocksInRange` indexes raw section blocks. A wrapper with multiple children makes `copy_section` copy the wrong or extra blocks, and the validation compares against that wrong clone. Translate expanded addresses back to raw blocks before cloning.

src/assistant/tools/docx/syncfusionDocumentOps.ts
```diff
@@ -8399,26 +8575,22 @@ export const ANCHORED_OP_HANDLERS: {
...
+    // paragraph's start. Unlike a model-supplied relocation target, it is
+    // derived from the already-resolved source and cannot land inside its
+    // cells. The block preceding it is the source table itself, so the
+    // segment builder's adjacency rule supplies the separator paragraph that
+    // keeps Word from rendering source and copy as one table.
+    const target: PasteTarget = {
+      anchor: source.endAnchor,
+      address: { ...sourceAddress, block: sourceAddress.block + 1 }
+    };
+    const clone = clonedWithoutRevisions(sfdt, container.block);
```
ayeshah: `container.block` can be an outer content control containing sibling blocks. This duplicates those siblings too, while the assertion only verifies the table. Clone only the addressed table and its required wrapper.

src/assistant/tools/docx/syncfusionDocumentOps.ts
```diff
@@ -12746,6 +13116,140 @@ function duplicateRowLiteralNumbers(
...
+function rewriteCloneIdentities(
+  cloned: any[],
+  sourceIndex: BindingIndex
+): void {
+  const { bindings, tableIds: clonedTableIds } = clonedBindingTags(cloned);
+  if (!bindings.length && !clonedTableIds.size) return;
+  const tableIds = new Map<string, string>();
+  const rowIds = new Map<string, string>();
+  for (const oldTableId of clonedTableIds) {
+    const newTableId = uniqueTableId(oldTableId, sourceIndex);
```
ayeshah: Generated IDs are not reserved during this loop. For example, `costs-us` and `costs_us` can both allocate `costs_us_copy`, merging two copied tables under one ID. Track newly allocated IDs as well as source IDs.

src/assistant/tools/docx/syncfusionDocumentOps.ts
```diff
@@ -12746,6 +13116,140 @@ function duplicateRowLiteralNumbers(
...
+    tableIds.set(oldTableId, newTableId);
+    const entry = sourceIndex.tables.get(oldTableId);
+    if (entry)
+      for (const [oldRowId, newRowId] of freshRowIdsFor(entry, newTableId))
+        rowIds.set(oldRowId, newRowId);
+  }
+  const { renameDocBindings, preservedDocBindings } = cloneIdentityRewrite(
+    sourceIndex,
+    bindings,
+    (name, used) => uniqueBindingName(name, used)
```
ayeshah, line 14085: A section copy becomes `project.name_2`, but `documentFieldIdentity` only recognizes table-prefixed copies. The source and copy then skip the new ambiguity flow. Preserve their binding-family relationship when allocating the new name.

ayeshah (commented):
A few correctness gaps remain in the new copy paths.

faizalsomani:
Thanks Ayesha - all four are fixed and verified end to end in a bound template (5 sections, 2 tables, global and non-global fields).

- **Expanded vs raw block indices** (`clonedRangeBlocks`): fixed in fe075dfd - expanded addresses are translated to raw block indices before cloning, so a wrapper with multiple children no longer shifts the copied range.
- **`container.block` duplicating siblings**: fixed in 48a0ac07 - duplication clones only the addressed table and the wrapper it actually needs, so a caption sharing the container is left alone.
- **Unreserved generated ids**: fixed in fe075dfd - ids are reserved as they are allocated, so `costs-us` and `costs_us` now yield `costs_us_copy` and `costs_us_copy_2` instead of merging.
- **`project.name_2` skipping the ambiguity flow**: fixed in 8ec1566f - `documentFieldIdentity` strips the numeric copy suffix as well as the table prefix, so a section copy stays in its source's binding family.

One more defect surfaced while testing your cases, fixed in dd78db43: a paragraph can hold several inline content controls, but the reader reported only the first, so a second bound value was invisible to the model and unwritable - asking to change `tax_rate` in "Effective [policy.date], quoted at [tax_rate] tax." was refused, naming `policy.date`. Inventory entries now carry `bindings` as a list and a write names which one with `field`.

Live verification: duplicating a captioned table left the caption at one occurrence; duplicating both colliding-id tables produced four distinct ids; copying a section produced `client.name_2`/`client.contact_2` while `tax_rate` stayed one global identity across all three occurrences, and the ambiguity prompt correctly offered all-or-one; appending to the two-control sentence landed cleanly with both bindings intact.

src/assistant/tools/docx/syncfusionDocumentOps.ts
faizalsomani: Fixed in fe075dfd: expanded addresses are translated to raw block indices before cloning, so a wrapper with multiple children no longer shifts the copied range.

src/assistant/tools/docx/syncfusionDocumentOps.ts
faizalsomani: Fixed in 48a0ac07: duplication clones only the addressed table and the wrapper it needs, so a caption sharing the container is left alone.

src/assistant/tools/docx/syncfusionDocumentOps.ts
faizalsomani: Fixed in fe075dfd: ids are reserved as they are allocated, so `costs-us` and `costs_us` now yield `costs_us_copy` and `costs_us_copy_2` instead of merging.

src/assistant/tools/docx/syncfusionDocumentOps.ts
faizalsomani: Fixed in 8ec1566f: `documentFieldIdentity` strips the numeric copy suffix as well as the table prefix, so `project.name_2` stays in its source's binding family and the ambiguity flow fires.

faizalsomani: Follow-up on this one: stripping `_<digits>` assumed that spelling is reserved to the allocator, but an author can write `revenue_2024`, which then claimed membership in `revenue`'s family and would have been fanned a write on a "choice: all" confirmation - so provenance is now recorded at copy time as `copyOf=<source>` in the tag rather than reconstructed from the name (e33581cf), with the behaviour you asked for unchanged.

src/elements/components/DocxEditor/bindings/core/tagDsl.ts
```diff
@@ -62,6 +69,17 @@ export interface TagOptions {
...
   row?: string;
+  /**
+   * The field this binding was copied FROM, recorded when the copy is made.
+   *
+   * Whether a binding is a copy is a fact known at that moment. Reconstructing
+   * it afterwards from the shape of the name can only guess - a template author
+   * may write `revenue_2024` for reasons that have nothing to do with copying -
+   * and a wrong guess joins two unrelated fields into one family.
+   *
+   * A copy of a copy keeps pointing at the original, so a family has one root.
```
ayeshah, line 80: Comments are a bit verbose

ayeshah (approved):

PR #1785, opened 2026-08-19
fix: run field-change rules for every field an address autocomplete changes

piran777:
## Changes

Selecting an address autocomplete suggestion writes line 1, city, state, country and zip in one go, but only ever dispatched a single change event naming the address-line servar. canRunAction matches change rules on trigger._servarId, so a rule bound to the country field never ran - even though the country had just changed. Picking a US address after selecting a country without states left a country-driven show/hide rule unfired.

Report every servar whose value actually moved as trigger._relatedServarIds and match change rules against all of them. Same for the country field's clear-on-change, which blanks state/city/zip.

Also hide a state field whose country has no states rather than rendering an empty dropdown, so the common case needs no rule at all. Opt out via the field's hide_without_states setting.

- [x] Tested end to end

## Related pull requests

be: https://github.com/feathery-org/feathery-backend/pull/3673
fe: https://github.com/feathery-org/feathery-frontend/pull/3852


src/utils/addressState.ts
```diff
@@ -0,0 +1,60 @@
...
+      const aNext = a.position[aMatching];
+      const bNext = b.position[bMatching];
+      const elNext = stateElement.position[aMatching];
+      return Math.abs(elNext - aNext) > Math.abs(elNext - bNext) ? 1 : -1;
+    })[0];
+  if (!field) return '';
+
+  let value = fieldValues[field.servar.key] as string | string[];
+  // Hacky patch for repeating country fields
+  // TODO: fix
```
trebitowski: if both country and state are repeating, can we match their indexes, instead of using the first?

src/utils/addressState.ts
```diff
@@ -40,9 +40,8 @@ export const getControllingCountryCode = (
     : findCountryByID(value, 'name')?.countryCode ?? '';
 };
 
-// A state field whose country has no states in our data has nothing to offer -
-// an empty dropdown, which a required field can never satisfy. Resolves the
-// country the way DropdownField does, so the two never disagree.
+// Resolves the country the way DropdownField does, so the two never disagree
+// about whether the dropdown has anything in it.
 export const stateFieldHasNoOptions = (
```
trebitowski, line 46: What is the difference between `stateFieldHasNoOptions` and `stateFieldIsOptionless`?

piran777: removing toggle and dropping stateFieldHasNoOptions (setting check)

trebitowski (approved):

PR #1794, opened 2026-08-21
feat: Support column sizing (Equal columns) in TableElement

freeman-zhang:
## What
Renders the Table element's new **Column Sizing** style. When `element.styles.column_sizing === 'equal'`, the table uses `table-layout: fixed` with an unsized `<colgroup>` so the data columns share the width evenly (e.g. 4 columns → 25% each); auxiliary action/delete columns keep a narrow fixed width. Any other value (including unset = `'auto'`) keeps the native content-driven auto layout. Not applied in the transposed layout.

## Companion PRs
- **feathery-frontend** `feat/table-column-sizing` — the Column Sizing (Auto/Equal) control that sets `column_sizing`.
- **feathery-backend** `feat/table-column-sizing` — theme default + backfill for `column_sizing`.

## Verification
- `TableElement.spec.tsx` — 7/7 pass (equal → colgroup; auto → no colgroup).
- eslint + typecheck clean.


src/elements/basic/TableElement/index.tsx
```diff
@@ -203,6 +203,11 @@ function TableElement({
   const showEmptyState = !hasData || !hasSearchResults;
   const showToolbar = enableSearch || showAddRow;
 
+  // 'equal' uses a fixed layout so data columns share the width evenly; any
+  // other value keeps the content-driven auto layout. Not supported transposed.
+  const useFixedColumns =
```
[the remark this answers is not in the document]
freeman-zhang: Addressed.

- **Both viewports through the responsive style path** — `tableLayout` is derived via `styles.apply('table', 'column_sizing', ...)`, so desktop comes from `styles` and mobile from `mobile_styles` as a media query. Desktop Auto + mobile Equal now enables fixed layout on mobile, and desktop Equal + mobile Auto resets to auto at the breakpoint.
- **Colgroup widths too** — this was still outstanding: the action/delete `<col>` widths were inline `style`, and a `<col>` width applies under `table-layout: auto` as well (measured: 80px vs 41px content width), so Desktop Auto + mobile Equal pinned the desktop action column. They now go through the same per-viewport path (`utilityColStyle`).
- **Transposed gate preserved** — kept, but moved inside the getter rather than wrapping the `apply`. `isTransposed` derives from the row count, so it flips after first render; guarding the call left a stale `tableLayout: fixed` on the memoized style target once a table became transposed.
- **Inverse coverage** — `keeps desktop equal fixed while mobile auto resets it via media query` and `enables fixed layout on mobile even when desktop is auto`, plus `does not pin the action column in the viewport that is still auto` and `drops fixed layout when row data turns the table transposed`. The last two fail against the prior code.

Fixed in `b23b6f6f`.

src/elements/basic/TableElement/index.tsx
```diff
@@ -673,6 +673,40 @@ function TableElement({
...
+  // applied to transposed tables. The colgroup keeps the action/delete columns
+  // at fixed widths while data columns stay unsized, so only the data columns
+  // split. The colgroup renders whenever either viewport is equal, since DOM
+  // structure can't be media-queried; its widths are, so that the viewport
+  // still on auto keeps sizing those columns to their content.
+  //
+  // `isTransposed` tracks the row count, so it flips as data loads. The apply
+  // stays unconditional to overwrite the memoized target — guarding it would
+  // strand a stale `fixed` on a table that has since become transposed.
+  styles.apply('table', 'column_sizing', (columnSizing: any) => ({
```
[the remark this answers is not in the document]
freeman-zhang: Good catch — confirmed and taken the second option (hide the control, narrow the contract).

`isSpreadsheet` branches to `SpreadsheetTable` before the classic `<table>`, so neither `styles.getTarget('table')` nor the `<colgroup>` reaches the DOM. Verified by rendering with `display_mode: 'spreadsheet'` and `column_sizing: 'equal'`: no colgroup, no classic table, and `tableLayout: fixed` computed into a style target nothing consumes. The spreadsheet sizes columns through TanStack's own `table.state.columnSizing` — same name, unrelated to the style key.

We went with hiding rather than carrying it through, since equal-width would have to reconcile with the virtualizer's `estimateSize`, drag-resize overrides, and auto-fit measurement, including precedence once a user has manually resized a column:

- **Builder** — the Columns section is gated on `display_mode !== 'spreadsheet'` (feathery-frontend `646bd4329`), matching how `TablePropertiesPanel` already hides Search and row Actions there: *"the spreadsheet has no place for them yet, so the setting is hidden rather than offered and ignored."* Covered both ways, and the hide test fails without the gate.
- **Backend contract** — `column_sizing` help text now states it applies to the classic table only and that spreadsheet mode sizes its own columns (feathery-backend `14de20619`).

One deliberate leftover: the `styles.apply` call stays unconditional, so it still computes `tableLayout` in spreadsheet mode into that unused target. Guarding it would reintroduce the stale-target bug it was made unconditional to fix — `isTransposed` derives from the row count and flips after first render.

src/elements/basic/TableElement/index.tsx
```diff
@@ -673,6 +673,40 @@ function TableElement({
...
+    tableLayout: !isTransposed && columnSizing === 'equal' ? 'fixed' : 'auto'
+  }));
+  const desktopEqual = element.styles?.column_sizing === 'equal';
+  const mobileSizing = element.mobile_styles?.column_sizing;
+  // An absent mobile override inherits desktop, matching `apply`.
+  const mobileEqual =
+    mobileSizing === undefined ? desktopEqual : mobileSizing === 'equal';
```
ayeshah: const mobileEqual = (mobileSizing ?? desktopEqual) === 'equal';

can be rewritten as this

ayeshah (approved):

src/elements/basic/TableElement/index.tsx
```diff
@@ -673,6 +673,40 @@ function TableElement({
...
+  // Pin the action/delete columns only in the viewport that is actually fixed;
+  // `auto` elsewhere so an auto-layout viewport still sizes them to content.
+  const utilityColStyle = (width: string) => ({
+    width: desktopEqual ? width : 'auto',
+    ...(styles.handleMobile && mobileSizing !== undefined
+      ? {
+          [styles.mobileBreakpointKey]: {
+            width: mobileEqual ? width : 'auto'
+          }
+        }
```
ayeshah: ...(styles.handleMobile && mobileSizing !== undefined && {
  [styles.mobileBreakpointKey]: {
    width: mobileEqual ? width : 'auto'
  }
})

ayeshah: can be rewritten ^

PR #1804, opened 2026-08-25
fix: scope repeat-row validation errors to the rows they apply to

kennnyyyy:
Inline validation errors were keyed by servar key alone, so an error validated on one repeat row rendered on every row sharing that key -- including rows the user had already filled and rows just added -- and sibling rows overwrote the single shared entry, making messages flash and disappear. Adding or removing a row also re-ran validation once a submit had enabled auto-validate, flagging a brand-new untouched row in both inline and browser-native modes.

Store errors keyed only by the real field key, with per-row messages in a nested byIndex map. Keeping the row out of the key string means a literal field named `foo-0` can never collide with row 0 of a repeated field `foo`, and row removal shifts entries inside byIndex rather than pattern matching key strings. An empty message clears rather than merges: a field-wide write clears the field and all its rows, an indexed write clears only that row, and emptied entries are dropped so neither the renderer nor the aggregate validity check sees a stale entry.

Also:
- Matrix fields apply the `${key}-${questionIndex}` suffix only for HTML5 DOM targeting; inline storage keeps the real servar key, so matrix required errors actually render.
- Button and payment errors merge into the current error map and carry the clicked row, instead of writing field-wide into a fresh map that wiped unrelated errors; the button retry clear now publishes its update.
- Row removal reindexes every error-owning element in the container -- servar fields, buttons and containers -- not just servar fields.
- Assistant snapshots keep structured (key, repeatIndex) identity through the diff, and click results await the producer's pending publication so a button's asynchronously published error is not missed.

Repairs setFieldErrors({ index, message }), which previously ignored the index. Submit-time validation is unchanged, so empty required rows are still caught.

- [ ] Tested end to end

## Related pull requests

https://github.com/feathery-org/ai-services/pull/725

### DIFF
--- a/src/Form/grid/Element/index.tsx
+++ b/src/Form/grid/Element/index.tsx
@@ -723,6 +723,7 @@ const Element = ({ node: el, form }: any) => {
                 formRef,
                 fieldKey: el.servar.key,
                 message,
+                index: el.repeat,
                 errorType: formSettings.errorType,
                 servarType: el.servar.type,
                 inlineErrors: { ...inlineErrors },
--- a/src/Form/grid/Element/utils/utils.spec.ts
+++ b/src/Form/grid/Element/utils/utils.spec.ts
@@ -1,6 +1,7 @@
 import { getFieldValue } from '../../../../utils/fieldHelperFunctions';
 import {
   fileFieldShouldSubmit,
+  getInlineError,
   handleCheckboxGroupSelectAllChange
 } from './utils';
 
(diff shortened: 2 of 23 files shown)


trebitowski (approved):
Reviewed the full diff at head `8c07e5d6` (includes the earlier review-fix commit).

**Approach**: inline errors keyed by real field key with per-row messages in a nested `byIndex` map is the right structure — no key-string collisions (literal `foo-0` vs row 0 of `foo`), and row removal reindexes `byIndex` instead of pattern-matching keys. Clear semantics (field-wide empty write clears everything, indexed empty write clears one row, empty entries dropped) are well-tested.

**Verified**: matrix html5-vs-inline key split, `repeat: number | undefined` kept distinguishable from row 0 through `submitStep`, button/payment errors merging into the live map instead of wiping it, row-removal shifting all error owners (fields + buttons + containers) after any pending async publish, html5 `clearBrowserErrors` on removal, and assistant tools carrying `(key, repeatIndex)` as separate fields and awaiting `pendingInlineErrorPublish`.

**Validation**: tsc clean in src/, eslint 0 errors, all 15 touched suites pass (101 tests). Pre-existing `integrationClient.spec.js` failures reproduce on master and are unrelated.

No blocking issues. Two advisory nits, not worth holding the PR: `applyInlineError` continues the existing in-place mutation convention of `setFormElementError`, and the `publishLater` test helper is duplicated across three assistant specs.

PR #1806, opened 2026-08-25
fix: render the origin step when the session request fails

FranklinGao123:
## Changes

Fixes [HOSTED-FORMS-3WY](https://feathery-forms.sentry.io/issues/7685671246/) — `TypeError: Cannot convert undefined or null to object`. 1611 events, 485 users, 1570 of them on 2026-08-24.

### What a visitor experiences today

When the `panel/session/v3` request fails, the form renders **nothing**. Not an error, not the first step — a blank page.

The form definition still loads fine from the CDN, so the fields are in React state. They are never displayed, because the code path that selects a step throws before it gets there.

### Root cause

`Form/index.tsx` fires two requests and joins them:

```
newClient.fetchSession(formPromise, true)
  .then(([session, steps]) => { ...normal path...  })
  .catch(async (error) => {    ...fallback...      })
```

The `.catch` is the only recovery path. There is no retry and no error boundary behind it. Its job is to call `setStepKey`, which is the switch that makes the form appear.

It could never do that:

```js
const [data] = await formPromise;          // <-- the bug
const newKey = (getOrigin as any)(data).key;
```

`formPromise` resolves to steps **keyed by step key**, built at `index.tsx:1750`:

```js
steps = steps.reduce((result, step) => { result[step.key] = step; return result; }, {});
```

So `{ step_abc: {...}, step_def: {...} }`. Array destructuring on that is wrong.

In modern JS it would be *loudly* wrong — `const [d] = {}` throws `TypeError: steps is not iterable`, right at the mistake. But this package compiles to ES5 (`tsconfig.json` → `"target": "es5"`, `"downlevelIteration": true`), so TypeScript rewrites the destructuring through tslib's `__read`:

```js
function __read(o, n) {
  var m = typeof Symbol === "function" && o[Symbol.iterator];
  if (!m) return o;          // <-- silent pass-through
  ...
}
```

That guard exists to let array-likes through untouched. It also swallows this mistake. Running both versions side by side:

```
1. source JS  : TypeError: steps is not iterable
2. shipped ES5: undefined
3. then       : TypeError: Cannot convert undefined or null to object
4. fixed      : { origin: true }
```

Line 2 is what ships. `data` becomes `stepsObject[0]` → `undefined`, and `getOrigin(undefined)` calls `Object.values(undefined)` one frame later.

That is why the Sentry stack points at `utils/stepHelperFunctions.ts:72`, a file with no bug in it, rather than at the line that is actually wrong.

### A second defect on the next line

The same handler's next two lines had a related fault, from the same stale closure:

```js
if (trackHashes.current) setUrlStepHash(navigate, steps, newKey);
else setStepKey(newKey);
```

`steps` here is the outer `useState({})` from `index.tsx:421`. The mount effect that holds this handler is guarded by `if (clientRef.current) return;`, so it runs once and its closure keeps `steps` at its first-render value — `{}`. `setSteps` populates the state, but that builds a *new* closure on a later render; this handler keeps the old one.

The sibling `.then(([session, steps]) => ...)` is not affected, because its `steps` is a **destructured parameter** that shadows the state and carries the real object.

So on a form with hash navigation:

- `setUrlStepHash(navigate, {}, newKey)` hits its `Object.keys(steps).length > 1` guard and never navigates.
- `setStepKey` sits in the `else`, so it never runs either.
- Neither the hash nor the step key is set. Blank form.

Compare the success path 8 lines above, which is correct:

```js
if (trackHashes.current) setUrlStepHash(navigate, steps, newKey);
setStepKey(newKey);        // unconditional
```

### The fix

```diff
-        const [data] = await formPromise;
+        const data = await formPromise;
         const newKey = (getOrigin as any)(data).key;
-        if (trackHashes.current) setUrlStepHash(navigate, steps, newKey);
-        else setStepKey(newKey);
+        if (trackHashes.current) setUrlStepHash(navigate, data, newKey);
+        setStepKey(newKey);
```

The fallback now mirrors the success path. I audited every `steps` reference inside the mount effect (`index.tsx:1720-1916`); all the others are destructured parameters, so this handler was the only one reading the stale state.

The likely origin is a copy-paste: the sibling `.then(([session, steps]) => ...)` destructures correctly, because `fetchSession` genuinely returns an array. The same brackets were carried into the `.catch`, where the awaited value has a different shape.

The handler was added on 2024-04-07 in #1176. `formPromise` already returned the reduced object at that commit, so **this line has never worked**. It stayed invisible because it only runs when the session request fails, which was rare — until one org crossed its daily session cap and produced ~1500 failures in three hours.

### Why no test caught it

The `getOrigin` mock in `Form/tests/testMocks.tsx` ignored its argument:

```js
getOrigin: () => ({ key: 'origin' }),
```

A stub that never looks at what it is passed cannot notice being passed `undefined`. It now mirrors the real implementation, including throwing on a nullish argument.

Added `session fetch failure fallback` in `Form/tests/index.spec.tsx`, three cases, all failing on `master` and passing here:

| Test | Pins |
| --- | --- |
| `renders the origin step when the session request rejects` | the destructuring fix |
| `selects the origin step when hash navigation is on` | `setUrlStepHash` receives the real steps, not `{}` |
| `selects the origin step on a single-step form with hash navigation` | `setStepKey` runs even when no hash is written |

`setUrlStepHash` in the mock was `() => {}`; it is now a `jest.fn()` so the second case can assert its argument.

### What triggers the failure upstream

Two different causes reach this same handler. Both are conditions the form is meant to survive, and neither is fixed here:

| Cause | Share of sampled events | Where |
| --- | --- | --- |
| HTTP 400 `Too many sessions` | 94% | `feathery-backend apps/api/views.py:1388` — tier-0 org over 5000 anonymous sessions/24h |
| `TypeError: Failed to fetch (api.feathery.io)` | 6% | `_fetch`'s ignore list uses an exact string match, so the host-suffixed message slips past |

## Follow-ups not in this PR

- `checkResponseSuccess` has no `429` case, so a throttled request becomes `FetchError('Unknown error')` and takes this same path.
- `TYPE_MESSAGES_TO_IGNORE.includes(e.message)` should be a substring match.
- The quota rejection returns `code: "default-error"`, so the SDK cannot distinguish it from any other 400. It needs a stable code before a form-off screen can key off it.
- Nothing notifies an org when it crosses the daily session cap. It loses submissions silently for up to 24 hours.
- If `fetchForm` also fails, `await formPromise` throws inside the `.catch` and that rejection is itself unhandled. Nothing is renderable in that case, so the visible outcome is the same, but the Sentry noise is avoidable.
- The 2026-08-24 spike was ~480 events/hour flat for three hours, 95% one frozen `Chrome 124.0.0 / Linux` UA across 35 AWS us-east-1 IPs, each URL carrying a different recipient's address — a link scanner walking an email campaign's send list. Those loads consumed the org's session cap ahead of its real recipients.

- [ ] Tested end to end
- [x] Included screenshots or walkthrough video of change if impacts UX — n/a, no visual change; covered by the new test

## Verification

- `npx jest` — 188/188 suites, 2957 passed, 3 skipped
- All three new tests fail on `master` and pass on this branch (checked in both directions)
- `eslint` clean on the changed files
- `tsc -p tsconfig.typecheck.json` — 4 errors, byte-identical to `origin/master` (pre-existing `HubActionOptions` drift in local `@feathery/client-utils` types)

Left unchecked: I have not run this against a live form with a failing session request. Worth doing before merge.

I also could not measure how many affected forms use hash navigation. `track_hashes` is a form setting, not an event tag, and the crash prevents a hash from ever being written — so its absence from the breadcrumbs is not evidence either way. `iHtDkN` demonstrably uses it, because the visitor arrived on a link that already carried `#Who%20is%20signing`.

### DIFF
--- a/src/Form/index.tsx
+++ b/src/Form/index.tsx
@@ -1908,8 +1908,10 @@ function Form({
         // Go to first step if origin fails
         const data = await formPromise;
         const newKey = (getOrigin as any)(data).key;
-        if (trackHashes.current) setUrlStepHash(navigate, steps, newKey);
-        else setStepKey(newKey);
+        // Mirrors the success path above: `steps` state is still {} inside this
+        // effect's closure, and the hash branch has to set the key too.
+        if (trackHashes.current) setUrlStepHash(navigate, data, newKey);
+        setStepKey(newKey);
       });
   }, [activeStep, setSteps, updateFieldValues]);
 
--- a/src/Form/index.tsx
+++ b/src/Form/index.tsx
@@ -1906,7 +1906,7 @@ function Form({
       .catch(async (error: any) => {
         console.warn(error);
         // Go to first step if origin fails
-        const [data] = await formPromise;
+        const data = await formPromise;
         const newKey = (getOrigin as any)(data).key;
         if (trackHashes.current) setUrlStepHash(navigate, steps, newKey);
         else setStepKey(newKey);
--- a/src/Form/index.tsx
+++ b/src/Form/index.tsx
@@ -1905,8 +1905,18 @@ function Form({
       })
       .catch(async (error: any) => {
         console.warn(error);
-        // Go to first step if origin fails
         const data = await formPromise;
+        // Collaborator status and field allow lists only ride on the session
+        // response, and nothing else on the client carries them. Fail closed
+        // rather than render a form this collaborator may not be permitted to
+        // fill. Anonymous collaboration is not detectable here — the CDN form
+        // payload carries no collaboration flag.
+        if (initState.collaboratorId || initState.collaboratorReview) {
+          formOffReason.current = CLOSED;
+          setRender((render) => ({ ...render }));
+          return;
+        }
+        // Go to first step if origin fails
         const newKey = (getOrigin as any)(data).key;
         // Mirrors the success path above: `steps` state is still {} inside this
         // effect's closure, and the hash branch has to set the key too.
--- a/src/Form/tests/index.spec.tsx
+++ b/src/Form/tests/index.spec.tsx
@@ -17,6 +17,7 @@ import {
 import { JSForm } from '..';
 import FeatheryClient from '../../utils/featheryClient';
 import internalState from '../../utils/internalState';
+import { initState } from '../../utils/init';
 import {
   _clearDocxDirtyRegistry,
   hasDirtyDocxEditors,
@@ -206,6 +207,8 @@ describe('session fetch failure fallback', () => {
   afterEach(() => {
     delete ClientMod._spies.overrides.fetchForm;
     delete ClientMod._spies.overrides.fetchSession;
+    initState.collaboratorId = '';
+    initState.collaboratorReview = '';
   });
 
   // The session request fails for causes the form is meant to survive: a 400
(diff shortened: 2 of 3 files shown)


FranklinGao123:
Closing in favor of a two-way split of this work. No content is lost — with both branches applied, all three files are byte-identical to this PR's head, and the combined diffstat matches this PR's 205/36 exactly.

- **#1809 — the crash fix.** The `downlevelIteration` destructuring bug (`HOSTED-FORMS-3WY`), the stale-closure `steps` → `data` fix, and the `getOrigin` test-mock alignment. Ready for review now.
- **#1810 — collaborator fail-closed branch** (draft, stacked on #1809). A policy decision about what a user sees when a gate check cannot run, with the open items carried over.

Why split: the crash fix is proven and uncontroversial; the fail-closed branch is a judgment call that leaves three gates bypassed and turns a transient network blip into a permanent-sounding message. Bundling them meant the blank-page fix waited on that debate.

The split also fixes a description/diff mismatch — this PR's description documents only the 4-line destructuring change and omits the collaborator branch, roughly 40% of the diff.

The unresolved thread here asking for a typed fallback-safe result belongs to #1810, along with the unchecked end-to-end item. Please continue there.

PR #1812, opened 2026-08-27
fix: name every form control and element, mirror custom-widget values into hidden inputs

trebitowski:
## Problem

Several rendered elements are anonymous in the DOM: they carry no `name` or `id`, so anything that reads the form DOM (autofill, analytics tags, third-party form scanners) has nothing to attribute an interaction or a value to and records an unnamed element instead.

- `<textarea>`, `<select>`, the date picker input and the Signature / QR / Audio error inputs had an `id` but no `name`.
- Matrix radios shared a group `name` with no per-option `id`.
- Custom widgets (button group, rating, slider, color picker, multiselect) never expose their value under the field key at all, because they are built from divs, svgs or third-party components rather than native inputs.
- The rating icons, color swatch and file drop zone had no identifier of any kind.
- Containers, text blocks, progress bars, images, videos and tables had no `name`.

## Changes

Attribute-only additions. No tags change, no styling changes, no focus, tab-order or screen-reader changes.

- **`name` matching the existing `id`** on the textarea, select, date picker input and the Signature / QR / Audio error inputs.
- **Stable per-option `id`** on each Matrix radio, alongside the existing group `name`.
- **`id`** on each rating icon, the color swatch and the file drop zone (button-group options already had one).
- **`HiddenValueInput`** (new, `src/elements/components`): mirrors a field's value into a `type="hidden"` input named by the field key. Used by button group, rating, slider, color picker and multiselect. The slider handle is also named via `handleRender`.
- **`setFormElementError`** filters `type="hidden"` inputs out of its `form.elements.namedItem` lookup so a mirror can never be targeted for error display or shift the `elements[listIndex]` indexing used for repeated fields. The `BUTTON` branch now null-guards its `#error_` lookup instead of throwing.
- **`name` on non-control elements** via a small `nameProps` helper (`src/utils/domName.ts`): containers by key, text blocks by content (64-char cap, key fallback), and fixed names `progress_bar`, `image`, `video`, `table`. Browsers and assistive tech ignore `name` outside form controls, so this is inert at runtime.

## Special attention

- `setFormElementError` resolves targets via `namedItem`, which matches `id` **or** `name`. Every new `name` here either equals the element's existing `id`, sits on a `type="hidden"` input (filtered), or sits on a non-listed element (div / svg / img / video, which never enter `form.elements`), so RadioNodeList membership for existing field keys is unchanged. Covered by `src/utils/__test__/setFormElementError.spec.ts`, which models the browser's `namedItem` because jsdom's never returns a RadioNodeList.
- The Rating spec's mocked `ErrorInput` was `type="hidden"`; it now matches the real component (`type="text"`) so hidden-input selectors hit the mirror and not the mock.

## Test plan

- [x] `yarn typecheck`, `yarn lint` clean
- [x] New DOM-naming specs for every touched field and element; `HiddenValueInput` and `setFormElementError` specs
- [x] Live pass on a local stack against a form with every field type: every control and value resolves by `name` or `id`; required-field errors still display on button group, rating, slider, color picker and multiselect
- [ ] CI green

### DIFF
--- a/src/Form/grid/Container/index.spec.tsx
+++ b/src/Form/grid/Container/index.spec.tsx
@@ -55,3 +55,20 @@ describe('Container document editor wiring', () => {
     expect(editor).toHaveAttribute('data-container-id', 'container-1');
   });
 });
+
+describe('Container DOM naming', () => {
+  it('names every container by its key', () => {
+    const { container } = render(
+      <Container
+        node={{ ...docxNode, properties: {} }}
+        viewport='desktop'
+        form={{
+          formInstanceId: 'internal-form-id',
+          activeStep: { id: 'step-1' },
+          formSettings: { mobileBreakpoint: 480 }
+        }}
+      />
+    );
+    expect(container.querySelector('[name="container-1"]')).toBeTruthy();
+  });
+});
--- a/src/Form/grid/Container/index.tsx
+++ b/src/Form/grid/Container/index.tsx
@@ -1,5 +1,6 @@
 import React, { PropsWithChildren, useRef, useState } from 'react';
 import { StyledContainer, getCellStyle } from '../StyledContainer';
+import { nameProps } from '../../../utils/domName';
 import { ACTION_STORE_FIELD } from '../../../utils/elementActions';
 import HoverTooltip from '../../../elements/components/HoverTooltip';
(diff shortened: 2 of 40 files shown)


trebitowski:
Superseded by #1843, which carries the same change rebuilt on a clean branch.

PR #1814, opened 2026-08-28
feat: add Send Files to Box click action

anthonyly-feathery:
## Summary
- Wires the new `box_send_files` action into the click-action dispatcher (`runElementActions`).
- New `src/integrations/box.ts`: `sendBoxFiles()` calls the new `boxSendFiles()` client method and surfaces errors via `setElementError`.
- The dispatch branch force-submits every uploaded file field (`submitAllUploadedFiles`) before calling the endpoint, since the button's own "Validate & Submit Step" toggle may be off — `submitPromise` alone would otherwise silently send zero files. This was the highest-risk silent-failure mode in the feature.
- Deliberately does **not** break on success (unlike Schwab's action, which navigates away): a following "Go To Step" action on the same button must still run. Also does not error when there's nothing to send, since a button may sit on a step where files are optional.
- `testMocks.tsx` changes: exposes a mutable `formState.steps` so tests can inject a `file_upload` servar field, and adds a `submitFiles` spy to `MockClient` — no prior test exercised the file-submission path.

Backend companion: [feathery-backend#3783](https://github.com/feathery-org/feathery-backend/pull/3783). Frontend companion: [feathery-frontend#3928](https://github.com/feathery-org/feathery-frontend/pull/3928).

Note: this branch does not include an unrelated in-progress `rollup.config.mjs` change that was already present locally — that stays out of this PR.

## Test plan
- [x] `yarn eslint` — clean
- [x] `yarn typecheck` — clean
- [x] `yarn test src/Form` — 94/94 pass, including the new `boxSendFilesAction.spec.tsx` (5 tests covering force-submit-before-send, chain-stops-on-failure, chain-continues-on-success, silent no-op with nothing to send, and error surfacing without escaping the action loop)

https://claude.ai/code/session_017Ahf9fsDAtGhkBaK46dRdG

### DIFF
--- a/src/Form/index.tsx
+++ b/src/Form/index.tsx
@@ -2811,7 +2811,11 @@ function Form({
         }
         // Nothing uploaded yet: this button may sit on a step where files are
         // optional, so fall through to the next action rather than erroring.
-        if (fileCount && !(await sendBoxFiles(client, setElementError))) break;
+        if (
+          fileCount &&
+          !(await sendBoxFiles(client, setElementError, action.field_id))
+        )
+          break;
       } else if (type === ACTION_TRIGGER_PLAID) {
         await submitPromise;
         await openPlaidLink(
--- a/src/Form/index.tsx
+++ b/src/Form/index.tsx
@@ -2811,9 +2811,12 @@ function Form({
         }
         // Nothing uploaded yet: this button may sit on a step where files are
         // optional, so fall through to the next action rather than erroring.
+        const boxFieldIds = (action.field_ids ?? [])
+          .map((f: any) => f.field_id)
+          .filter(Boolean);
         if (
           fileCount &&
-          !(await sendBoxFiles(client, setElementError, action.field_id))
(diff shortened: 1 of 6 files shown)


src/Form/index.tsx
```diff
@@ -547,6 +549,32 @@ function Form({
...
   };
 
+  // Box reads this fuser's stored file rows, so every file field must be
+  // persisted first — the triggering button's "Validate & Submit Step" toggle
+  // may be off (submitPromise is a no-op then) and a normal step submit only
+  // covers the active step. Re-submitting an already-uploaded file is a safe,
+  // deduped no-op (see submitExtractionFiles above). Returns how many file
+  // fields had a value, so callers can skip the network call when there's
+  // nothing to send.
+  const submitAllUploadedFiles = async () => {
```
bo-dun-1: not sure why this is needed. just submit the files from the field IDs specified in action.field_ids

PR #1816, opened 2026-08-28
fix(overlay): render pages offscreen so checkboxes never blank; stop TextLayer null-append crash

ayeshah:
## Changes

Two defects in the document review overlay (`DocumentViewer`), both in `DocumentCanvas`. Reported by IFP (Generate Documents → Overlay, PDF template with checkbox fields).

### 1. Checked checkboxes render blank until you resize
On the first render, checked-checkbox checkmarks came out blank and only appeared after a window resize.

**Root cause (confirmed by reading the canvas backing store, not screenshots):** `PdfPage` painted directly to the visible canvas, and a `page.render` is **cancelled** whenever the effect re-runs — e.g. the `ResizeObserver` corrects `pageWidth` from the initial `MAX_PAGE_WIDTH` constant to the real container width right after mount. pdf.js draws form-field **appearances (the checkmarks) last**, so a render cancelled mid-paint left the visible canvas showing page content with the checks missing, until a later clean render happened to finish (a resize is just another full render). Verified this was **not** a font race (`page.render` awaits fonts — tested by delaying the font fetch 3s) and **not** a data problem (the PDF's `/V`,`/AS`,`/AP` are correct).

**Fix:** render each page to an **offscreen canvas** and blit to the visible one only once the render **fully completes**. An interrupted render is now a no-op instead of a partial, blank-checkbox frame.

### 2. Uncaught `Cannot read properties of null (reading 'append')`
With `includeMarkedContent` on, pdf.js's TextLayer restores its container via `endMarkedContent -> parentNode`; these generated PDFs have unbalanced marked-content operators, so the container walks past the root to `null` and the next `append` throws — from an async stream callback our try/catch can't reach. The read-only viewer only needs selectable/SR text, so `includeMarkedContent` is turned off.

## Verification

Filled the real IFP staging form end-to-end and generated the paperwork. On the **first** render, with **no resize**: Financial Professional, Charles Schwab, Accounts Receivable, Accumulated Savings, Short-term, Conservative, Capital Appreciation all render checked, and the console is clean (no `append` error, no "1 error" badge). Typecheck + all 37 DocumentViewer tests pass.

- [x] Tested end to end (real generated forms, first render, no resize)

## Related pull requests

- hosted-forms-next: `fix/cashco-csp-pdfjs-fonts` (#258) — allows the pdf.js standard-font fetch under Cashco's custom CSP (only matters for Cashco's custom-CSP domain; standard `*.feathery.io` subdomains need nothing).

### DIFF
--- a/src/elements/components/DocumentViewer/DocumentCanvas.tsx
+++ b/src/elements/components/DocumentViewer/DocumentCanvas.tsx
@@ -6,7 +6,7 @@ import { loadPdfjs, PDFJS_STANDARD_FONT_DATA_URL } from './pdfjsLoader';
 import { color, radius, shadow, fontSize } from './tokens';
 import { secondaryButtonCss } from './buttonStyles';
 import { AlertIcon } from './icons';
-import { featheryWindow } from '../../../utils/browser';
+import { featheryDoc, featheryWindow } from '../../../utils/browser';
 
 const PAGE_GAP = 24;
 // US Letter aspect for loading placeholders; actual pages size themselves.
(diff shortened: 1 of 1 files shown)


ayeshah:
Superseded by https://github.com/feathery-org/feathery-react/pull/1818

PR #1818, opened 2026-08-31
fix(overlay): render checkbox checkmarks reliably (async standard-font race)

ayeshah:
Supersedes #1816.

## Problem

In the read-only document review **overlay**, checked checkboxes render **blank** — the box shows but the checkmark is missing — until the user **resizes the window**. The **downloaded** PDF is always correct. Reported on IFP (Quik + Feathery documents, PDF, Overlay mode); only the **Feathery** document's checks were affected, not the Quik forms.

## Root cause

pdf.js draws a checkbox's checked-state glyph with a **base-14 substitute font** (`FoxitDingbats` for `ZapfDingbats`), which it fetches **asynchronously** via `standardFontDataUrl` and does **not** await inside `render()`. So the first render after a document opens paints the checkmark as **nothing** until that font lands in the pdf.js worker; any later render (e.g. the one a resize triggers) then draws it. Only documents whose checks use the ZapfDingbats standard font hit this — the Feathery-generated PDF — while Quik forms use an embedded/other check font and are fine.

### How it was confirmed

Measuring ink inside each checked checkbox's **exact rect** (coords from pdf.js annotation data): on first paint the checkmark ink was `0`, and stayed `0` even after `document.fonts.ready` — the wrong signal, that's the browser `FontFaceSet`, not pdf.js's worker font loader — yet a re-render seconds later measured full ink (~80–140). This is why earlier width-flip and offscreen-blit attempts couldn't help: **the render output itself lacked the glyph**.

## Changes (all in `DocumentViewer/DocumentCanvas.tsx`)

- **The fix:** after a page renders, detect any checked checkbox whose glyph did not paint (zero ink in its rect) and **re-render on a short backoff (120ms → 3s) until it does**, then blit the healed frame. No-op once the font is loaded — the common case — so the extra work is paid only on pages that raced.
- Pass `standardFontDataUrl` to `getDocument` so pdf.js can fetch the base-14 substitute fonts at all.
- Render each page **offscreen** and blit only a completed frame, so an interrupted render can't leave a partial frame.
- `streamTextContent` with `includeMarkedContent: false` — these generated forms have unbalanced marked-content operators that made pdf.js's TextLayer walk past the container root and throw `Cannot read properties of null (reading 'append')` from an async callback the surrounding try/catch couldn't reach.

## Testing

- 37/37 DocumentViewer tests pass.
- Verified end-to-end on the IFP staging form (hosted-forms, prod backend): the Feathery document's checkmarks now appear on first view with **no resize**.

### DIFF
--- a/src/elements/components/DocumentViewer/DocumentCanvas.tsx
+++ b/src/elements/components/DocumentViewer/DocumentCanvas.tsx
@@ -6,11 +6,38 @@ import { loadPdfjs, PDFJS_STANDARD_FONT_DATA_URL } from './pdfjsLoader';
 import { color, radius, shadow, fontSize } from './tokens';
 import { secondaryButtonCss } from './buttonStyles';
 import { AlertIcon } from './icons';
-import { featheryWindow } from '../../../utils/browser';
+import { featheryDoc, featheryWindow } from '../../../utils/browser';
 
 const PAGE_GAP = 24;
 // US Letter aspect for loading placeholders; actual pages size themselves.
 const SKELETON_ASPECT = '8.5 / 11';
+
+// A checked checkbox's glyph (ZapfDingbats, drawn with a standard font pdf.js
+// fetches async and doesn't await in render()) can paint blank; re-render on
+// this backoff until it does.
+const HEAL_BACKOFF_MS = [120, 300, 700, 1500, 3000];
+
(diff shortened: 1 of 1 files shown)


trebitowski:
Closing since this is fixed by the pdf fillable field changes

PR #1820, opened 2026-08-31
feat: drag to reorder repeat container rows

FranklinGao123:
## What

Repeat containers could add and remove rows but never move one, so a filler who entered rows in the wrong order had to clear and retype them.
This adds an opt-in reorder affordance.

Enabled per container by a new `properties.reorderable` flag (the dashboard toggle is in feathery-org/feathery-frontend#3937).
The reorder UI is off by default, so no existing form grows a drag handle.

**The row-cap fixes are not gated by that flag**, because a container that can be grown past its configured limit is a bug either way.
Three behaviours change on upgrade for any form with a capped add-row action, with or without `reorderable`: Add Row stops at the container's row count rather than each field's own length, the `set_value` trailing row is withheld at the cap, and the `set_value` auto-append answers to a limit at all.
See **The row cap** below.

- Drag handle in a gutter beside each row, revealed on hover
- Live drag: the dragged row follows the pointer, displaced rows animate to open the gap
- Arrow keys to move the focused handle, `Escape` to abandon a drag in progress.
  A press that is not a drag focuses the grip, so the keys are reachable by pointer.
- A `+` that follows the pointer to the nearer seam of the row it is over, so a row can be inserted above or below, subject to the container's row cap.
  A lone row keeps its seam and loses only the grip: one row is exactly where you want to add the second.

## Why it is more than an array swap

A repeat row has no entity behind it - it is the same index across several independent arrays plus two side channels (`filePathMap` and `servar.metadata.repeat_options`), and nothing checks that those stay in step.
Every mutation funnels through `updateRepeatValues` and moves all three together.
Getting it wrong surfaces only as a submission wearing another row's uploaded file.

https://github.com/user-attachments/assets/de14711c-0334-427d-a01e-e8802dee2e8c

https://github.com/user-attachments/assets/cf402d15-5ed3-400f-ac4f-fa4b6508dde0

## Decisions worth reviewing

- **Rows are addressed by absolute repeat index**, published per row as a data attribute.
  A `hide_if` can hide a row in the middle, so rendered position and repeat index diverge.
  The index a rendered slot holds therefore changes as logic opens and closes gaps, and a pointer gesture snapshots it once at grab time so the paint and the drop math cannot disagree.
- **Drop targeting compares the dragged row's centre to neighbours' near edges**, not the pointer to their midpoints.
  The grip sits at the row's top corner so the pointer trails the row it carries - on the test form the old rule needed 317px of drag where this needs 121px.
- **Live shifting is transforms only.**
  Nothing reorders in the DOM until the drop commits, which keeps the positional React keys for repeat rows safe.
- **New `overlay` slot on `StyledContainer`.**
  Anchored to `.inner-container` the chrome drifts inward as border and padding grow and ends up drawn over the outline.
  The gutter offset also measures the row's border width, since an absolute offset is taken from the padding box.
- **`currentColor` and opacity, no palette.**
  The handle inherits the form's theme; a fixed neutral looks wrong on anything but a light form.
- **Up/down chevrons in the gutter, not a Move menu.**
  SC 2.5.7 wants a single-pointer alternative to dragging, and the arrow keys do not satisfy it - that criterion is about pointer input.
  Chevrons either side of the grip run the same step the keys do, so a button, a key and a one-row drag cannot diverge.
- **The seam follows the pointer to the nearer edge of the row**, rather than living on the bottom edge.
  A fixed bottom seam made inserting above a row a puzzle about whose seam you were using, and left no way at all to insert ahead of the first row.
  Only the block axis is considered, which is the axis rows stack on for every track the SDK lays out itself.
- **Movement is worked in screen order**, not DOM order, so a `*-reverse` track from a customer's own CSS drags, steps and displaces the same way.
  Labelling is not: `ordinal` is counted in absolute order, so on a reversed track `aria-label="Row N of M"` and the seam's above/below wording are wrong.
  Numbering by screen position needs DOM measurement, and the hook that builds the label runs at render time with nothing to measure.
- **Announcements and the instructions node are keyed by form instance.**
  A page can carry several Feathery forms, and a single global channel would have every one of them read out a drag that happened in another form.

## The row cap

`max_repeats` lives on the add-row *action*, not on the container, so a container can be targeted by several actions with different limits.
`getRepeatMaxRows` resolves the container's ceiling as the **loosest** defined limit, treating a blank limit as unlimited.
The loosest, because a filler uses whichever button still adds: with actions capped at 5 and 2, five rows are reachable, and taking the tightest made the seam refuse at two while the other button carried on.

Each add-row action still enforces its own `max_repeats` - that is unchanged and per-action.
The resolver is the number the newer paths use so they cannot contradict the buttons:

- The `+` seam refuses past the ceiling, and `useRepeatRowReorder` withdraws the button rather than leaving one that silently does nothing.
- `addRepeatedRow` measures the cap against the container's row count instead of each field's own array.
  **Behaviour change worth a look:** at the cap, Add Row no longer tops up a field that is shorter than its siblings (a file field that trails empty rows).
- The `set_value` auto-append passed no limit at all and now passes one.
- `clampRepeatCountToCap` withholds the trailing row a `set_value` trigger renders past the data once the container is at its cap.
  That row was the last way around a limit: typing into it grows the array through `justInsert`, upstream of every cap check, and the newly filled row immediately earns another - so a capped container could be grown without bound, one row per fill.
  It never clamps below the rows the data already holds, so a cap lowered after submissions hides no existing answers.

## No backend change

Non-file repeated fields are deleted and recreated in submitted array order on every step submit, and `in_repeat_order` falls through to `display_order_time`/`id`, so a reordered array persists as sent.
Verified end to end against a local backend: reordered, submitted, and read the rows back in the new order.

## Two unrelated bugs fixed

Both were silent blank pages hit while building a test form, each fixed with tests:

- `buildGridTree` walked children by incrementing until a lookup missed, so a gap in child positions truncated the branch - and a gap at index `0` blanked the whole step with nothing logged.
- `getOrigin` returned `{ key: '' }` when no step was flagged as the origin, so the SDK rendered nothing in silence.
  It now falls back to the first step and warns.
  A form with no steps still yields `''` and does not warn.

## Accessibility

Reordering has a single-pointer alternative (the gutter chevrons), so SC 2.5.7 is met.
Every control is 24x24 of hit area, which is SC 2.5.8; the painted glyphs are unchanged, the extra is transparent padding.
Focus is placed with `preventScroll` and the page is only nudged when a handle is genuinely off screen, which keeps SC 2.4.11 honest without the page lurching.

## Testing

Full suite green: **204 suites, 3168 passed, 3 skipped**.
Around 155 tests cover the pure permutation helpers, the row-cap resolver, drop-target geometry (variable row heights, hidden rows, reversed tracks), the drag hook (pointer capture released on every exit path, tap vs drag, an index that changes under a live slot, inline styles handed back as found, a row unmounting mid-drag handing the whole track back), `Container` gating and announcements, and the `Form`-level move and insert paths.

Verified in a browser on the hosted form against a local backend, with the SDK built from this branch:

- Drag, arrow keys, the insert seam, hover reveal, and focus landing on the moved row.
- **The seam and the lone row.** The seam tracks the pointer to either edge with the matching label and insert index, and a single-row container shows a seam with no grip.
- **The phantom row under a cap.** With a `set_value` trigger and a cap of 2, filling both rows stops at two instead of growing on every keystroke.
- **The hidden-row case end to end.** Four rows with the middle one hidden by a `hide_if`, so the slot under the pointer holds a different absolute index than it did at mount. The row moves exactly one place; reading the stale index instead sent it to the top and lifted nothing.
- **The cap end to end.** With a limit of 3: seams present at two rows, no seams at three while every grip stays live, and further Add Row clicks leave the count at three.
- Announcements match the handle's own label (visible ordinal, not absolute index), and a drag leaves no inline `position` or `transform` behind.

---

## Round 2 - what changed after the first review pass

Franklin used the feature and reported eight issues. Each was reproduced with measurements against a
locally built SDK before anything was changed; the diagnosis page is in the thread.

**A root cause none of the eight named.** Repeat rows are not nested inside their own container in
the DOM - every container on a step renders its rows as siblings of every other container's, under
one shared parent. `useRowDrag` took the track to be `row.parentElement`, so with three sortable
containers one track saw fourteen rows with the indices `0,1,2,3` repeated three times. Displacement
landed on the wrong container, the dragged row's own transform was overwritten so it did not visibly
move at all, and drop targets resolved across the merged list. Rows now publish their track
alongside their index and every lookup filters on both.

| | Reported | Now |
|---|---|---|
| Lift escaped the track | 900px above, 1588px below, off screen | clamped to the track's own extremes |
| `+` far from the handle | 317px apart on a 600px row | 1px - same column in the gutter |
| Sort and insert were one switch | - | `reorderable` and `insertable`, the latter opt-out so existing containers are unchanged |
| Hidden rows | `display:none` rows measured zero and poisoned the sort | excluded from the track |
| Hit targets | 20x24, 20x14, 18x18 | all 24x24 |
| Chrome colour | badge hard-coded `#fff` | resolved from the nearest painted ancestor |
| Lift stacking | `z-index: 2`, under the form's fixed chrome at 10 | 1000 |
| Focus after a drop | jumped to the top of the page | `preventScroll`, and the destination is in the right container |

**Also fixed, not reported:** a `set_value` trigger's phantom trailing row inflated the handle
labels ("Row 1 of 4" with three rows) and left the last row's down button enabled with nowhere to
step; and a submit landing mid-drag strips every row marker step-wide, which left transforms
stranded because cleanup re-queried by attribute instead of using the nodes it captured.

**One deliberate behaviour change worth a look.** The `+` is now withheld when no add-row action
targets the container at all. It borrows its ceiling from those actions, so a container none of them
named was treated as uncapped and could be grown without limit - on a container the author had
deliberately given no way to grow.

**Corrected from the first pass:** I had claimed a field hidden inside a row made the displacement
gap the wrong size. It does not - rows of uneven height land exactly where they should, verified
against the real geometry on a 216/114/216 track.

---

## The full chain

This feature spans four PRs. Merge in this order:

| # | PR | What it does |
|---|---|---|
| 1 | [feathery-react#1820](https://github.com/feathery-org/feathery-react/pull/1820) | The renderer: drag handle, keyboard moves, insert seam, remove button |
| 2 | *SDK release* | Run **Release & Publish**, type `minor` → **2.92.0** (master is already at 2.91.0) |
| 3 | *consumer bump* | `hosted-forms-next` `@feathery/react` 2.67.0 → 2.92.0 — this is what makes the feature real for form fillers |
| 4 | [feathery-frontend#3937](https://github.com/feathery-org/feathery-frontend/pull/3937) | The builder toggle that writes `properties.reorderable` / `insertable` |
| 5 | [feathery-backend#3879](https://github.com/feathery-org/feathery-backend/pull/3879) | Declares both keys on `SubgridPropertiesSerializer` so Robin can see and set them |
| 6 | [ai-services#813](https://github.com/feathery-org/ai-services/pull/813) | Teaches Robin when to use them |

The renderer is inert until the builder ships (nothing writes the property), so it lands safely first. The reverse order would give authors a toggle that does nothing. ai-services must follow the backend: `WritableContract` is derived from that serializer, so until it merges the flags are not in the schema Robin reads.

### DIFF
--- a/src/Form/grid/Container/index.spec.tsx
+++ b/src/Form/grid/Container/index.spec.tsx
@@ -1,4 +1,4 @@
-import { render } from '@testing-library/react';
+import { fireEvent, render } from '@testing-library/react';
 import { Container } from '.';
 
 // Records the props the real container would register dirty state under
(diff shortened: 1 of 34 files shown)


trebitowski:
This is pretty smooth. Some functionality/style nits:

1. Can we prevent the drag preview from going past the first or last slot? I like how it's already locked from moving side-to-side

2. The plus buttons aren't positioned correctly for reordering a row.

3. Can we make the sorting and plus buttons two independent properties? There are use-cases where you would want to sort but not allow adding new containers.

4. We need to handle hidden fields.

5. The target areas for the handles and plus buttons could be a bit larger.

6. Any ideas for allowing the plus buttons and handles to be styled?

7. The drag preview should show above all of the other content on the screen, but sometimes it is displayed underneath text inputs or other elements.

8. If I have multiple sorting containers, or maybe if the container causes the page to scroll, I see weird focus scrolling issues after I drop a container. It seems to jump to the first sortable row/top of page.

PR #1823, opened 2026-09-01
feat: fillable PDF fields in the Generate Documents review editor

trebitowski:
## Summary

Re-enables fillable AcroForm fields in the Generate Documents review editor (`DocumentViewer`). The old Quik viewer supported this; the unified review editor shipped read-only. Editable fields matter more than validation/masking, so those stay out of scope — fields are simply editable.

- **Interactive widget layer**: pages render with `AnnotationMode.ENABLE_FORMS` (replacing the read-only print-intent render) and pdf.js's `AnnotationLayer` draws the PDF's own form fields — text, checkbox, radio, dropdown — as live HTML inputs seeded from each field's `/V`. pdf.js 5.x takes `linkService`/`annotationStorage` in the AnnotationLayer constructor (4.x took them in `render()`); verified against the pinned 5.4.296 CDN build. Widget CSS is ported from that version's `pdf_viewer.css`, including the load-bearing `--total-scale-factor`/`--scale-round-x/y` definitions its `setLayerDimensions` sizing expressions require — without them the whole layer computes to 0x0 (caught in real-browser testing; invisible to jsdom). A widget-layer failure degrades the page to read-only instead of blocking the render.
- **Persistence**: edits are tracked through pdf.js's annotation-storage modified flag. Before any toolbar action finalizes (Sign / Download / Save / Continue / Create Draft), each edited document is serialized with `saveDocument()` and PATCHed to its envelope via the existing `saveEnvelopeFile` client method — the docx-editor endpoint is type-generic — so every outcome acts on the edited file. A failed upload re-marks the document dirty so retrying the action re-saves it.
- **Close gating**: Escape and the toolbar Back button are inert while a save/finalize is in flight. Closing mid-action would hide the viewer while the side effect (save/sign/send) completed invisibly after the surrounding flow was told the review was cancelled, and unmounting would race `pdfProxy.destroy()` against an in-flight `saveDocument()`.

No backend changes needed.

## Test plan

- [x] Unit: 6 new specs in `fillableFields.spec.tsx` — save-before-finalize ordering, edited-docs-only uploads, no-op when clean, dirty re-tracking after a failed upload, close-while-busy gating, and graceful behavior without `onSaveEnvelopeFile`
- [x] Full jest suite, lint, typecheck green
- [x] Real browser (Chrome, local page mounting `DocumentViewer` against a pdf-lib fixture with real CDN pdf.js): text/checkbox/radio/dropdown widgets render and accept input; the saved bytes re-parsed by a fresh pdf.js contain every entered value; save lands before finalize; a clean second action skips the upload; a further edit saves again; Sign finalizes and closes; a `/V`-only prefilled value (no appearance stream — the Quik-generated shape) displays in its widget
- [ ] Manual full-stack: fill fields in a real Generate Documents review flow and confirm the finalized output (download + DocuSign sign) carries the edited values


trebitowski:
Companion backend PR: feathery-org/feathery-backend#3802 makes editor-drawn annotations real AcroForm widgets so they're editable here too (this PR needs no changes for that — it renders whatever widgets exist). Independent — either can ship first.

ayeshah (approved):

src/elements/components/DocumentViewer/index.tsx
```diff
@@ -262,11 +279,48 @@ export default function DocumentViewer({
...
+  // persist it to the envelope, so every finalize outcome — download, sign,
+  // save — acts on what the filler sees. saveDocument resets the storage's
+  // modified flag, which clears the doc from dirtyDocs via onResetModified.
+  const saveEditedDocuments = async () => {
+    if (!onSaveEnvelopeFile) return;
+    for (const doc of payload.documents) {
+      if (!doc.envelope_id || !dirtyDocs.current.has(doc.pdf_url)) continue;
+      const pdfProxy = loadedDocs.current[doc.pdf_url];
+      if (!pdfProxy) continue;
+      const bytes = await pdfProxy.saveDocument();
```
ayeshah: **A failed save wipes the "needs saving" flag, so retrying signs the old file.**

pdf.js clears its modified flag in a `.finally()` — even when `saveDocument()` fails. That removes the doc from `dirtyDocs`, and the `catch` here only covers upload failures, not `saveDocument()` itself.

So: user edits → clicks Sign → save fails → they click Sign again → the save is skipped and the **unedited** file gets signed. No warning.

Fix: wrap `saveDocument()` in the same try/catch and re-add the doc to `dirtyDocs` on failure. (The unit test misses this because its mock only resets the flag on success.)

ayeshah (commented):
Reviewed the save pipeline against the pinned pdf.js 5.4.296 build. Four issues below — all of them can silently lose or mix up a filler's edits, which is the core promise of this feature. Three share one theme: treat `saveDocument()` as fallible and freeze input while saving. The fourth is a one-line render param.

Also checked and ruled out: a stale-render/focus race in PdfPage, and a missing clientRef guard (the viewer only mounts after the client exists).

src/elements/components/DocumentViewer/index.tsx
```diff
@@ -262,11 +279,48 @@ export default function DocumentViewer({
...
+  // pdf.js's annotationStorage modified flag: set on the first widget change,
+  // cleared by saveDocument()/resetModified() — so a doc saved once and then
+  // edited again goes dirty again.
+  const dirtyDocs = useRef<Set<string>>(new Set());
+
   const onDocLoad = useCallback((pdfUrl: string, pdfProxy: any) => {
     loadedDocs.current[pdfUrl] = pdfProxy;
+    const storage = pdfProxy.annotationStorage;
+    if (storage) {
+      storage.onSetModified = () => dirtyDocs.current.add(pdfUrl);
```
ayeshah, line 303: **Anything typed while a save is running gets lost.**

Fields stay editable while a save is in flight (only the toolbar disables). pdf.js only fires the modified callback when the flag flips false→true — during a save it's already true, so the new edit never re-marks the doc dirty. When the save finishes, the flag resets and the doc looks clean.

Result: the edit is visible on screen but never saved, and a closing action (Sign/Continue) throws it away.

Fix: block field input while busy (e.g. `pointer-events: none` on the widget layer), same as the toolbar gate.

ayeshah (changes requested):

src/elements/components/DocumentViewer/index.tsx
trebitowski: Fixed in c841293f — `saveDocument()` now sits inside the same try/catch, so whichever step rejects the doc goes back into `dirtyDocs` and the retry re-saves it. The spec's mock now resets the flag in a `finally` like pdf.js does, plus a new test where `saveDocument()` itself rejects (fails on the old code, passes now).

src/elements/components/DocumentViewer/index.tsx
trebitowski: Fixed in c841293f — every page's `.annotationLayer` gets the `inert` attribute while `busyKey` is set, so clicks, typing and focus are all blocked (not just pointer events: a field already focused would otherwise keep taking keystrokes). Set imperatively since React 18 drops a boolean `inert`; `.annotationLayer[inert] section { pointer-events: none }` covers browsers without inert. New test asserts the layers are inert during an in-flight save and released after.

ayeshah (approved):

trebitowski:
Pushed `fba06b36`: the review editor now honours the action's `editor_read_only` — read-only PDFs render with pdf.js ENABLE mode (widget values painted into the canvas) and never mount the live form layer, so nothing dirties or saves. Dashboard checkbox for the overlay mode: feathery-org/feathery-frontend#3952.

PR #1827, opened 2026-09-01
feat(table): rebuild spreadsheet mode on TanStack Table v9, with buffered saves and validation

trebitowski:
## Changes

Rebuilds the Table element's `display_mode: 'spreadsheet'` on **TanStack Table v9**, then adds a save/validation model on top of it. Classic mode is untouched throughout.

### The grid
Ported from `tanstack/table/examples/react/spreadsheet` into `TableElement/spreadsheet/`: virtualized rows *and* columns, range selection, drag-fill, clipboard, undo/redo, a row context menu and a trailing add-row strip. Headers show Feathery column names, not A/B/C. Only the header row and the row-number gutter are sticky. Deliberately dropped from the example: ribbon, formula bar, sheet tabs, zoom, merged cells, freeze panes and edge autoscroll.

Spreadsheet mode **overrides** rather than falls back — pagination, search, sort and transpose are forced off on the data hook's copy of the properties, so switching back to Classic restores whatever the builder had configured.

### Editing is no longer autosave
Edits and row deletions buffer in `usePendingEdits` and apply on **Save**. Buffering lives in `index.tsx` rather than in either data source, so one implementation covers both Data Hub and field-backed tables and the batched write path underneath is unchanged. Row insertion is not buffered (it was never a server write) — Discard leaves a blank provisional row. Discard also drops the undo history, so Ctrl+Z cannot resurrect a discarded edit as a phantom pending change.

A status bar above the grid shows unsaved-change and issue counts, steppers that walk the failing cells in reading order, Discard (confirmed) and Save.

### Validation
`spreadsheet/validation.ts` mirrors the hub's cheap, stable field rules in TS — type, required, format, min/max, length, options, and uniqueness within the loaded rows — so counts update as you type. Cross-field `constraint_rules` and hub-wide uniqueness stay server-side and merge in after a save. Messages are worded like the backend's so a cell doesn't reword itself once the server has seen it. The client rules match the hub exactly where it matters: phone numbers are raw digits, file fields are never `required`, and staged rows neither claim nor are flagged for a unique value.

**Blocking is per row**: an error on a verified row is red and disables Save; the same rule on a staged (unverified) row is orange and saves anyway, because correcting extracted data is the point of editing one. Needs feathery-backend #3803. Staged rows can be deleted (the hub deletes by entry id) but not added to (`create` into the staged set is a batch replace).

### Cell editors follow the column
Option lists and booleans edit through a `<select>` (menu dropped on open via `showPicker`), numbers refuse non-numeric input, dates and datetimes get native pickers (a `datetime-local` picker shows the stored UTC instant in local time and writes back UTC), tax IDs are masked to their last four digits until edited, and hub `file` columns are read-only on every write path — editor, paste, fill and Delete — since they store `[{url, path}]` upload references that no editor can author.

**Cell input is parsed by the column's type**, not by its shape: `012345678` in a tax-ID or text column is nine digits, not the number 12345678, while number and boolean columns keep numbers and booleans as themselves.

Paste and fill are filtered through the column's rule, with the bar reporting how many values were skipped. Typing an invalid value is still allowed — that's what the error/warning states are for.

### Hub writes
A Data Hub `update` or `delete` that matches zero rows (the row was verified or removed elsewhere since it loaded) is treated as a failure: the cell or row reverts and the bar says the row changed in the Hub.

### Unsaved work now guards navigation
`utils/unsavedWork.ts` is a form-wide registry owning one `beforeunload` listener and one leave prompt. `docxDirtyRegistry` becomes a thin adapter over it, so the document editor and the table share a single dialog instead of queueing two. Next/Back and browser history are guarded through the same path.

- [x] Tested end to end — full stack locally against a real Data Hub (82 rows, 42 columns): the client's 327 warnings matched an independent recomputation from the database exactly; verified/staged blocking confirmed with DB checks; virtualization, editors, paste filtering and the guards all exercised in Chrome

Table, Form and unsaved-work suites: 250 tests passing; lint and typecheck clean.

**Note:** this branch also upgrades TypeScript 4.7 → 5.9.3. `@tanstack/react-table@9` ships `const` type parameters that TS 4.7 cannot *parse*, and `skipLibCheck` does not help — it skips type checking of `.d.ts`, not parsing. `target` stays `es5`, so emitted output is unchanged.

## Related pull requests
- feathery-backend #3803 — **must deploy first**
- feathery-frontend #3946 — needs this released first, since the designer canvas renders this component

Supersedes #1787 (the hand-rolled spreadsheet mode) — that one should be closed rather than merged.


src/elements/basic/TableElement/spreadsheet/validation.ts
```diff
@@ -0,0 +1,347 @@
...
+// Exactly the hub's `_PHONE_RE`: digits only, no punctuation and no leading
+// `+`. Anything looser here shows a cell as clean that the hub then rejects.
+const PHONE_PATTERN = /^\d{7,15}$/;
+
+function isEmpty(value: CellValue): boolean {
+  return value === null || value === undefined || value === '';
+}
+
+function decimalPlaces(value: number): number {
+  const text = String(value);
```
ayeshah: `String(value)` formats very small/large numbers in exponential form (e.g. `1e-7`), so this reads 0 decimals and passes a value the hub will reject. Could we format without the exponent before counting the decimals?

src/elements/basic/TableElement/spreadsheet/model.ts
```diff
@@ -0,0 +1,351 @@
...
+
+/**
+ * Escape one value for a TSV clipboard payload. Leading `=`, `+`, `-` and `@`
+ * are prefixed with an apostrophe so pasting into Excel or Sheets cannot turn
+ * form data into a live formula.
+ */
+export function escapeTsvValue(value: CellValue): string {
+  const text = formatCellValue(value);
+  const safeText =
+    typeof value === 'string' && /^[\t\r ]*[=+@-]/.test(value)
```
ayeshah, line 81: We prefix `-5`/`+1`/`=`/`@` with `'` on copy as a formula-injection guard, but `parseTsv` doesn't strip it on paste — so an internal copy→paste turns `-5` into `'-5`. Should paste undo the escaping for our own format?

src/elements/basic/TableElement/spreadsheet/fieldEditors.ts
```diff
@@ -0,0 +1,272 @@
...
+  return Number.isNaN(parsed.getTime()) ? null : parsed;
+}
+
+const pad2 = (value: number) => (value < 10 ? `0${value}` : String(value));
+
+/** `yyyy-MM-ddTHH:mm` in the browser's zone — what `datetime-local` shows. */
+function toLocalPickerValue(instant: Date): string {
+  return (
+    `${instant.getFullYear()}-${pad2(instant.getMonth() + 1)}-` +
+    `${pad2(instant.getDate())}T${pad2(instant.getHours())}:` +
```
ayeshah: `toLocalPickerValue` only emits `…THH:mm` (no seconds), so just opening and committing a cell that had seconds silently truncates them (12:00:30 → 12:00:00). Could we keep seconds in the format, or preserve the original when the field wasn't changed?

src/Form/grid/StyledContainer/styles.ts
```diff
@@ -250,7 +256,18 @@ export const getContainerStyles = (
...
 
         if (isFill(heightUnit)) {
+          // `fill` has to GROW into the available space, not merely cap at it:
+          // capping alone leaves a scrolling element (a spreadsheet table) at
+          // its natural height with empty space below. gig's parent_axis
+          // naming is inverted vs CSS — 'row' means the parent stacks its
+          // children vertically, so height is the main axis there.
           s.maxHeight = '100%';
+          if (parentAxis === 'row') {
+            s.flex = '1 1 auto';
```
ayeshah: This fill-height grow (`flex: 1 1 auto` + `minHeight: 0`) isn't gated to tables the way the width changes just above are, so now every fill-height element (images, text, containers) grows to fill space it didn't before. Was this meant only for the spreadsheet table? Might want the same `isTable` guard here.

src/elements/basic/TableElement/spreadsheet/SpreadsheetGrid.tsx
```diff
@@ -0,0 +1,1173 @@
...
+function measureTextWidth(value: string, bold = false): number {
+  if (!textMeasurementContext) {
+    textMeasurementContext = featheryDoc()
+      .createElement('canvas')
+      .getContext('2d');
+  }
+  // jsdom has no 2d context; fall back to a rough per-character estimate.
+  if (!textMeasurementContext) return value.length * 6;
+
+  textMeasurementContext.font = `${bold ? '600 ' : ''}11px Arial, sans-serif`;
```
ayeshah: Auto-fit measures at `11px Arial`, but cells render at 16px with the grid font stack, so double-click-to-fit comes out ~30% too narrow and truncates its own content. Should the measuring font match the actual cell font?

src/elements/basic/TableElement/spreadsheet/validation.ts
```diff
@@ -0,0 +1,347 @@
...
+        errors[cellErrorKey(rowIndex, fieldKey)] = message;
+        return;
+      }
+      if (!rule.unique || isEmpty(value) || staged) return;
+      // A duplicate flags the LATER row, so the first occurrence stays clean
+      // and the user fixes the copy rather than the original.
+      const values = seen.get(fieldKey) as Map<string, number>;
+      const normalized = String(value);
+      if (values.has(normalized)) {
+        errors[cellErrorKey(rowIndex, fieldKey)] = 'Must be unique';
```
ayeshah: This "flag the later row" rule assumes the duplicate comes *after* the original — but new rows are inserted at the top (`index.tsx`, `next.add(0)`), so pasting a duplicate onto a freshly-added row makes the **new** row the "first occurrence" and flags the **pre-existing** row with "Must be unique" instead. The error ends up on the original, not the copy the user just made. Could uniqueness flag the row that was actually added/edited rather than whichever has the higher index?

ayeshah:
**Row ordering on save (pre-existing, but newly visible via spreadsheet mode)**

`useHubTableSource.loadEntries` sorts every row by raw `id` on load, and that also runs on the post-save resync. So the order a user builds in the grid — new rows added at the top, manual inserts — is discarded and replaced by id-sort order after Save, and rows visibly jump around.

The id-sort predates this PR (it came in with #1795, `useHubTableSource.ts:217`), so this isn't a regression from these changes — but spreadsheet mode is what surfaces it for users, so flagging here for visibility. Might be worth preserving the grid's order (or sorting on something stable/user-facing) instead of raw `id`.

src/elements/basic/TableElement/spreadsheet/validation.ts
trebitowski: Fixed in 65a9172f: decimalPlaces splits off the exponent and folds it back in (1e-7 → 7 places, 1.5e21 → 0). Test added.

src/elements/basic/TableElement/spreadsheet/model.ts
trebitowski: Fixed in 65a9172f: parseTsv now strips a leading apostrophe only when it is guarding =, +, - or @ — the exact mirror of escapeTsvValue, and how Excel/Sheets read a leading apostrophe too. Unrelated apostrophes ('tis, O'Brien) are left alone. Round-trip test added.

src/elements/basic/TableElement/spreadsheet/fieldEditors.ts
trebitowski: Fixed in 65a9172f: toLocalPickerValue writes :ss (and .SSS) whenever the instant has them, so the picker shows a seconds field and preserves them on edit. One nuance: open-and-commit untouched did not actually truncate (the draft stays the stored string unless the input fires change), but changing the date did drop the seconds — that is what this fixes. Tests updated.

src/Form/grid/StyledContainer/styles.ts
trebitowski: Good catch — it was meant only for the table. Gated behind isTableElement(node) in 65a9172f (helper shared with the width branch), so every other fill-height element is back to the original cap-only behaviour.

src/elements/basic/TableElement/spreadsheet/SpreadsheetGrid.tsx
trebitowski: Fixed in 65a9172f: auto-fit now measures in the grid font — 400 16px for cells, 600 14px for the header label — via constants shared with styles.ts. Test added.

src/elements/basic/TableElement/spreadsheet/validation.ts
trebitowski: Fixed in 65a9172f: validateGrid collects every holder of a value and blames all but the original, where the original is the first holder the user did NOT change or add (new isCellChanged option, wired to the pending-edit buffer and new-row state). It falls back to lowest-index-wins only when every holder was changed (two pasted copies). Tests added.

trebitowski:
Re row ordering on save: addressed in 65a9172f on the SDK side. loadEntries no longer re-sorts by id on every load — rows already on the grid keep their position across resyncs (post-save and window-focus), and only entries the grid has not seen fall back to id order, placed at the top where the grid puts its own new rows. Sorting on something user-facing (created_at) would need the SDK hub get to return or order by it — it has no ordering today — happy to do that as a BE follow-up if preferred.

ayeshah (approved):

PR #1830, opened 2026-09-02
ci: open consumer bump PRs after an SDK release

FranklinGao123:
## Changes

Adds a matrix job to **Release & Publish** that reports as two independently named legs, **`bump-hosted-forms-next`** and **`bump-feathery-frontend`**, so a failure names its repo and branch protection can require either on its own. After the package is published and tagged, it opens a `chore: bump @feathery/react to <version>` PR in both consumers and requests review from @trebitowski.

| Consumer | Base branch | Range today |
|---|---|---|
| `hosted-forms-next` | `main` | `^2.85.0` |
| `feathery-frontend` | `master` | `^2.85.0` |

This replaces a manual step. The 2.85.0 bumps in both repos were opened by hand on 1 Sep (feathery-org/hosted-forms-next#262, feathery-org/feathery-frontend#3947), as were 2.82.0, 2.82.1 and 2.81.0 before them. The branch name and PR title here deliberately match that convention, so the automated PRs look like the ones they replace.

A `bump-consumers` boolean input (default `true`) lets a quick internal patch skip the two PRs.

### Design notes

- **The lockfile is the payload.** Both consumers already carry a caret range that a patch or minor release satisfies, so editing `package.json` alone would open an empty PR. The job runs `yarn upgrade @feathery/react@^<version>`, which moves `yarn.lock` and the range together.
- **Separate job, not extra steps.** `needs: release` means publish and the tag push are already done. A bump failure leaves a valid, published release behind rather than a half-finished one.
- **`fail-fast: false`.** A `feathery-frontend` registry problem must not cancel the `hosted-forms-next` PR.
- **Registry propagation.** npm does not serve a just-published version immediately. The job polls `npm view` for up to 5 minutes; without it the upgrade silently resolves the previous version.
- **Node version is read, not pinned.** `hosted-forms-next` has no `.nvmrc`; `feathery-frontend` pins 20.19.5. The job reads the consumer's file when it has one and falls back to the Node 20 that `hosted-forms-next` CI already uses.
- **A targeted range edit, not `yarn upgrade`.** Yarn 1's `upgrade` re-resolves the whole tree and buries the change. Rewriting the one range line and running `yarn install` keeps the diff tight: checked against the real manifests, a 2.67.0 to 2.85.0 jump moves 19 lockfile lines, touching only the SDK and its own `client-utils`.
- **The ping is belt and braces.** The PR body @-mentions the reviewer, and `gh pr edit --add-reviewer --add-assignee` runs afterwards with a tolerated failure. A reviewer request needs push access on the target repo; if that ever lapses, the job warns instead of failing and the @-mention still notifies.
- **Free breaking-change detector.** `hosted-forms-next` runs lint, build, unit tests, and full Playwright e2e on every PR. Each bump PR therefore exercises the new SDK against a real app before anyone merges.

### One prerequisite to confirm before the first run

**Branch protection on the consumers** must allow the bot to push a non-protected feature branch. Both target branches are only the PR base, never a push target.

An earlier revision of this PR claimed `FEATHERY_BOT_TOKEN` also needed `read:packages`. That was wrong. `feathery-frontend` has no `@feathery-org` dependencies at all - those are local `packages/*` workspaces - and its lockfile has zero `npm.pkg.github.com` resolutions. The auth step was dead code and has been removed.

- [ ] Tested end to end - the job only runs on `workflow_dispatch` of a real release, so it cannot be exercised from this PR. YAML parses and every `run` block passes `bash -n`. Worth watching the first release after merge.
- [x] Included screenshots or walkthrough video of change if impacts UX - not applicable, CI only

## Related pull requests

None.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -120,6 +120,9 @@ jobs:
   # and the tag push are already done by the time this runs, so a bump failure
   # leaves a valid release behind instead of a half-finished one.
   bump-consumers:
+    # Each matrix leg reports as bump-<repo>, so a failure names the repo it
+    # belongs to and branch protection can require the legs individually.
+    name: bump-${{ matrix.repo }}
     needs: release
     if: inputs.bump-consumers
     runs-on: ubuntu-latest
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -139,7 +139,7 @@ jobs:
             github_packages: true
     env:
       VERSION: ${{ needs.release.outputs.version }}
-      BUMP_BRANCH: bump/feathery-react-${{ needs.release.outputs.version }}
+      BUMP_BRANCH: chore/bump-feathery-react-${{ needs.release.outputs.version }}
       TARGET_REPO: feathery-org/${{ matrix.repo }}
       REVIEWER: trebitowski
     steps:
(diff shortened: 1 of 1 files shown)


FranklinGao123:
Closing PR due to duplicate of #1833

PR #1832, opened 2026-09-03
feat(docx): delete tables, rows & bound paragraphs, with orphan confirmation and atomic undo

ayeshah:
## What & why

Deleting a bound table, row, or paragraph in the docx editor was a **silent no-op** unless you happened to click a cell with no content control. Syncfusion refuses any destructive command whose selection touches a `lockContentControl` control (which every binding is) and refuses without any feedback. This PR makes deletion work from every natural gesture, safely, and never silently.

## What you can now do

- **Delete a bound table** — toolbar button, right-click → Delete Table, or select-the-table + Delete/Backspace.
- **Delete a row** — right-click → Delete Row, or select-the-row + Delete/Backspace, including rows whose cells hold locked formula values.
- **Delete a paragraph / prose range** that fully covers content control(s).
- When a deletion would **orphan a formula elsewhere** (something that reads a value you're deleting), a confirmation dialog appears; on confirm, those dependent values are converted to plain text (the same cached-number semantics the export strip uses).
- **Undo/redo is atomic** — one gesture restores the table/row/selection *and* re-wraps the dependent formulas as live content controls.
- **Locked-edit hint** — trying to edit a genuinely locked cell (typing into a computed total, a partial-overlap delete) now shows a brief "This content is locked and can't be edited here." It stays while the cursor is on the locked cell and clears when you move somewhere editable. Editable fields never trigger it.

Partial-overlap selections (deleting half a binding) stay blocked on purpose, with the hint explaining why.

## How it works

- `bindings/core/tableDeleteImpact.ts` — pure dry-run that runs the reconcile engine with and without the deleted content and diffs which formulas newly fail (a repeated value with a survivor strands nothing). No reference resolution is re-implemented; the engine is the authority.
- `bindings/tableDeleteGuard.ts` — wraps `deleteTable`/`deleteRow`, routes whole-table/row/prose Delete-Backspace, lifts the lock gate for these explicit whole-control gestures, confirms, unwraps orphans, and keeps undo atomic (native grouping for table/prose; gesture-chaining for rows, whose table-clone history entries can't be natively grouped on this Syncfusion version).
- `bindings/attachBindings.ts` + `index.tsx` — the locked-edit hint (gated on `canEditContentControl`, selection-driven visibility).

## Notable gotchas fixed along the way (each with a regression test)

- Unwrap via typing the value, not `removeContentControl` (whose undo re-splices markers at stale line indexes → duplicated values after a reflow).
- Restore document order in `contentControlCollection` after undo (out-of-order entries are invisible to Syncfusion lookups → a restored control behaves as plain text while still present in `serialize()`).
- Refocus the editor after the confirm dialog (Ctrl+Z afterward was going nowhere).

## Testing

- 410 tests green (unit tests for the impact analyzer + real-`DocumentEditor` integration tests for every gesture, undo/redo discoverability, partial-overlap, track-changes passthrough, and the hint).
- ⚠️ This class of behavior (a control present in `serialize()` but invisible to lookups when the collection is out of order) **cannot be caught by serialize-based or jsdom tests** — jsdom never reflows. A manual browser QA pass asserting control *discoverability* is recommended before merge; the flows were verified live in a kiwi env during development.

### DIFF
--- a/src/elements/components/DocxEditor/bindings/attachBindings.ts
+++ b/src/elements/components/DocxEditor/bindings/attachBindings.ts
@@ -65,6 +65,12 @@ export interface BindingsOptions {
    * the refusal is not silent.
    */
   onLockedEdit?: () => void;
+  /**
+   * Fired when the caret leaves a refused/locked spot for an editable one, so
+   * the host can hide the locked-edit hint immediately instead of waiting out
+   * its timeout.
+   */
+  onLockedEditResolved?: () => void;
   persistence?: DocumentPersistence | null;
   setTimeoutFn?: (fn: () => void, ms: number) => TimerId;
   clearTimeoutFn?: (id: TimerId) => void;
@@ -116,6 +122,7 @@ export function attachBindings(
     onSuppressContentChange,
     confirmTableDelete,
     onLockedEdit,
+    onLockedEditResolved,
     persistence = null,
     setTimeoutFn,
     clearTimeoutFn
@@ -216,8 +223,6 @@ export function attachBindings(
     }
   };
   const onContentChange = () => runGuarded(() => triggers.onContentChange());
-  const onSelectionChange = () =>
-    runGuarded(() => triggers.onSelectionChange());
   const onKeyDown = (args: any) =>
     runGuarded(() => triggers.onKeyDown(args?.event?.key));
   const onBlur = () => runGuarded(() => triggers.onEditorBlur());
@@ -230,6 +235,7 @@ export function attachBindings(
   // key is one hint, and stay quiet while the delete guard is mid-operation -
   // those refusals are ours to resolve, not to complain about.
   let lockedHintTimer: TimerId | null = null;
+  let lockHintActive = false;
   const scheduleTimeout = setTimeoutFn ?? ((fn, ms) => setTimeout(fn, ms));
   const cancelTimeout = clearTimeoutFn ?? ((id) => clearTimeout(id as never));
   const editRefused = (): boolean => {
@@ -241,11 +247,23 @@ export function attachBindings(
   const onLockedControl = () =>
     runGuarded(() => {
       if (!onLockedEdit || isDeleteGuardBusy() || !editRefused()) return;
+      lockHintActive = true;
       if (lockedHintTimer !== null) return;
       onLockedEdit();
       lockedHintTimer = scheduleTimeout(() => {
         lockedHintTimer = null;
       }, 600);
+    });
+
+  // Dismiss the hint the instant the caret lands somewhere editable; keep it up
+  // while a locked cell stays selected (until the host's own short timeout).
+  const onSelectionChange = () =>
+    runGuarded(() => {
+      triggers.onSelectionChange();
+      if (lockHintActive && !editRefused()) {
+        lockHintActive = false;
+        onLockedEditResolved?.();
+      }
     });
 
   eventful.addEventListener?.('contentChange', onContentChange);
--- a/src/elements/components/DocxEditor/bindings/attachBindings.ts
+++ b/src/elements/components/DocxEditor/bindings/attachBindings.ts
@@ -159,6 +159,21 @@ export function attachBindings(
     ...(clearTimeoutFn ? { clearTimeoutFn } : {})
   });
   const uninstallGuard = installKeystrokeGuard(editor);
+  // Explicit whole-table/row deletion: lift the content-control locks that
+  // made deleteTable/deleteRow silent no-ops, and confirm/unwrap when the
+  // deleted values feed formulas elsewhere (tableDeleteGuard has the full
+  // story). Installed BEFORE watchRowCommands so the row watcher's wrapper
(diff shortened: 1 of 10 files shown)


src/elements/components/DocxEditor/bindings/attachBindings.ts
```diff
@@ -181,15 +223,61 @@ export function attachBindings(
...
+  const cancelTimeout = clearTimeoutFn ?? ((id) => clearTimeout(id as never));
+  const editRefused = (): boolean => {
+    const module = editor.editorModule as
+      | { canEditContentControl?: boolean }
+      | undefined;
+    return module ? module.canEditContentControl === false : false;
+  };
+  const onLockedControl = () =>
+    runGuarded(() => {
+      if (!onLockedEdit || isDeleteGuardBusy() || !editRefused()) return;
```
lanthony42: `canEditContentControl`'s getter fires `contentControl` via `checkContentControlLocked`, so this loops infinitely

lanthony42 (changes requested):

src/elements/components/DocxEditor/bindings/attachBindings.ts
ayeshah: Good catch — fixed in c584e37d. The handler now reads `selection.currentContentControl` directly instead of `canEditContentControl`, so it never re-enters `checkContentControlLocked`. It suppresses the hint when the caret is in an editable inner field and shows it otherwise. Added a regression test that fires `contentControl` from a `canEditContentControl` getter to prove there's no recursion.

src/elements/components/DocxEditor/bindings/attachBindings.ts
```diff
@@ -151,6 +179,21 @@ export function attachBindings(
...
+  // story). Installed BEFORE watchRowCommands so the row watcher's wrapper
+  // stays outermost and its reconcile runs after the guard's grouped history
+  // entry closes.
+  const uninstallTableDelete = installTableDeleteGuard(editor, {
+    ...(confirmTableDelete ? { confirm: confirmTableDelete } : {}),
+    onDeleted: () => {
+      if (controller.phase !== 'idle') return;
+      const history = editor.editorHistoryModule;
+      if (history?.isUndoing || history?.isRedoing) return;
+      controller.flush({ mode: 'self-heal' });
```
lanthony42, line 194: A plain row delete flushes here and again in `rowCommandWatch`'s `onRowChange` the same turn, can it be one go?

src/elements/components/DocxEditor/bindings/attachBindings.ts
```diff
@@ -181,15 +224,86 @@ export function attachBindings(
...
-  const onBlur = () => runGuarded(() => triggers.onEditorBlur());
+    runGuarded(() => {
+      const key = args?.event?.key;
+      triggers.onKeyDown(key);
+      // Backspace/Delete inside a locked control is refused by Syncfusion
+      // WITHOUT firing 'contentControl' - canEditContentControl short-circuits
+      // on lockContents before it reaches checkContentControlLocked (the only
+      // thing that fires the event). So surface the hint from here. A
+      // multi-cell/structural selection has no currentContentControl, so this
+      // never fires on a real row/table delete.
```
lanthony42: A row selection reports the `[[table=...]]` wrapper here, not null, it stays quiet only because the wrapper isn't `lockContents`

lanthony42 (approved):

src/elements/components/DocxEditor/bindings/attachBindings.ts
ayeshah: There's a deliberate case for keeping both — they cover different entry points. `rowCommandWatch.onRowChange` is the *only* reconcile for **native** `insertRow`/`deleteRow` (context-menu delete, Tab-to-insert): those bypass `runCommands`, and it also drives insert-row adoption, so it can't be dropped. The guard's `onDeleted` ensures a reconcile runs after the guard's own delete + unwraps, once the grouped history entry closes. A plain row delete is the one path where both fire — but the second is a `self-heal`, and per the controller's invariant ("a reconcile that changes nothing does not touch the editor at all") it's a no-op on the already-consistent document: no writes, no history entry, no dirty flag. Consolidating would couple the two otherwise-independent mechanisms for no real gain, so we're keeping them separate.

src/elements/components/DocxEditor/bindings/attachBindings.ts
ayeshah: Good catch — fixed in 8d376aed. You're right the null-`currentContentControl` assumption was fragile. The hint no longer leans on it: it uses the same discriminator the delete guard uses — whether the selection spans multiple cells (`startCell !== endCell`). In-cell edit → hint; a genuine row/table span → silent. (An interim commit 67420704 tried excluding `isRowSelected`/`isTableSelected`, but a single locked cell reports `isRowSelected` true, which then suppressed the legitimate in-cell hint — `spansMultipleCells` is the correct discriminator.)

PR #1833, opened 2026-09-03
ci: open @feathery/react bump PRs in consumer repos after release

trebitowski:
## What

Adds a `bump-consumers` job to **Release & Publish** that runs after `npm publish` and opens a `chore: bump @feathery/react to X` PR in each consumer repo:

- `feathery-org/feathery-frontend` (base `master`, Node 20.19.5)
- `feathery-org/hosted-forms-next` (base `main`, Node 20)

The `release` job now exposes the new `version` and GitHub release URL as outputs. The bump job checks out the consumer with `FEATHERY_BOT_TOKEN`, runs `yarn add --ignore-scripts @feathery/react@^X` (keeps the caret pin, updates `yarn.lock`), pushes `chore/bump-feathery-react-X`, and creates the PR via `gh`. Logic lives in `.github/scripts/open-consumer-bump-pr.sh`.

## Behaviour

- Waits up to 2 min for the version to appear on npm before installing.
- No-op if the consumer is already on the version.
- Re-runs force-push the same branch and skip PR creation if one is already open.
- `fail-fast: false`: one consumer failing does not block the other. The release itself is already published and pushed by then.

## Requirements / assumptions

- `FEATHERY_BOT_TOKEN` needs `contents` + `pull_requests` write on both consumer repos (it is already used with the same name for PR creation in feathery-frontend's `claude-ticket.yml`).

## Test plan

- [x] Script dry-run against a local clone of hosted-forms-next with stubbed `gh` / `git push`: produced the expected package.json + yarn.lock commit and `gh pr create` call; no-op path confirmed.
- [x] `shellcheck`, `bash -n`, prettier, YAML parse clean.
- [ ] First real run: dispatch a patch release and confirm both PRs open (hosted-documents is intentionally not included; it is pinned exactly at 2.29.0).

### DIFF
--- a/.github/scripts/open-consumer-bump-pr.sh
+++ b/.github/scripts/open-consumer-bump-pr.sh
@@ -1,0 +1,68 @@
+#!/usr/bin/env bash
+# Opens (or refreshes) a PR in a consumer repo that bumps @feathery/react to
+# the version just published. Runs from the root of a checked-out consumer.
+#
+# Required env:
+#   VERSION       - the newly published @feathery/react version (e.g. 2.86.0)
+#   REPO          - consumer repo slug (e.g. feathery-org/feathery-frontend)
+#   BASE_BRANCH   - consumer default branch (master / main)
+#   RELEASE_URL   - link to the @feathery/react GitHub release
+#   GH_TOKEN      - token with contents + pull_requests write on $REPO
+set -euo pipefail
+
+: "${VERSION:?}" "${REPO:?}" "${BASE_BRANCH:?}" "${RELEASE_URL:?}" "${GH_TOKEN:?}"
+
+BRANCH="chore/bump-feathery-react-${VERSION}"
+TITLE="chore: bump @feathery/react to ${VERSION}"
+
+current=$(node -p "require('./package.json').dependencies['@feathery/react']")
+echo "Current pin in ${REPO}: ${current}"
+
+# The publish finished moments ago; give the npm registry time to serve it.
+for attempt in $(seq 1 12); do
+  if npm view "@feathery/react@${VERSION}" version >/dev/null 2>&1; then break; fi
+  if [ "$attempt" -eq 12 ]; then
+    echo "::error::@feathery/react@${VERSION} not visible on npm after 2 minutes"
+    exit 1
+  fi
+  echo "Waiting for @feathery/react@${VERSION} on npm (attempt ${attempt})..."
+  sleep 10
+done
+
+# Keep the caret range the consumers already use; yarn rewrites package.json
+# and yarn.lock together. --ignore-scripts skips husky/postinstall hooks the
+# runner doesn't need for a lockfile update.
+yarn add --ignore-scripts "@feathery/react@^${VERSION}"
+
+if git diff --quiet -- package.json yarn.lock; then
+  echo "${REPO} already depends on @feathery/react ${VERSION}; nothing to do."
+  exit 0
+fi
+
+git checkout -B "$BRANCH"
+git add package.json yarn.lock
+git commit -m "$TITLE"
+# Force-push so a re-run of the release workflow refreshes the same branch.
+git push --force -u origin "$BRANCH"
+
+existing=$(gh pr list --repo "$REPO" --head "$BRANCH" --state open --json number --jq '.[0].number // empty')
+if [ -n "$existing" ]; then
+  echo "PR #${existing} already open for ${BRANCH}; branch updated."
+  exit 0
+fi
+
+body=$(cat <<BODY
+Automated bump of \`@feathery/react\` from \`${current}\` to \`^${VERSION}\`.
+
+Release notes: ${RELEASE_URL}
+
+Opened by the feathery-react **Release & Publish** workflow.
+BODY
+)
+
+gh pr create \
+  --repo "$REPO" \
+  --base "$BASE_BRANCH" \
+  --head "$BRANCH" \
+  --title "$TITLE" \
+  --body "$body"
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -20,6 +20,9 @@ permissions:
 jobs:
   release:
     runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.version.outputs.version }}
+      release_url: ${{ steps.gh_release.outputs.url }}
     steps:
       - name: Validate User
         env:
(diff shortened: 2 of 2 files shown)


.github/scripts/open-consumer-bump-pr.sh
```diff
@@ -0,0 +1,68 @@
...
+    exit 1
+  fi
+  echo "Waiting for @feathery/react@${VERSION} on npm (attempt ${attempt})..."
+  sleep 10
+done
+
+# Keep the caret range the consumers already use; yarn rewrites package.json
+# and yarn.lock together. --ignore-scripts skips husky/postinstall hooks the
+# runner doesn't need for a lockfile update.
+yarn add --ignore-scripts "@feathery/react@^${VERSION}"
```
FranklinGao123, line 35: **Blocker: this line fails on every release for feathery-frontend.**

That repo declares `"workspaces": ["packages/*"]` in its root `package.json`, and it holds `"@feathery/react": "^2.85.0"` at the root as well. Yarn 1.22.22 refuses `yarn add` at a workspace root without `-W`.

I reproduced this on yarn 1.22.22 in a scratch workspace:

```
error Running this command will add the dependency to the workspace root rather than
the workspace itself, which might not be what you want - if you really meant it, make
it explicit by running this command again with the -W flag
```

The exit code is 1 without the flag and 0 with it. `set -euo pipefail` then aborts the feathery-frontend leg. The `hosted-forms-next` leg still passes, because that repo has no workspaces.

The root is the correct target here, so the flag is the whole fix. Please detect the case instead of hardcoding the flag, so a future workspace consumer also works:

```bash
add_flags=(--ignore-scripts)
node -p "!!require('./package.json').workspaces" | grep -q true && add_flags+=(-W)
yarn add "${add_flags[@]}" "@feathery/react@^${VERSION}"
```

.github/scripts/open-consumer-bump-pr.sh
```diff
@@ -0,0 +1,68 @@
...
+
+BRANCH="chore/bump-feathery-react-${VERSION}"
+TITLE="chore: bump @feathery/react to ${VERSION}"
+
+current=$(node -p "require('./package.json').dependencies['@feathery/react']")
+echo "Current pin in ${REPO}: ${current}"
+
+# The publish finished moments ago; give the npm registry time to serve it.
+for attempt in $(seq 1 12); do
+  if npm view "@feathery/react@${VERSION}" version >/dev/null 2>&1; then break; fi
```
FranklinGao123, line 23: The probe and the install read different registry hosts.

`npm view` queries `https://registry.npmjs.org`. Inside both consumer repos, `yarn config get registry` returns `https://registry.yarnpkg.com`, and the current `@feathery/react` entry in each `yarn.lock` resolves through that host.

The two hosts front the same registry, but they use different CDN edges. The probe can therefore pass while `yarn add` still gets a 404 for a few seconds. Please probe the registry that yarn will use:

```bash
if npm view --registry "$(yarn config get registry)" "@feathery/react@${VERSION}" version >/dev/null 2>&1; then break; fi
```

This failure is loud, so today it costs a re-run. It does not publish a wrong version.

.github/scripts/open-consumer-bump-pr.sh
```diff
@@ -0,0 +1,68 @@
...
+body=$(cat <<BODY
+Automated bump of \`@feathery/react\` from \`${current}\` to \`^${VERSION}\`.
+
+Release notes: ${RELEASE_URL}
+
+Opened by the feathery-react **Release & Publish** workflow.
+BODY
+)
+
+gh pr create \
```
FranklinGao123, line 63: No reviewer and no assignee. The PR opens and nobody hears about it. That is the manual step this job removes.

Please assign the person who dispatched the release, and add a reviewer for each consumer:

```bash
gh pr create \
  --repo "$REPO" \
  --base "$BASE_BRANCH" \
  --head "$BRANCH" \
  --title "$TITLE" \
  --body "$body" \
  --assignee "$GITHUB_ACTOR" \
  ${REVIEWERS:+--reviewer "$REVIEWERS"}
```

`GITHUB_ACTOR` is already in the runner environment. Add `REVIEWERS` to the matrix, and pass it through the step `env` block.

.github/workflows/release.yml
```diff
@@ -97,9 +104,64 @@ jobs:
...
+  bump-consumers:
+    needs: release
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - repo: feathery-org/feathery-frontend
+            base: master
+            node: '20.19.5'
```
FranklinGao123, line 127: `'20.19.5'` copies feathery-frontend's `.nvmrc` into this file. The value is correct today. It becomes stale without warning after that repo changes its Node version.

This job only resolves a lockfile, so the patch version does not matter. Please use `'20'` for both entries and delete the duplication. `hosted-forms-next` has no `.nvmrc`, and its own CI uses Node 20, so a shared `'20'` is correct for both.

.github/scripts/open-consumer-bump-pr.sh
```diff
@@ -0,0 +1,68 @@
...
+#   RELEASE_URL   - link to the @feathery/react GitHub release
+#   GH_TOKEN      - token with contents + pull_requests write on $REPO
+set -euo pipefail
+
+: "${VERSION:?}" "${REPO:?}" "${BASE_BRANCH:?}" "${RELEASE_URL:?}" "${GH_TOKEN:?}"
+
+BRANCH="chore/bump-feathery-react-${VERSION}"
+TITLE="chore: bump @feathery/react to ${VERSION}"
+
+current=$(node -p "require('./package.json').dependencies['@feathery/react']")
```
FranklinGao123, line 18: Minor and latent. This line reads `dependencies` only. Both consumers keep the SDK there today, so the current behavior is correct.

If a consumer moves the SDK to `devDependencies`, two things break. `current` becomes `undefined`, and the PR body reads "from `undefined`". `yarn add` also moves the dependency into `dependencies` without a warning.

A tolerant read costs one line:

```bash
current=$(node -p "const p=require('./package.json'); (p.dependencies||{})['@feathery/react'] || (p.devDependencies||{})['@feathery/react'] || '(absent)'")
```

FranklinGao123 (commented):
Thanks for this. The shape is right: a separate job after publish, `fail-fast: false`, one branch per version, and the logic in a script instead of inline YAML. I want to ship this. One item blocks it.

**Blocker**
- `yarn add` fails at feathery-frontend's workspace root. It needs `-W`. That leg aborts on every release.

**Please fix before merge**
- The readiness probe polls a different registry host than the install uses.
- The PR opens with no reviewer and no assignee.

**Minor**
- The hardcoded Node patch version duplicates feathery-frontend's `.nvmrc`.
- The `current` read looks at `dependencies` only.

I checked three things that are correct. `actions/checkout@v6` and `actions/setup-node@v6` match every other workflow in this repo. The `hosted-forms-next` matrix entry is right, because that repo has no workspaces and uses Node 20 in its own CI. The step named "Draft GitHub Release" does not set `draft: true`, so `release_url` points at a public release, and the link in the PR body works.

One question. `RELEASE_URL` is decorative, but line 12 makes it required with `:?`. A missing release URL then fails this job after the publish succeeds. Can you default it to empty instead?

PR #1834, opened 2026-09-03
fix: keep a percentage-height field inside its own cell

FranklinGao123:
## Changes

Three defects in the geometry a field draws itself with, plus a follow-up commit
correcting the flex basis the first fix originally shipped with -- see the boxed
note in #1 below. All were found by
measuring real layouts in a browser; jsdom has no layout engine, so the existing
suite cannot see any of them.

### 1. A percentage height overflowed the cell by the label's height

`applyHeight` gives the field's box `height: 100%`, and that box is a sibling of
the field's label inside `fc` - which is already the cell's full height. So 100%
was the whole cell, the label's block pushed the box down by its own height, and
the box hung out the bottom **over whatever element came next**.

Concretely: three stacked fields where the middle one has `height: 100%`. Before,
its box overruns its cell by 28px (an 18px label plus its 10px gap) and the
"Below" label renders on top of the Email box's bottom border. After, each field
sits inside its own cell.

The box now shrinks out of the label's way rather than claiming the cell whole:
`fc` distributes its own height, and the box takes `flex: 1 1 auto` -- a basis of
`applyHeight`'s own `height: 100%` -- with the `min-height` floor it already had.

It is scoped to the input-box types, whose `fc` is a label stacked above a box.
Other field types spend `fc` differently -- a checkbox lays it out as an inline
row -- so a column there would restack them. And both callbacks emit their
neutrals for a non-percentage unit rather than nothing, per RESET SEMANTICS, so
a mobile pixel override can take the box back off the flex path.

**A basis of `auto`, not `0`, and that distinction is load-bearing.** The rule is
keyed on the unit and cannot know whether the percentage resolved. Under a Fit
container it does not -- `height: 100%` falls back to `auto` -- and a basis of `0`
there makes the box contribute nothing to its own cell's intrinsic height, so the
cell collapses to the `min-height` floor. Measured against the shipped
declarations in headless Chrome, inside a genuine Fit row:

| percentage-height text area in a Fit container | box height |
| --- | --- |
| `master` | 147 |
| basis `0` | **50** |
| basis `auto` (this branch) | 147 |

A 20-row text area went from 392px to 50px on the basis-`0` revision. A
single-line input is unaffected either way, its content already sitting under the
floor. `auto` is identical to `0` everywhere the percentage does resolve -- the
box is still `cell - labelBlock`, and a short cell still stops at the floor.

### 2. A content alignment cancelled a pinned label's reserve

A `shrink_top` label is a fixed overlay at the box top, so the value is held
clear of it by the `height / 3` reserve - which is why a floating-label field
renders its value a little below the box midline. Setting only
`content_vertical_align` dropped that reserve and handed placement to the
alignment, so **"middle" re-centred the value into the label's ink and no option
in the panel could put it back**. The offset is the placement, not an accident.

The reserve is now a floor on the top padding rather than something an alignment
cancels, threaded through `inputBoxVertical`, `valueLineY` and `inputValueDelta`
so the input, the resting label and the chevron/flag cannot disagree. "middle"
is a no-op on an untouched field again for the value and the resting label --
the reserve moves the value, never the label -- and "top" tucks under the label
instead of through it:

```
placeholder_transition = "shrink_top"     input padding      value midline
  content_vertical_align unset            16.6667/6          30.33px   (original)
  ""                                      16.6667/6          30.33px   same
  center                                  16.6667/6          30.33px   same as original
  flex-start                              16.6667/14.1333    26.27px   below the label
  flex-end                                     24.8/6        34.40px   unchanged
```

Because `""` now lands on the unset geometry too, no builder change is needed:
the `allowUnselect` gap in `AlignContentInput` no longer affects correctness.

**"No-op" is about the value and its label, not the whole field.** Setting an
alignment also emits the inline icon transforms (chevron, eye, tooltip, phone
flag), which move `(reserve - 6) / 2` onto the value's line. Untouched, nothing
is emitted and those icons sit at the box midline while the value sits below it.
So `center` leaves the value and label exactly where they were and brings the
icons into agreement with them -- a change, and the right one, but not nothing.

Running the value under the label stays available by storing
`inner_padding_top` explicitly, which is a separate path.

**Two things worth a reviewer's attention:**

- The floor is gated on a placement actually having been asked for. Without that
  gate, the phone flag's anchor - the one call site with no
  `verticalPlacementAsked` guard - emitted a `transform` on every untouched
  floating-label phone field. The suite caught it.
- A multiselect keeps its reserve when an alignment is set. `applyInputBoxAlignment`
  places an input's value by padding the same target the reserve uses, so the
  reserve stands down for it -- but a multiselect's chips are laid out by flexbox
  on its value container instead, so standing down there dropped the reserve
  outright and left nothing holding the chips clear of the label.
- This reverses one decision `innerPadding.spec.ts` recorded as deliberate:
  top-aligned no longer sits at the raw 6px padding under a pinned label. That
  call was made for an explicitly stored `inner_padding_top: 0`, which still
  behaves that way. The assertion is updated with the rationale.

### 3. `isDirectionColumn` threw on an unstored flex direction

It reaches `.includes` through `apply()`, which runs as soon as any of its keys
is set - so a theme carrying `text_align` without `flex_direction` hands it
`undefined`. Only the button tiers seed `flex_direction` today, which is the
whole reason this has never thrown in production, and it is one theme migration
away from being real.

- [x] Tested end to end

Verification, all against the shipped components in headless Chrome:

- minimised repro for #1: 10/10 cases green; the box becomes `cell - labelBlock`
  (216 to 188), and the pixel and fit cases stay byte-identical at 120 and 30
- the same rig across both bases and three container kinds: `auto` matches `0`
  wherever the percentage resolves (216 to 188, spill 28 to 0) and preserves the
  content height where it does not (147 and 392, against 50 for basis `0`)
- property loop over random nested layouts: **0 red / 150** at desktop 1100px
  **and** mobile 390px, down from 36/200 before
- desktop-to-mobile unit transitions: **0/8** where a desktop declaration leaks
  past a mobile override
- a percentage-height checkbox is untouched: `fc` stays `flex/row`
- differential loop for #2: red to green, with the unset geometry preserved
  exactly -- input padding and resting label both
- `yarn jest`: **218 suites, 3276 tests, all passing** (3 skipped)
- `yarn lint`: clean
- `yarn typecheck`: clean

### End to end, on a real hosted form

Verified through the whole stack, not just the components: a throwaway form in
the backend, served by `hosted-forms-next` on its `/forms/<domain>/<slug>`
route, rendered by a real `rollup` build of this branch (`FEATHERY_LOCAL_SDK`).
Baseline and fixed builds came from the same worktree with only this change
between them, measured at an identical 900x620 viewport against identical form
data:

Measured at **two** viewports, because a single-viewport run is what let a
mobile regression through an earlier revision of this branch. The field under
test carries a desktop percentage height with a mobile pixel override, so the
mobile column exercises the reset path directly.

| | desktop, baseline | desktop, this branch | mobile, baseline | mobile, this branch |
| --- | --- | --- | --- | --- |
| box height | 207 | 176 | 120 | 120 |
| box `flex` | `0 1 auto` | `1 1 auto` | `0 1 auto` | `0 1 auto` |
| box spill past its cell | **+31** | **0** | 0 | 0 |
| paints over the field below | **+15** | -16 (a gap) | -16 | -16 |
| email `padding-top` | **6px** | **16.6667px** | 6px | 16.6667px |
| email value midline | **25** (box centre) | **30.33** (the offset) | 25 | 30.33 |

At mobile the box takes its pixel override on both builds and `flex` falls back
to `0 1 auto`, which is the neutral doing its job: the desktop flex path does
not leak past the breakpoint. The throwaway form and its rows were removed
afterwards.

`yarn typecheck` was reporting 4 errors in `featheryClient/index.ts` and
`formContext.ts`, all `HubActionOptions` members. They were never from this
change and needed no code fix -- `package.json` pins
`@feathery/client-utils@0.23.0`, which declares `idFieldId`, `idValue` and
`verification`, and they only appeared against a `node_modules` still holding
0.21.1. On a correct install typecheck is clean.

## Related pull requests

Follows #1760 (themeable inner padding for input-box fields), which introduced
the `content_vertical_align` path fixed in #2 here.


src/elements/styles.ts
```diff
@@ -943,6 +969,50 @@ export default class ResponsiveStyles {
...
+   * A shrunken label is a fixed overlay at the box top. It takes no room in the
+   * content box, so the value has to be held clear of it -- which is what the
+   * height/3 reserve has always done, and what makes the value sit a little
+   * below the box's midline on a floating-label field. An alignment is not a
+   * statement about padding, so it must not cancel that reserve: without this
+   * floor, "middle" re-centres the value into the label's ink and no option in
+   * the panel can put it back. Zero wherever there is nothing to clear, so
+   * every other field emits exactly what it emits today.
+   */
+  private pinnedTopFloor(
```
trebitowski, line 985: **Advisory.** `pinnedTopFloor` takes 8 positional `any` params and is called from 6 sites, each reshuffling the same values into a different order for `inputBoxVertical` / `valueLineY` / `inputValueDelta`. Those three also each repeat `Math.max(paddingSide(padTop, legacy), topFloor)`. A small `flooredPadTop(type, padTop, floor)` helper would remove the triplication, and matching this function's parameter order to the placement functions would remove the reshuffle at every call site.

src/elements/styles.ts
```diff
@@ -1285,14 +1383,53 @@ export default class ResponsiveStyles {
...
+   * The box is a sibling of the field's label inside 'fc', and 'fc' is already
+   * the cell's full height -- so a percentage height on the box resolves to the
+   * whole cell, the label's block pushes it down, and the box hangs out the
+   * bottom over whatever element follows.
+   *
+   * Scoped to the input-box types, whose 'fc' is a label stacked above a box.
+   * Other field types spend 'fc' differently -- a checkbox lays it out as an
+   * inline row -- and a column there would restack them.
+   */
+  applyInputBoxLabelFlow() {
```
trebitowski, line 1408: **Advisory.** The docblock plus the two inline blocks in this method run to about 24 lines and carry measured regression numbers (123px to 50px, 392px). Per the comment policy those belong in the PR description. The one sentence the code needs is: basis `auto` rather than `0`, because under a Fit container the percentage does not resolve and a zero basis collapses the box to its min-height floor.

trebitowski (commented):
Reviewed with a multi-perspective pass across this PR, #1836 and feathery-frontend#3961.

No blocking findings. Verified in code:
- Every input-box component renders `fc` as a bare div with no inline `display`, so the `display: block` neutral for px/fit heights changes nothing, and the flex column only appears for a `%` height.
- All six `pinnedTopFloor` call sites pass arguments in the declared order.
- `isDirectionColumn` guard and test are fine.

Process note: #1836 is stacked on this branch, and the react workflows only trigger on PRs targeting master, so #1836 gets no CI until this merges and its base retargets.

Two advisory comments inline.

src/elements/styles.ts
trebitowski: Left as is in this PR: it is a refactor with no behaviour change, and the call-site ordering was verified correct. Fine as a follow-up.

src/elements/styles.ts
trebitowski: Left as is: style only, not a defect.

FranklinGao123:
I measured this branch end to end on a real hosted form.

The rig is a local backend and a `rollup` build of this branch. `hosted-forms-next` serves the build with `FEATHERY_LOCAL_SDK`. Headless Chrome measures at 1100px and 390px. Every page loaded its SDK chunks from `feathery-react-local` and none from npm.

The percentage-height repro (`zz-e2e-height-repro`, field `e2e_middle`):

| build | box | `flex` | spill past its cell |
| --- | --- | --- | --- |
| `master` | 207 | `0 1 auto` | +31 |
| this branch | 176 | `1 1 auto` | 0 |

Mobile stays at 120 with `0 1 auto` on both builds.

A percentage-height text area inside a Fit container (`e2e_fit_area`):

| build | box height |
| --- | --- |
| `master` | 147 |
| basis `0` (the earlier revision) | 50 |
| basis `auto` (this branch) | 147 |

The Fit case is the reason the basis is `auto`. I also measured the rest of the stack with this flex path in place. The tables are on #1836 and feathery-org/feathery-frontend#3961.

**Retested on the builder canvas as well**, which is where the report was made. The "before" row renders with the builder's pinned `@feathery/react` 2.76.0. The "after" row renders the `551342b1` build through a vite alias. The structure is the same 400px stretched column with 50px, 100% and fill fields, cloned into a test panel.

| canvas, desktop | 100% field | its cell | spill past the cell | over the next field |
| --- | --- | --- | --- | --- |
| before, 2.76.0 | 208 | 208 | +32 | +16 |
| after, this stack | 176 | 208 | 0 | 0, with a 16px gap |

The 2.76.0 build has no `data-feathery-input-box` marker, so its row measures the input element itself. The input sits inside the box, so the two rows compare within a border width.

Mobile carries a 120px override on this field and was never affected. It measures 120 in a 152 cell with spill 0 on both builds.

PR #1836, opened 2026-09-03
feat: reserve what a pinned label occupies, not a third of the box

FranklinGao123:
Stacked on #1834 -- review that one first.

## Changes

The room behind a pinned floating label was `height / 3`, which is not a
property of the label at all. A shrunken label's footprint is its `marginTop`
(half the shrunken font) plus its line box, which inherits the field's font
size:

```
font  minFont  label ends   marginTop + font_size
   8        8          12   4 + 8   = 12
  10       10          15   5 + 10  = 15
  16       10          21   5 + 16  = 21
  28       10          33   5 + 28  = 33
```

That does not change with the box, so a ratio drifts. Measuring where the value
text actually lands:

```
height/3                              label's footprint
  height=  50  off-centre by   5.3      off-centre by 7.5
  height=  90  off-centre by  12.0      off-centre by 7.5
  height= 160  off-centre by  23.7      off-centre by 7.5
  height= 250  off-centre by  38.7      off-centre by 7.5
  height= 500  off-centre by  80.3      off-centre by 7.5
```

At 50px the ratio reserved 17px for a label needing 21. At 500px it reserved
167px and left the value 80px below the box's centre. The footprint holds the
value a steady 7.5px below centre -- just clear of the label -- at any height.

The reserve is now that footprint, clamped so a short box still fits a line of
text (`height - line - 6`); below about 45px the label wants more room than the
box has to give. A text area keeps its own `min(font_size, 10) * 2.5`, which was
already a constant and never drifted.

It is also a **whole number**. `inner_padding_top` is an integer column -- the
backend declares it `serializers.IntegerField` -- so a fractional reserve is a
value the builder can only store by rounding. At `height: 40` the clamp lands on
14.8 and at `font_size: 9` the footprint is 13.5; unrounded, the field rendered
one number while the panel showed and seeded another. Rounding inside
`shrinkLabelReservePx` keeps the synthesized reserve, the floor and the stored
value identical by construction.

## The floor, per breakpoint

`pinnedTopFloor` reads the same numeric definition (`shrinkLabelReservePx`)
rather than recomputing a ratio of its own, so the floor and the padding it
floors cannot describe different geometry. Making that actually true took the
corrections below. The first two are in `51755a46`; review added `96427911` and
`b0cf867a`; the last is `551342b1`.

**It has to read the breakpoint's own typography.** The floor reached around
`apply()` for `font_size` and `line_height`, which are always the desktop values,
while the reserve it is meant to match uses the ones `apply()` resolved. On a
200px field with a 16px desktop font and a 28px mobile font the label ends 33px
down at mobile, the floor claimed 21, and "middle" pulled the value 12px up into
the label's ink -- the exact defect #1834 introduced this floor to prevent,
surviving at the one breakpoint nothing measured. Both values already reach every
call site through `VERTICAL_PLACEMENT_KEYS`, so they are threaded through rather
than read off the element.

`placeholder_transition` stays a desktop read, deliberately:
`applyPlaceholderStyles` decides whether to pin a label from that same key and
never routes it through `apply()`, so resolving it per breakpoint here would leave
the floor disagreeing with the label it is reserving for. A comment says so, since
it otherwise reads like the same mistake one line above. That the renderer ignores
a mobile `placeholder_transition` at all is a separate, pre-existing divergence
from the builder canvas: #1840.

**It must not be re-floored at the block reset.** The aligned top padding was
`max(floor, 6)`, but the floor is already clamped to what the box has to give, so
raising it back overflowed a short box -- 6 plus a 30px line plus 6 needs 42px of a
40px field -- and left the centred value 2px below where the untouched field renders
it. The floor now stands as the clamp left it.

**A missing `font_size` has to fall back, not become `NaN`** (`96427911`, from
review). `shrinkLabelFootprint` had no numeric guard, so a theme that omits
`font_size`, or stores `'16px'`, emitted `padding-top: NaNpx`. The browser drops
that declaration and the value renders under the label. `fontSizePx` resolves the
font once, falling back to the 16px backend default, and feeds the footprint, the
text area branch and the line-height fallback. Three rows joined the shared table.

**A reserve the clamp drove to zero is a floor of 0, not no floor** (`b0cf867a`).
`pinnedTopFloor` returned 0 for both, and the two need opposite handling: read as
the absence of a floor, a 30px box with a 30px line put the 6px reset back under a
label with no room for it, so untouched rendered 0px and centred rendered 6px. It
now returns `null` for no floor. The same conflation had `paddingTopCss` and
`inputBoxVertical` measuring against different numbers; both read one definition,
`flooredTopPadding`.

**The pinned label's targets have to be merged, not spread** (`551342b1`).
`Placeholder` spread `getTarget('placeholder')` and `getTarget('placeholderFocus')`
into one css object. Each carries its own `@media` block, so the second replaced the
first, and the resting label's mobile `lineHeight` -- the line box the footprint is
computed from -- never reached the page. The spread predates this stack; the
footprint made it visible, because the reserve now assumes the label follows the
field's font at every breakpoint. `getTargets()` merges the blocks, and the three
spreads are one call with the same layering. Measured on the hosted form:

| pinned label at mobile | reserve | label bottom before | label bottom now |
| --- | --- | --- | --- |
| desktop font 16, mobile 28 | 33 | 21 (12px gap) | 33 |
| desktop font 28, mobile 16 | 21 | 33 (12px collision) | 21 |

Desktop fonts 9, 16 and 28 were already flush; a downward mobile override was the
case that collided. The same spread pattern exists in other components; it is
harmless wherever the two targets' mobile blocks do not both matter, and out of
scope here.

## What this moves

Existing floating-label fields whose height is far from 60px, by design -- where
they sit today *is* the drift. A stored `inner_padding_top` still wins, so
anything deliberately placed stays exactly where it is.

Note the value still centres in the room the reserve leaves, because middle is
the default alignment. Setting Align Content to top puts it hard against the
label. Both are one click apart; this PR only changes where the room ends.

### Two consumers describe this reserve

Neither blocks this PR; both are gated on the release that carries it.

- feathery-org/feathery-frontend#3961 stores the reserve as the field's top
  padding, so the style panel shows the number the canvas draws instead of an
  empty box. It reads `shrinkLabelReservePx`'s arithmetic from its own copy --
  the two repos cannot import each other, so both pin the identical case table
  (`the reserve both repos compute`, here and in `boxSpacingHelper.spec.ts`).
- feathery-org/ai-services#812 tells Robin what the reserve costs a field in
  height. Its guidance still recommends ~48px for text-like fields, which was
  comfortable at `height / 3` (16px at 48) and leaves 1.8px of slack at the 21px
  footprint. Robin never writes `inner_padding_top`, so nothing it authors
  changes shape -- only the height it should aim for.

- [x] Tested end to end

`yarn jest`: 218 suites, 3297 tests, 3 skipped -- the three suites #1841 tracks failed under load and pass in isolation. `yarn lint`: clean.
`yarn typecheck`: 0 errors. Browser loops from #1834 all still green, including
the desktop-to-mobile transition matrix.

The invariants above are pinned by tests that fail without their fix. Reverting
the typography threading fails `the reserve a mobile typography override asks
for`, which asserts the mobile padding against the *untouched* mobile padding
rather than against a literal. `Placeholder.spec.tsx` asserts the component reaches
for `getTargets` and never a `getTarget` spread -- jsdom does not materialise
emotion's `@media` rules, so the wiring is what a component test can pin -- and
`merging the placeholder targets for a pinned label` pins that `getTargets` keeps
both targets' mobile declarations in one block.

### End to end, on a real hosted form

Measured through the whole stack -- local backend, a real `rollup` build of this
branch behind `hosted-forms-next` via `FEATHERY_LOCAL_SDK`, real Chrome at 1100px
and 390px -- on a field carrying a desktop `font_size: 16` with a mobile override
to 28, and on a 40px field with a 30px line:

| computed `padding-top` | `master` | first commit only | this branch |
| --- | --- | --- | --- |
| pinned label, untouched, desktop | 66.67px | 21px | 21px |
| pinned label, centred, desktop | 6px | 21px | 21px |
| pinned label, untouched, **mobile** | 66.67px | 33px | 33px |
| pinned label, centred, **mobile** | 6px | **21px** | **33px** |
| short box, untouched | 13.33px | 4px | 4px |
| short box, centred | 6px | **6px** | **4px** |

Untouched and centred now agree at both breakpoints, which is the whole rule.
Every value is a whole pixel, which is what the column can store. The rig was
validated first by reproducing #1834's own published figures -- box 207 with a
+31 spill on `master`, 176 with no spill on that branch. Every measured page
loaded its SDK chunks from `feathery-react-local`, none from npm, and carried a
string unique to the build under test.


src/elements/styles.ts
```diff
@@ -183,25 +183,71 @@ const blockPaddingCss = (type: string, value: any) =>
...
-  type === 'text_area'
-    ? Math.min(fontSize, SHRINK_LABEL_MAX_FONT_SIZE) * 2.5
-    : Number(height) / 3;
+// The font size a field renders with when the theme stores none -- the backend
+// serializer's own default. The reserve needs a number even then: every sibling
+// helper falls back through isNum, and an unguarded NaN here became
+// `padding-top: NaNpx`, which the browser drops and the label overlaps the value.
+const DEFAULT_INPUT_FONT_SIZE = 16;
+
+const fontSizePx = (fontSize: any) => {
```
trebitowski, line 192: **Critical, fixed.** `shrinkLabelFootprint` did `Math.min(fontSize, 10) / 2 + Number(fontSize)` with no numeric guard, unlike every sibling helper in this file. A shrink_top field whose theme stores no `font_size`, or stores `'16px'`, emitted `padding-top: NaNpx`. The browser drops the invalid declaration, so the value renders under the label with no error anywhere. `master` emitted height/3 for the same input, so this was a regression. The backend serializer allows `font_size` to be omitted when inherited, so the input is reachable.

src/elements/styles.ts
```diff
@@ -183,25 +183,71 @@ const blockPaddingCss = (type: string, value: any) =>
...
+export const shrinkLabelFootprint = (fontSize: any) => {
+  const size = fontSizePx(fontSize);
+  return Math.min(size, SHRINK_LABEL_MAX_FONT_SIZE) / 2 + size;
+};
+
+// The room a pinned floating label needs behind the value when the theme sets no
+// top padding of its own: exactly what the label occupies, so the value starts
+// just under it.
+//
+// This used to be height/3, which is not a property of the label at all. The
```
trebitowski, line 210: **Advisory.** This 18-line block is mostly history and measurements (17px at 50px, 167px at 500px, the 0.2px rounding story). Per the comment policy that belongs in the PR description. The code needs one sentence: the label's own footprint, clamped so a short box still fits a line, in whole pixels because `inner_padding_top` is an integer column.

src/elements/styles.ts
```diff
@@ -183,25 +183,71 @@ const blockPaddingCss = (type: string, value: any) =>
...
+): number => {
+  const size = fontSizePx(fontSize);
+  if (type === 'text_area')
+    return Math.round(Math.min(size, SHRINK_LABEL_MAX_FONT_SIZE) * 2.5);
+  const footprint = shrinkLabelFootprint(size);
+  // A percentage or fit height has no pixel box to clamp against, so the
+  // footprint stands on its own.
+  if (heightUnit !== 'px' || !isNum(height)) return Math.round(footprint);
+  const line = inputLineHeight(lineHeight, size);
+  const room = Number(height) - line - RESET_INPUT_PADDING_Y;
```
trebitowski, line 239: **Advisory.** The clamp subtracts the flat `RESET_INPUT_PADDING_Y` (6), but `legacyPaddingY('dropdown_multi')` is 8, so a short pinned-label multiselect can be handed a reserve 2px larger than the box has room for. The frontend mirrors the same 6, so the panel and the renderer still agree with each other.

src/elements/styles.ts
```diff
@@ -183,25 +183,71 @@ const blockPaddingCss = (type: string, value: any) =>
...
+  return Math.round(Math.max(0, Math.min(footprint, room)));
+};
 
 const shrinkLabelReserve = (
```
trebitowski, line 243: **Advisory.** `shrinkLabelReserve` is now a wrapper that only appends `px`, with a single caller in `applyPlaceholderStyles`. It could be inlined.

src/elements/styles.ts
```diff
@@ -244,10 +290,10 @@ const MULTISELECT_LEGACY_PADDING = {
 };
 
 // A shrink_top label is a fixed overlay pinned to the box top: it takes no room
```
trebitowski, line 303: **Important, fixed.** This comment said "nothing about the label enters the geometry below" and that running the value under the label's ink "is allowed by design". Both are contradicted by `pinnedTopFloor`, which feeds the label's footprint into every placement helper once an alignment is set.

src/elements/styles.ts
```diff
@@ -975,8 +1021,8 @@ export default class ResponsiveStyles {
    *
    * A shrunken label is a fixed overlay at the box top. It takes no room in the
    * content box, so the value has to be held clear of it -- which is what the
-   * height/3 reserve has always done, and what makes the value sit a little
-   * below the box's midline on a floating-label field. An alignment is not a
+   * reserve has always done, and what makes the value sit a little below the
```
trebitowski, line 1035: **Important, fixed.** The docblock still described "the height/3 reserve" after this PR removed height/3 everywhere else.

src/elements/fields/tests/innerPadding.spec.ts
```diff
@@ -663,22 +673,22 @@ describe('input box inner padding', () => {
...
-    ).toBe('max(8px, 20px)');
+    ).toBe('max(8px, 21px)');
   });
 
   it('keeps the plain reserve on a field whose input owns its padding', () => {
     // payment_method's inner padding belongs to Stripe's card element, so it
     // is not an input-box type and never takes these keys. The reserve stays
-    // the plain height/3 it always rendered with, whatever a theme happens to
-    // store under the key.
+    // the label's own footprint, whatever a theme happens to store under the
```
trebitowski, line 690: **Important, fixed.** The comment said the reserve "stays the plain height/3" directly above an assertion of `21px`.

src/elements/fields/tests/innerPadding.spec.ts
```diff
@@ -2037,3 +2090,111 @@ describe('which fields count as carrying a pinned label', () => {
...
+        { font_size: 28 }
+      ).getTarget('field', true).paddingTop
+    ).toBe('21px');
+  });
+});
+
+// The builder carries its own copy of this arithmetic and cannot import ours,
+// so both repos pin an identical table. Nothing enforces the two stay identical
+// -- each suite only checks its own repo -- so a change to one row here must be
+// made by hand to 'the reserve both repos compute' in feathery-frontend's
```
trebitowski, line 2220: **Important, fixed.** The comment claimed a drift between the two repos "fails a test instead of shipping". It does not: the two tables are independent literals in separate repositories, and nothing checks one against the other. Updating this repo's rows to a new formula keeps this suite green while the frontend's copy stays stale.

src/elements/fields/tests/innerPadding.spec.ts
```diff
@@ -1997,7 +2036,21 @@ describe('which fields count as carrying a pinned label', () => {
...
     expect(centredPhone.paddingTop).toBe(untouchedPhone.paddingTop);
-    expect(centredPhone.paddingTop).toBe(`${60 / 3}px`);
+    expect(centredPhone.paddingTop).toBe('21px');
+  });
+
+  // Every other type only pins a label when there is placeholder text to pin,
+  // so without it there is no label, no reserve and no floor: the field keeps
+  // the block reset whether or not an alignment is set. The builder mirrors
+  // this gate in boxSpacingHelper.tsx (rendersPlaceholder) and must agree.
+  it('draws no reserve for a text field with no placeholder text', () => {
```
trebitowski, line 2096: **Test gap, fixed.** No renderer test covered a shrink_top text field with no placeholder text. That is the invariant the builder now mirrors in feathery-frontend#3961, so it needs to be pinned on this side too.

trebitowski (commented):
Reviewed with a multi-perspective pass. One blocking bug and several stale comments, all fixed in 96427911 which I pushed to this branch; the inline comments below record each finding and what was done.

Also note: **no CI has run on this PR.** The react workflows trigger only on PRs targeting master, so only GitGuardian ran here. Run locally on 96427911 with Node 20: full jest suite 218 suites / 3287 passed (3 skipped), `yarn lint` and `yarn typecheck` clean. CI will run once #1834 merges and the base retargets.

src/elements/styles.ts
trebitowski: Fixed in 96427911: `fontSizePx` resolves the font once, falling back to the 16px backend default and rejecting unit-suffixed strings, and feeds the footprint, the text_area branch and the line-height fallback. Three rows added to the shared table (no font, `'16px'`, text_area with no font). feathery-frontend#3961 mirrors the same fallback in b505f7c83.

src/elements/styles.ts
trebitowski: Left as is: not a defect, and the PR description already carries the same material.

src/elements/styles.ts
trebitowski: Left as is: 2px in a narrow edge case, and both repos agree. If it is changed, `INPUT_BLOCK_PADDING_RESET` in feathery-frontend `src/utils/boxSpacingHelper.tsx` must change with it.

src/elements/styles.ts
trebitowski: Left as is.

src/elements/styles.ts
trebitowski: Fixed in 96427911: rewritten to say a stored padding stands as set, an unset one falls back to the reserve, and the floor keeps an alignment from cancelling it.

src/elements/styles.ts
trebitowski: Fixed in 96427911.

src/elements/fields/tests/innerPadding.spec.ts
trebitowski: Fixed in 96427911.

src/elements/fields/tests/innerPadding.spec.ts
trebitowski: Fixed in 96427911: reworded to say the two copies are kept identical by hand. The frontend table carries the same wording.

src/elements/fields/tests/innerPadding.spec.ts
trebitowski: Added in 96427911: asserts no reserve is emitted, and that a centred field renders the same padding an unpinned field does.

FranklinGao123:
I measured this branch end to end on a real hosted form. The rig is the same as on #1834. Every page loaded its chunks from `feathery-react-local` and none from npm. One chunk carried `flooredTopPadding`, and three carried the `getTargets` call from `551342b1`.

Computed `padding-top` of the input. `a3733e24` is the head that was reviewed.

| field | `master` | `a3733e24` | this branch |
| --- | --- | --- | --- |
| pinned label, untouched, desktop | 66.67px | 21px | 21px |
| pinned label, centred, desktop | 6px | 21px | 21px |
| pinned label, untouched, mobile font 28 | 66.67px | 33px | 33px |
| pinned label, centred, mobile font 28 | 6px | 21px | 33px |
| 40px box with a 30px line, untouched | 13.33px | 4px | 4px |
| 40px box with a 30px line, centred | 6px | 6px | 4px |

Untouched and centred agree at both breakpoints. Every value is a whole pixel.

On this branch, a pinned field with no placeholder text renders 6px and no label. A pinned field with no `font_size` renders 21px. Those are the `96427911` and `b505f7c83` gates.

`551342b1` fixes the label itself. `Placeholder` spread two targets, and the second replaced the first target's `@media` block. The resting label lost its mobile `line-height`.

| pinned label at mobile | reserve | label bottom, `b0cf867a` | label bottom, this branch |
| --- | --- | --- | --- |
| desktop font 16, mobile 28 | 33 | 21 (12px gap) | 33 |
| desktop font 28, mobile 16 | 21 | 33 (12px collision) | 21 |

Desktop fonts 9, 16 and 28 are flush on both builds. No other row moved.

The React workflows run only for PRs that target `master`, so CI has not run here. The full suite is green locally. The three suites #1841 tracks failed under load and pass in isolation.

PR #1844, opened 2026-09-09
fix: decode document filenames in processing toasts

trebitowski:
## Changes

Processing toasts displayed URL-encoded document names such as `003_%E7%99%BB%E9%8C%B2%E6%9B%B8.pdf` instead of `003_登録書.pdf`.

Three code paths derived a filename from a file URL and each did it differently, so the same file could render three ways. They now share one decoder, `getFilenameFromUrl` in `src/utils/fileNames.ts`:

- the processing toast, which previously did not decode at all
- `getFileData` (`utils/browser.ts`), which set the name a download or zip entry saves under
- `fetchS3File` (`utils/formHelperFunctions.ts`), which named rehydrated file fields with `decodeURI` — that left reserved characters such as `%26` encoded and threw `URIError` on a malformed escape

The helper strips query strings and fragments, then decodes the last path segment with `decodeURIComponent`. A malformed escape falls back to the raw segment instead of throwing. Direct upload labels remain literal text.

Validation: full React suite green (220 suites, 3286 tests), including new specs for the helper, the rehydration path, and the saved download name. ESLint and `tsc` clean. Browser end-to-end verification is pending.

- [ ] Tested end to end

## Related pull requests

https://github.com/feathery-org/feathery-backend/pull/3843 preserves Unicode filenames during assistant attachment uploads.

### DIFF
--- a/src/Form/components/ActionToast/ToastItem.spec.tsx
+++ b/src/Form/components/ActionToast/ToastItem.spec.tsx
@@ -1,0 +1,60 @@
+import React from 'react';
+import { cleanup, render, screen } from '@testing-library/react';
+import ToastItem from './ToastItem';
+
+// Jest's classic JSX transform requires React for fragments in imported files.
+const originalReact = (globalThis as any).React;
+beforeAll(() => {
+  (globalThis as any).React = React;
+});
+afterAll(() => {
+  (globalThis as any).React = originalReact;
+});
+
+afterEach(cleanup);
+
+describe('processing toast filenames', () => {
+  it.each([
+    ['003_%E7%99%BB%E9%8C%B2%E6%9B%B8.pdf', '003_登録書.pdf'],
+    ['caf%C3%A9%20%26%20%E6%9B%B8%E9%A1%9E.pdf', 'café & 書類.pdf'],
+    ['%F0%9F%93%84.pdf', '📄.pdf'],
+    ['登録書.pdf', '登録書.pdf'],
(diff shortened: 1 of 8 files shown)


arykhoda:
Decode is only applied in the toast, so the same filename now renders three ways: toast shows 003_登録書.pdf, the download saves as 003_%E7%99%BB%E9%8C%B2%E6%9B%B8.pdf (utils/browser.ts:92, getFileData doesn't decode), and rehydrated file fields use decodeURI (utils/formHelperFunctions.ts:323) which leaves reserved chars like %26 encoded and throws URIError on a malformed escape

trebitowski:
Good catch — fixed by making all three paths share one decoder.

New `src/utils/fileNames.ts` exports `getFilenameFromUrl`, and the toast, `getFileData` (utils/browser.ts), and `fetchS3File` (utils/formHelperFunctions.ts) all call it. That replaces `decodeURI` with `decodeURIComponent`, so `%26` decodes, and the try/catch means a malformed escape falls back to the raw segment instead of throwing URIError during rehydration.

Tests: the decoding table moved to `__test__/fileNames.spec.ts`; added `__test__/fetchS3File.spec.ts` (reserved chars + the `%E7` URIError case) and `__test__/downloadFileNames.spec.ts` (the saved download name); ToastItem.spec keeps three cases for its own wiring. Full suite green (220 suites / 3286 tests), lint and typecheck clean.

arykhoda (approved):

PR #1847, opened 2026-09-10
feat: Data Hub table row filters, find-in-grid, column sort, dropdown chips

trebitowski:
## Changes

A Data Hub-backed Table element can now filter the rows it shows against live form field values. The builder stores `properties.hub_filters` (one entry per filter: hub column, operator, form field); this PR makes the SDK honour them.

- `useHubTableSource` resolves each filter to a hub API `where` condition on every render, reading the compared field from `fieldValues`. The conditions are keyed by content, so the rows reload only when a compared field's value actually changes (the existing focus-resync guards still apply).
- The hub column is addressed by its key, taken from the live hub schema when loaded (so a renamed column keeps filtering) and from the key stored on the filter otherwise.
- Filters combine with AND (every condition is sent in one `where`). Operators: `equals` (default), `not_equals`, `in` / `not_in` (list value), `greater_than`, `greater_than_or_equal`, `less_than`, `less_than_or_equal`, `contains`, `not_contains`, `starts_with`, `ends_with` (text matches are case-insensitive, via the JSON key `i*` lookups), and `is_empty` / `is_filled` (no value). `equals` is sent without an `operator` key, so plain filters keep the request shape they had. `in` / `not_in` send a list: a multi-value field as-is, a single text value split on commas (e.g. a hidden field set from `?ids=a,b,c`). A filter whose field is empty is left off, so an unfilled field behaves like no filter rather than matching nothing. `is_empty` / `is_filled` bind no form field and are always sent.
- A filter whose form field was deleted has no `field_key` (the backend drops those on deletion) and is skipped.
- Unfiltered tables send exactly the same `get` request as before.

**Find in table (spreadsheet mode).** Mod+F while the grid has focus opens a small toolbar floating over the grid's top-right (below the header): a query input, a `n of m` counter, previous/next steppers and a close button. Matching runs on each cell's displayed text, case-insensitively, in reading order; typing jumps to the first match, Enter / Shift+Enter (or the steppers) walk through them with wrap-around, Escape closes and hands focus back to the grid. Matches are tinted amber and the current one darker, applied through the existing cell-shading hook underneath any validation shading. Code: `spreadsheet/useGridSearch.ts` (pure `findMatches` + hook), `spreadsheet/SearchBar.tsx`, wiring in `SpreadsheetTable`, hotkey in `SpreadsheetGrid`; public class names `feathery-table-grid-search*`. Tests in `tests/spreadsheetSearch.spec.tsx`.

**Column sorting (spreadsheet mode).** Right-click a column header for `Sort A → Z`, `Sort Z → A` and, once sorted, `Clear sort`. The sorted header shows ▲/▼ and `aria-sort`. Sorting reuses `useTableData`'s existing sort state (new `setSort`), so spreadsheet mode now always enables it regardless of the classic table's toggle. The row menu and the new header menu share one `ContextMenu` shell.

**Dropdown cells as chips, with a grid-driven menu.** A cell with a fixed set of values draws its value as a chip with a chevron (an empty cell shows a small chevron-only chip). Clicking the chip opens the menu at once; clicking anywhere else in the cell only selects it, so range selection behaves like every other cell. The menu is no longer a native `<select>`: it is a listbox rendered inside the cell (`spreadsheet/ChoiceMenu.tsx`) while keyboard focus stays on the grid — arrows move the highlight, Enter/Tab commit and move on, Escape closes, a letter jumps to the first match, a press outside closes without change, and the chip stays visible under the open menu. This removes the whole class of "the select kept focus and the arrow keys changed the value / reopened the picker" bugs along with the `showPicker`/`:open`-polling/keyup-grace workarounds. Empty option is offered as `(empty)`.

**Resize handle.** The visible line is now 2px, centered on the column's grid line (the 9px grab area is unchanged in spirit).

**Virtualization.** Row overscan raised from 8 to 16 so fast scrolls and find-jumps land on painted rows.

**Review fixes.** A filter change that arrives while edits are buffered is applied once the buffer clears (only if the filters actually changed, so a plain save/discard costs no extra fetch). A filter on a hub column that no longer exists in the loaded schema is skipped instead of failing the read. Find indexes cell text once per data change and only scans strings per keystroke.

**Depends on** feathery-org/client-utils#40 for any non-`equals` operator to reach the backend — on the current 0.23.0 pin the `operator` key is dropped by `dataHubAction`, so those filters degrade to exact matches until this package's `@feathery/client-utils` is bumped to the release containing it. `equals` works today.

- [x] `yarn tsc --noEmit` clean; TableElement suites 295/295 pass, incl. new `tests/hubFilters.spec.tsx`, `tests/spreadsheetSearch.spec.tsx` and `tests/spreadsheetSort.spec.tsx`
- [ ] Bump `@feathery/client-utils` once client-utils#40 is released
- [ ] Tested end to end against feathery-org/feathery-backend#3858

## Related pull requests

- feathery-org/feathery-backend#3858 (validation, `field_key` hydration, `in` operator)
- feathery-org/client-utils#40 (`operator` on where conditions)
- feathery-org/feathery-frontend#4004 (Filters UI)

### DIFF
--- a/src/elements/basic/TableElement/useHubTableSource.ts
+++ b/src/elements/basic/TableElement/useHubTableSource.ts
@@ -31,7 +31,26 @@ type HubRow = {
   errors?: Record<string, string>;
 };
 
-export type HubFilterOperator = 'equals' | 'in';
+export type HubFilterOperator =
+  | 'equals'
+  | 'not_equals'
+  | 'in'
+  | 'not_in'
+  | 'greater_than'
+  | 'greater_than_or_equal'
+  | 'less_than'
+  | 'less_than_or_equal'
+  | 'contains'
+  | 'not_contains'
+  | 'starts_with'
+  | 'ends_with'
+  | 'is_empty'
+  | 'is_filled';
+
+// Operators whose value is a list of alternatives.
+const LIST_OPERATORS: HubFilterOperator[] = ['in', 'not_in'];
+// Operators that test the hub column alone and bind no form field.
+const VALUELESS_OPERATORS: HubFilterOperator[] = ['is_empty', 'is_filled'];
 
 // A builder-configured row filter: keep the rows whose hub column matches the
 // live value of a form field. `field_key` is filled in server-side from
@@ -682,18 +701,29 @@ export function hubFilterWhere(
   if (!filters?.length) return [];
   const conditions: HubWhereCondition[] = [];
   filters.forEach((filter) => {
-    if (!filter.field_key) return;
     const hubFieldKey =
       schemaFields?.find((field) => field.id === filter.hub_field_id)?.key ??
       filter.hub_field_key;
     if (!hubFieldKey) return;
+    const operator = filter.operator;
+    if (VALUELESS_OPERATORS.includes(operator)) {
+      conditions.push({ fieldId: hubFieldKey, operator });
+      return;
+    }
+    if (!filter.field_key) return;
     const raw = values[filter.field_key];
-    if (filter.operator === 'in') {
+    if (LIST_OPERATORS.includes(operator)) {
       const list = hubFilterList(raw);
       if (!list.length) return;
-      conditions.push({ fieldId: hubFieldKey, operator: 'in', value: list });
+      conditions.push({ fieldId: hubFieldKey, operator, value: list });
     } else if (raw != null && raw !== '') {
-      conditions.push({ fieldId: hubFieldKey, value: raw });
+      // Exact match is the Hub's default, so plain filters keep the request
+      // shape they had before other operators existed.
+      conditions.push({
+        fieldId: hubFieldKey,
+        ...(operator === 'equals' ? {} : { operator }),
+        value: raw
+      });
     }
   });
   return conditions;
--- a/src/elements/basic/TableElement/useHubTableSource.ts
+++ b/src/elements/basic/TableElement/useHubTableSource.ts
@@ -663,12 +663,16 @@ function omitKeys(
 }
 
 /**
- * The `where` conditions a table's row filters currently resolve to. The hub
- * column is addressed by key (what the Hub API takes), taken from the live
- * schema when it has loaded so a renamed column keeps filtering, and from the
- * key stored on the filter otherwise. A filter whose form field no longer
- * exists has no key to read and is skipped; the backend drops such filters
- * when the field is deleted.
+ * The `where` conditions a table's row filters currently resolve to. Every
+ * filter must match (the Hub ANDs its conditions). The hub column is
+ * addressed by key (what the Hub API takes), taken from the live schema when
+ * it has loaded so a renamed column keeps filtering, and from the key stored
+ * on the filter otherwise.
+ *
+ * A filter is left off while its form field is empty, so an unfilled field
+ * behaves like no filter (all rows) rather than matching nothing. A filter
+ * whose form field no longer exists has no key to read and is skipped too;
+ * the backend drops such filters when the field is deleted.
  */
 export function hubFilterWhere(
   filters: HubFilter[] | undefined,
@@ -685,15 +689,11 @@ export function hubFilterWhere(
     if (!hubFieldKey) return;
     const raw = values[filter.field_key];
     if (filter.operator === 'in') {
-      conditions.push({
-        fieldId: hubFieldKey,
-        operator: 'in',
-        value: hubFilterList(raw)
-      });
-    } else {
-      // An unset field compares as empty rather than being left off: the Hub
-      // requires a value, and a filter must never widen to every row.
-      conditions.push({ fieldId: hubFieldKey, value: raw ?? '' });
+      const list = hubFilterList(raw);
+      if (!list.length) return;
+      conditions.push({ fieldId: hubFieldKey, operator: 'in', value: list });
+    } else if (raw != null && raw !== '') {
+      conditions.push({ fieldId: hubFieldKey, value: raw });
     }
   });
   return conditions;
--- a/src/elements/basic/TableElement/useHubTableSource.ts
+++ b/src/elements/basic/TableElement/useHubTableSource.ts
@@ -1,5 +1,6 @@
 import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { featheryWindow } from '../../../utils/browser';
+import { fieldValues } from '../../../utils/init';
 import { HubFieldSchema, HubSchema } from '../../components/dataMapping/types';
 import { CellRules, hubCellRules } from './spreadsheet/validation';
 import { CellWrite, Column } from './types';
@@ -30,12 +31,31 @@ type HubRow = {
   errors?: Record<string, string>;
 };
 
+export type HubFilterOperator = 'equals' | 'in';
+
+// A builder-configured row filter: keep the rows whose hub column matches the
+// live value of a form field. `field_key` is filled in server-side from
+// `field_id`/`field_type`, like a column's, and is absent once that field has
+// been deleted.
+export type HubFilter = {
+  hub_field_id: string;
+  hub_field_key: string;
+  operator: HubFilterOperator;
+  field_id: string;
+  field_type: string;
+  field_key?: string;
+};
+
+export type HubWhereCondition =
+  | { entryId: string }
+  | { fieldId: string; value?: any; operator?: HubFilterOperator };
+
 type DataHubAction = (options: {
   hubId: string;
   operation: 'get' | 'create' | 'update' | 'delete';
   entryId?: string;
   data?: Record<string, any>;
-  where?: Array<{ entryId: string } | { fieldId: string; value?: any }>;
+  where?: HubWhereCondition[];
   verification?: HubVerification;
 }) => Promise<any>;
 
@@ -48,6 +68,7 @@ type UseHubTableSourceProps = {
       hidden_hub_fields?: string[];
       readonly_hub_fields?: string[];
       hub_verification?: HubVerification;
+      hub_filters?: HubFilter[];
     };
   };
   client:
@@ -146,6 +167,19 @@ export function useHubTableSource({
   // table's own row filter is the best guess: a verified-only table has
   // nothing to distinguish.
   const showStatusColumn = unverifiedEnabled ?? verification !== 'verified';
+
+  // Row filters read the global `fieldValues`, which is mutated outside React
+  // state, so the conditions are rebuilt every render and keyed by content:
+  // the reference only changes (and the rows only reload) when a compared
+  // field's value actually changed.
+  const hubFilters = element.properties?.hub_filters;
+  const whereKey = JSON.stringify(
+    hubFilterWhere(hubFilters, schemaFields, fieldValues)
+  );
+  const where: HubWhereCondition[] = useMemo(
+    () => JSON.parse(whereKey),
+    [whereKey]
+  );
 
   // Columns derive from the live Hub schema minus the hidden (blacklisted)
   // fields, so fields added to the Hub later show up without republishing the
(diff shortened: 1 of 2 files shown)


src/elements/basic/TableElement/useHubTableSource.ts
```diff
@@ -312,6 +372,22 @@ export function useHubTableSource({
...
+  const blockedWhereKey = useRef<string | null>(null);
+  useEffect(() => {
+    if (blockRefetch) {
+      blockedWhereKey.current ??= whereKey;
+      return;
+    }
+    const stale =
+      blockedWhereKey.current !== null && blockedWhereKey.current !== whereKey;
+    blockedWhereKey.current = null;
+    if (stale) refetch();
```
ayeshah: The filter catch-up after buffered edits is lost on Save: `handleSaveEdits` clears `pendingEdits` and synchronously enqueues writes, so the unblock effect's `refetch()` hits the `pendingRef > 0` guard and returns — but `blockedWhereKey` is still reset to null, so nothing retries when the queue drains. The grid keeps showing rows for the old filter until refocus. Any filter change arriving while a write is in flight is dropped the same way.

src/elements/basic/TableElement/spreadsheet/useGridInteractions.ts
```diff
@@ -569,33 +615,80 @@ export function useGridInteractions(options: GridInteractionOptions) {
...
+    [editing, moveTab]
+  );
+
+  /**
+   * Keys while a choice menu is open. The grid holds focus, so they arrive
+   * here: arrows move the highlight, Enter/Tab commit it (moving on like the
+   * text editor does), Escape closes, a letter jumps to the first match.
+   * Everything else is swallowed so nothing reaches the grid underneath.
+   */
+  const handleChoiceMenuKeyDown = React.useCallback(
```
ayeshah, line 630: If the row hosting an open choice menu is scrolled out of the virtualizer window, CellEditor/ChoiceMenu unmount with no blur or cancel but `editing` stays set — `handleChoiceMenuKeyDown` keeps intercepting every keystroke, arrows move an invisible highlight, and Enter/Tab commit into a cell the user can't see. The menu should cancel (or the grid scroll-lock) when its cell unmounts.

src/elements/basic/TableElement/spreadsheet/SpreadsheetGrid.tsx
```diff
@@ -1011,23 +1128,18 @@ function SpreadsheetCell({
...
-        const target = event.target as HTMLElement;
-        if (target.closest(`.${TABLE_CLASS.gridFillHandle}`)) return;
-        interactions.startEditing(cell.row.id, cell.column.id);
-      }}
       onDoubleClick={() => {
         if (!isEditing) interactions.startEditing(cell.row.id, cell.column.id);
       }}
     >
-      {isEditing ? null : (
+      {isDropdown ? (
```
ayeshah, line 1176: `isDropdown` ignores `chipOpens`, so read-only dropdown cells (readOnlyFieldKeys, canEdit false, hub loading) still render the interactive chip — role=button, "Choose X for row N" label, pointer cursor, hover state — while `openFromChip` early-returns and clicks do nothing. The old code gated the affordance on `canEdit`; the chip's role/label/cursor should be gated on `chipOpens`.

src/elements/basic/TableElement/spreadsheet/useGridInteractions.ts
```diff
@@ -156,6 +165,45 @@ export function useGridInteractions(options: GridInteractionOptions) {
     [scrollToActiveCorner, table]
   );
 
+  /**
+   * Tab order: along the row, then on to the next row's first cell (or back
+   * to the previous row's last). False only past the grid's first or last
+   * cell, where Tab is left to move focus out of the grid.
+   */
+  const moveTab = React.useCallback(
```
ayeshah, line 174: Tab now wraps row-to-row through the whole grid, so keyboard focus can only leave at the very first/last cell — on a large hub table a keyboard user parked mid-grid must Tab through every remaining cell to reach the rest of the form (the Escape hatch is undiscoverable). This is a focus-trap regression from the previous per-row exit (WCAG 2.1.2).

src/elements/basic/TableElement/spreadsheet/useGridInteractions.ts
```diff
@@ -569,33 +615,80 @@ export function useGridInteractions(options: GridInteractionOptions) {
...
+   * here: arrows move the highlight, Enter/Tab commit it (moving on like the
+   * text editor does), Escape closes, a letter jumps to the first match.
+   * Everything else is swallowed so nothing reaches the grid underneath.
+   */
+  const handleChoiceMenuKeyDown = React.useCallback(
+    (event: React.KeyboardEvent<HTMLElement>): boolean => {
+      if (!editing) return false;
+      const choices = choicesFor?.(editing.columnId);
+      if (!choices) return false;
+      const rows = ['', ...choices];
```
ayeshah: The menu's row model `['', ...choices]` is built independently here and in ChoiceMenu.tsx, and only agrees because both prepend the same clear-row today. Any change to one (header row, filtered choices, empty-row position) desynchronizes arrow-key navigation from what's drawn, and Enter would commit a different value than highlighted. Suggest one exported `choiceRows(choices)` helper used by both.

src/elements/basic/TableElement/spreadsheet/SpreadsheetTable.tsx
```diff
@@ -284,45 +337,85 @@ export function SpreadsheetTable({
...
+  // The bar wraps its text on narrow tables, so its height is measured rather
+  // than assumed; the constant only stands in until the first measurement.
+  const barRef = useRef<HTMLDivElement>(null);
+  const [barHeight, setBarHeight] = useState(PENDING_BAR_HEIGHT);
+  useEffect(() => {
+    const bar = barRef.current;
+    if (!bar) return;
+    const report = () => setBarHeight(bar.offsetHeight || PENDING_BAR_HEIGHT);
+    report();
+    const Observer = (featheryWindow() as any).ResizeObserver;
```
ayeshah: This measure-then-ResizeObserver block is duplicated in SpreadsheetGrid (scrollbar height) with only the measured expression differing — a fix to one (box-size options, missing-Observer fallback) will predictably miss the other. A small `useMeasured(ref, measure)` hook covers both.

src/elements/basic/TableElement/useHubTableSource.ts
```diff
@@ -620,3 +696,73 @@ function omitKeys(
   );
   return remaining.length ? Object.fromEntries(remaining) : undefined;
 }
+
+/**
```
ayeshah, line 702: Nit: per the repo comment convention (keep comments to 1–2 lines; longer rationale belongs in the PR description), several multi-line rationale blocks in this PR (this 12-line docblock, the blockedWhereKey and fieldValues comments, similar ones in ChoiceMenu/SpreadsheetGrid/styles) could be trimmed — much of the rationale is already duplicated in the PR description.

src/elements/basic/TableElement/useHubTableSource.ts
```diff
@@ -268,7 +321,14 @@ export function useHubTableSource({
         client.getHubSchemas
```
ayeshah, line 324: If a stored `hub_field_key` is stale (column renamed since publish), the table bricks permanently: the Hub rejects the entries `get`, `Promise.all` rejects, and `setSchemaFields` never runs even though `getHubSchemas` succeeded — so every retry sends the same stale key and the advertised rename recovery is dead code. Awaiting/applying the schema before (or independently of) the entries fetch fixes it.

src/elements/basic/TableElement/useTableData.ts
ayeshah, line 1: (re useTableData.ts:375 — line is outside this PR's hunks) Sort is keyed by column display name: with two columns named the same (builder allows it; `field_key` is the unique id), `columns.find(col => col.name === sortColumn)` sorts the first one's values while both headers render the sort indicator and `aria-sort`. Keying by field_key/column index fixes both.

src/elements/basic/TableElement/useHubTableSource.ts
trebitowski: Fixed in cff6fa2e. The hook now records the filters the loaded rows reflect and refetches once the write queue drains and no edits are buffered, instead of resetting a marker when the block lifts. Regression covers a filter change landing while an update is in flight: one refetch after the queue empties, none after that.

src/elements/basic/TableElement/spreadsheet/useGridInteractions.ts
trebitowski: Fixed in 2183ad61. The grid cancels the editor once its cell has been rendered and then leaves the virtual window, so neither the choice menu nor a text editor is left open on a cell the user cannot see. Regression scrolls the editing row out and back and checks the editor is gone and the arrow keys work.

src/elements/basic/TableElement/spreadsheet/SpreadsheetGrid.tsx
trebitowski: Fixed in 2183ad61. The chip's role, label, click handler, cursor and hover are gated on chipOpens, which now also checks the column's read-only flag, so a read-only dropdown cell renders a plain value pill. Regression covers readonly_hub_fields.

src/elements/basic/TableElement/spreadsheet/useGridInteractions.ts
trebitowski: Addressed in 2183ad61. Tab still walks the rows (that data-entry flow is what the PR added), but the grid now advertises its exit per 2.1.2: aria-describedby points at a hint that is visible while the grid has keyboard focus, and Escape clears the selection, then a second Escape moves focus to the next tabbable element after the grid. Regression checks the hint and that the second Escape lands on the element after the table. If you'd rather have the per-row exit back, that's a small revert of moveTab; say so and I'll do it.

src/elements/basic/TableElement/spreadsheet/useGridInteractions.ts
trebitowski: Done in 8081afc4: one exported choiceRows helper in ChoiceMenu.tsx builds the row model for both the menu and the grid's keyboard handling.

src/elements/basic/TableElement/spreadsheet/SpreadsheetTable.tsx
trebitowski: Done in 8081afc4: both measurements go through a useMeasured(ref, measure, remeasureKey) hook, so the observer setup and the no-ResizeObserver fallback live once.

src/elements/basic/TableElement/useHubTableSource.ts
trebitowski: Trimmed in cff6fa2e, 8081afc4 and 2183ad61: the multi-line rationale blocks this PR added (hub source, ChoiceMenu, SpreadsheetGrid, styles, search) are down to one or two lines; the rationale stays in the PR description.

src/elements/basic/TableElement/useHubTableSource.ts
trebitowski: Fixed in cff6fa2e. The schema is applied whether or not the entries read succeeds; once the renamed key is known the where changes and the hook refetches on its own, so recovery needs no manual retry. Regression covers a stored key the Hub rejects.

src/elements/basic/TableElement/useTableData.ts
trebitowski: Fixed in 2183ad61. Sort state is keyed by columnSortKey (the field key, or the column position when it has none) in useTableData, the classic header, the grid header's aria-sort/indicator and the header menu. Regression covers two same-named columns.

ayeshah (approved):

PR #1848, opened 2026-09-11
feat(docx): add composable document editing primitives

faizalsomani:
## Purpose

- Fix table and section operations that looked correct in the Syncfusion editor but lost or reused the wrong internal SFDT structure when copied.
- Use SFDT as the structural source of truth for copy and rewrite operations. The engine rewrites content-control identities, row bindings, formulas, and section structure before returning a coherent document to Syncfusion.
- Retire the old public `split_table` operation, its `keepRows` special case, obsolete command-revision code, and duplicate one-off execution paths. The same behavior is now composed from shared primitives through one change-set engine.
- Give plain and content-control-backed tables one public operation vocabulary. Robin does not choose a different API based on how the table is implemented.
- Make the frontend authoritative. Every target and input is checked against the live document, and success is reported only after the resulting document is verified.

## Replaces and pairs with

- Replaces Anthony's earlier React work in [#1828](https://github.com/feathery-org/feathery-react/pull/1828). The useful behavior was carried forward onto a fresh branch while obsolete paths and redundant tests were removed.
- Pairs with [ai-services #800](https://github.com/feathery-org/ai-services/pull/800).
- This PR owns SFDT mutation, Syncfusion execution, bindings, formulas, tracked changes, striping, and verification. The AI PR owns Robin's planning and response behavior.

## Capabilities unlocked

- **Split a table:** compose `duplicate_table` with complementary `delete_row` operations. There is no dedicated split operation.
- **Move rows between tables:** compose `insert_row`, `set_cell_text`, and `delete_row` while keeping calculated cells live.
- **Continue editing pending tables:** add, delete, copy, or move rows before earlier table changes are accepted.
- **Create live value columns:** add a column such as Quantity x Rate = Cost using formula bindings rather than fixed model-authored totals.
- **Create live subtotal rows:** bind a subtotal to its table and propagate changes into dependent summary controls.
- **Copy bound tables safely:** copied local controls receive independent identities while controls tagged `global` retain their shared identity.
- **Use the same operations on plain tables:** row, column, cell, formatting, and split requests do not require content controls.
- **Create and copy sections:** use neighboring SFDT section patterns, preserve real Word section boundaries, and support additional requested tables or components.
- **Maintain table appearance:** preserve widths and formatting, and rebuild alternating row stripes after structural operations.
- **Create reusable bindings anywhere:** create input or formula controls in paragraphs as well as table cells.

## What is different from before

- Copies are SFDT structural rewrites instead of only visible editor copies, so internal controls and formula graphs remain coherent.
- The old `split_table` API and bound-only split path are deleted. Splitting is now a reusable primitive chain.
- Plain and bound tables no longer expose separate model-facing operation families.
- Row and column changes trigger formula recomputation, summary propagation, and restriping from the resulting table structure.
- Mixed native Syncfusion and SFDT work is preflighted and verified as one change set, preventing partial edits from being reported as success.
- One user message produces one review group and one review card.

## Architecture

- Robin submits primitives such as `insert_row`, `delete_row`, `duplicate_table`, `set_cell_text`, `insert_column`, `delete_column`, and `create_binding`.
- The frontend indexes the live SFDT and verifies anchors, expected text, table shape, bindings, and stable references before writing.
- Structure-sensitive table and section work uses the SFDT binding engine. Simpler edits can use native Syncfusion behind the same public contract.
- Stable references allow later operations to address a table, row, or column created earlier in the same change set.
- Calculated outputs come from the binding engine and are tracked when Syncfusion supports a content revision.
- Formatting is immediate because Syncfusion does not create formatting revisions.
- Every operation records its primitive, internal route, tracking mode, outcome, details, and warnings.

## Main implementation areas

- `syncfusionDocumentOps.ts`: primitive planning, SFDT routing, preflight, verification, rollback, formulas, and traces.
- `tableStructure.ts` and `tableAppearance.ts`: row roles, totals, widths, and banding.
- `DocxEditor/bindings/`: content-control identity, formula bindings, row adoption, and native/SFDT coordination.
- `tests/headless/`: real Syncfusion coverage against production-like SFDT documents.

## Testing completed

- TypeScript, ESLint, and the production SDK build passed.
- Real Syncfusion headless lane: 19 suites and 61 tests passed.
- Scoped jsdom lane: 48 suites and 1,061 tests passed.
- Chrome QA covered inserting an unaccepted bound row, splitting the table with that row, verifying live totals and summary controls, checking primitive routes, and validating accepted-state striping.
- Additional coverage includes plain tables, scattered-row splits, row moves, copied controls, live columns, tracked subtotals, section creation, rollback, accept/reject, and width handling.

## Known limitations

- Formatting changes cannot appear as tracked revisions because Syncfusion does not support formatting revisions in this path.
- Dependency-aware grouping across several related messages is deferred. This PR keeps one card per message.
- Ctrl+Z behavior around locked calculated controls is unchanged.
- Final exploratory browser acceptance remains with the product owner.

## Review size

- 80 files changed, with 14,734 additions and 3,934 deletions.
- Tests and fixtures account for 6,406 additions and 2,148 deletions.
- This replaces the superseded 32k-line branch and includes the real-engine harness needed to verify Syncfusion-specific behavior.


src/elements/components/DocxEditor/bindings/controller.ts
```diff
@@ -494,23 +635,29 @@ export class ReconciliationController {
...
       } else if (apply === 'structural') {
         const started = Date.now();
         try {
-          patched =
-            this.editor.applyStructuralMutations?.(
-              result.structuralMutations
-            ) === true;
-          if (patched && result.writes.length && this.editor.updateValues)
-            patched = this.editor.updateValues(result.writes) === true;
+          patched = this.inTrackedScope(provenance, () => {
```
ayeshah: The structural mutation can succeed and then `updateValues(result.writes)` can return false. In that case `patched` becomes false, but the failure path only records a diagnostic; it does not undo the native mutation. I reproduced this with an authored `remove-row` for `r-2` and a failed derived write: the live editor no longer contained `r-2`, while `workingSfdt` still did. This is newly reachable for authored commands now that they use the native path. Please roll back the structural edit if the subsequent write fails, or make both steps one transaction.

src/assistant/tools/docx/syncfusionDocumentOps.ts
```diff
@@ -14313,6 +16122,1027 @@ function boundDuplicateTablePlan(
...
+    op.valueType === undefined
+      ? occurrence.def.fieldType
+      : parseType(String(op.valueType), `create_binding:${occurrence.name}`);
+  return {
+    route: 'engine',
+    index,
+    op,
+    anchor: block.anchor,
+    execute(state) {
+      const live = state.index.formulas.get(occurrence.name) ?? [];
```
ayeshah: `scanBindings` stores row-scoped formulas under `table.rows[].bindings`, not `index.formulas`. Consequently, redefining a row formula such as `line_total` through `create_binding` finds zero targets here and returns `binding_redefinition_ambiguous`. There is a second gap in `diffBindingCommands`: it compares only `next.formulas`, so a row formula expression change would produce no live `set-expression` command even after this lookup is fixed. Please resolve and diff these formulas by their table and row identity.

src/elements/components/DocxEditor/bindings/controller.ts
```diff
@@ -341,28 +417,54 @@ export class ReconciliationController {
...
+          at: at + 1,
+          count: insertedBlocks.length
+        });
+      } else if (command.type === 'replace-table') {
+        const table = index.tables.get(command.tableId);
+        if (!table) throw new Error(`table ${command.tableId} not found`);
+        const blocksPath = table.markerPath.slice(0, -1);
+        const at = Number(table.markerPath[table.markerPath.length - 1]);
+        const blocks = getAt(mutated, blocksPath) as SfdtBlock[];
+        const insertedBlocks = insertedTableBlocks(
```
ayeshah: `insertedTableBlocks` always prepends an empty paragraph, which is useful when adding a second table but remains in the document when replacing the only table at this position. In a real-editor probe, replacing an unchanged `costs` table added one blank paragraph and moved its marker one block later. Bound column and blank-row edits use this replacement path, so they can add an unintended blank line and shift later anchors. Please keep the temporary separator from becoming part of the final replacement.

ayeshah (commented):
Three issues found in the native binding edit path; details are inline.

src/elements/components/DocxEditor/bindings/controller.ts
faizalsomani: Fixed in c3600dfb by separating insertion topology from in-place replacement so replacement removes its temporary separator and preserves the original table anchor, with plain and bound topology coverage.

src/elements/components/DocxEditor/bindings/controller.ts
faizalsomani: Fixed in c3600dfb by wrapping structural mutations and derived writes in one rollback boundary that restores both editor and SFDT state on false or throw, with fault-injection coverage.

src/assistant/tools/docx/syncfusionDocumentOps.ts
faizalsomani: Fixed in c3600dfb by resolving and diffing row-scoped formulas through table and row identity so expression changes emit tracked set-expression commands with accept/reject coverage.

ayeshah (approved):

PR #1851, opened 2026-09-14
feat: DOCX editor version history

ayeshah:
## Summary
Session-based version history for the in-form DOCX editor. Each editing session becomes a version: the browser autosaves the document as it changes, diffs its own in-memory snapshots at session close, and uploads a compact final SFDT plus anchored change hunks. A read-only second editor replays those hunks to render per-author redlines.

## What's included
- Autosave engine — session tracker, slice store, debounced scheduler (`useDocxHistorySession`)
- Close-time SFDT diff → per-author anchored hunks (`sfdtDiff`)
- History panel — version list, date grouping, author avatars, rename, restore
- Read-only version viewer — author-coloured highlights + prev/next stepping
- Assistant ("Robin") edits authored and attributed, with a pending-vs-approved visual distinction
- Recent polish: cached version documents, full-height loading skeleton, content-control token population on the highlighted path, durable Robin attribution (stored versions + live "Current" preview), Current → last-session diff fallback, Robin bird avatar (matches the dashboard), Ctrl/Cmd+S saves instead of downloading the raw SFDT, and Restore disabled on the current version

## Testing
- `yarn typecheck` and `yarn eslint` clean
- Full DocxEditor Jest suite: 56 suites / 481 tests green


trebitowski (commented):
Review of the SDK half of version history. AI-assisted review (Claude Code), so please sanity-check anything that looks off — I've split what I traced in code from what needs a real browser or a real document to settle.

Context on what I reviewed: base `ad2df9a00`, head `9dd3890c`. Note that's **6 commits behind** the current PR head (`406e112a`) — the `pending highlights` / `bypass cache on restore` / `font colors` fixes came in after, and they touch `useDocxHistorySession.ts`, so some of the below may already be addressed.

Two housekeeping items first:

- The branch is currently **`CONFLICTING` / `DIRTY`** against master, so it can't merge as-is, and the conflicted files will need re-review after resolution.
- `Validate` is red, but only on commitlint checking the **PR title** — "DOCX editor version history" needs a conventional prefix like `feat:`. `Test` (3799 tests) and `Integration` both pass; `tsc -p tsconfig.typecheck.json` and `eslint ./src` are clean locally too. Worth knowing that `gh pr checks` only surfaces GitGuardian and hides all of this.

The diff engine and the slice/checkpoint model are genuinely nice work, and the degradation paths are thought through. The findings cluster in the session lifecycle rather than the diff logic.

### Blocking — `flush()` can never resolve when the save gate is closed

`history/autosaveScheduler.ts:105-109`:

```ts
if (!options.canSave()) {
  setStatus('blocked');
  arm(idleMs);
  return;            // waiters pushed by flush() are never settled
}
```

Every other exit path reaches `.finally(settleWaiters)` or calls `settleWaiters()` directly; this one does neither, and the re-armed timer just re-enters the same branch. `useDocxHistorySession.ts:339` awaits exactly that promise inside `finalizeSession`, before the closing PATCH and `queueClose`.

So: user types, then hits Save or Ctrl+S while `canSave()` is false — a Robin turn mid-flight (`isAssistantWriting`), or `readOnly` having flipped (the effect at `:553-555` only calls `scheduler.cancel()`, which clears the timer without settling waiters). `finalizeSession` never reaches its PATCH, the version row keeps only its last autosaved docx with no final SFDT and no highlights, and `explicitSave` never resolves — so `handleSave` shows neither a success nor an error toast, and `restoreViewingVersion` hangs after dismissing its modal. A one-line `settleWaiters()` on that path, plus a test — `autosaveScheduler.spec.ts` covers blocked-without-flush and flush-while-clean/dirty, but not flush-while-blocked.

### Blocking — nothing cancels the scheduler on unmount

The hook has five effects; the only cleanups are the idle-timer `clearInterval` and the assistant unsubscribe. There's no `useEffect(() => () => scheduler.cancel(), [])` and no unmount flush.

If the user types and the form advances a step (or the modal closes) inside the 3 s debounce, the pending timer still fires. The hook doesn't re-render on unmount, so `editorRef.current` still points at the `destroy()`ed Syncfusion instance, `canSave()` returns true, and `exportRef.current()` calls `saveAsBlob` on a destroyed editor → rejects → `autosaveScheduler.ts:132-138` sets `dirty` and re-arms → **retries forever at 5/15/45/45 s for the life of the page**, calling `setStatus`/`setSavedAt` on an unmounted component each time. Each armed timer also retains the destroyed editor plus `baselineRef`, `s0Ref`, `preTurnSnapshotRef` and up to `MAX_SLICES = 20` full SFDT strings — in an embedded SDK that's tens of MB leaked per editor mount. The user's last ≤3 s of edits are also lost.

### Blocking — two explicit saves can run `finalizeSession` concurrently

`index.tsx:722-736` has no reentrancy guard, and `onSaveShortcut` (`:356`) isn't gated by `saving`. The upload queue is FIFO but nothing serializes the function bodies.

Type → Ctrl+S → type → Ctrl+S before the first finishes. Save 1 snapshots `F_A` synchronously (correct), then suspends at `await scheduler.flush()`. The editor stays interactive, so the new keystrokes open session B, and the second Ctrl+S starts a second `finalizeSession`. When A resumes it calls `exportRef.current()` **now** — exporting a document that already contains B's edits — and PATCHes it as session A. Version row A ends up holding B's bytes while its uploaded `final_sfdt` describes different content, and the hashes no longer describe the stored docx. The toolbar button is `disabled={saving}`, but `saving` is only true during the inner save, not across `flush()`/`closeQueue`.

### High — autosave bypasses the binding commit that every explicit save runs

Two halves:

- `index.tsx:501-512` constructs `useDocxHistorySession` **without `canSave`** — the option documented at `useDocxHistorySession.ts:65` as the binding soft-gate. It appears nowhere outside the hook, so `canSaveRef.current` is always `undefined` and `:382` returns `true` unconditionally.
- `exportDoc` (`useDocxEditor.tsx:1268-1271`) is a bare `editor.saveAsBlob('Docx')` and doesn't call `bindingsState.commitForSave()`. Every user-initiated path goes through `exportBlockMessage()` → `commitForSave()` first; the autosave path calls `exportRef.current()` directly.

So if the user types into a bound content control and pauses 3 s without moving the caret out, autosave persists bytes where that value was never committed by the binding engine — stale computed formulas, and the version row for that session records the uncommitted state. Documents the engine flags as `severity: 'error'` also get persisted every 3 s with no gate, which is what the gate exists to prevent. This one may be intentional, but the divergence from the explicit-save paths looks accidental.

### High — a stale rejection aborts the next Restore

`useDocxHistorySession.ts:419-423` stores `finalizeSession(meta)` in `finalizeRef` and attaches a `.catch()` to a *new* promise, leaving the rejecting one in the ref. `explicitSave` awaits `finalizeRef.current`, and when no session is open `close()` returns early so `onClose` never fires and the ref still holds the **previous** session's promise.

Concretely: a 5-minute idle close fires and its PATCH fails on a flaky network → status `error`, rejection swallowed but retained. Ten minutes later the user opens History and clicks Restore → `restoreViewingVersion` awaits `historySession.save()` → no session open → it rethrows that old error → `history.restoreVersion()` is never called and the user sees "Could not restore this version" with nothing having been attempted. Mirror image: when the ref holds a *resolved* promise and no session is open, `handleSave` flashes "Document saved" having persisted nothing.

### High — permanent 4xx are retried forever, and there are no timeouts

`autosaveScheduler.ts:132-138` retries any rejection with backoff capped at 45 s, with no attempt ceiling and no status inspection. Two of the backend's errors are permanent for the life of the session:

- `{"session_id": "Session is no longer current"}` — **two tabs, or two collaborators, on one envelope reach this deterministically.** Tab B's autosave creates the newer row, after which every autosave from tab A 400s until its session closes on idle. Tab A's edits are unsaved the whole time and the user is told nothing. This is a conflict expressed as a 400 with no client handling at all.
- `{"session_id": "Session is already closed"}` — same shape if a PATCH races behind the closing PATCH.

Separately, `apiFetch` is a bare `fetch` with no `AbortController`, and `closeQueue` (`:292-302`) is advanced only by `.catch()` — so a never-settling upload never advances it. `finalizeSession` awaits `closeQueue` *before* the closing PATCH, so one hung checkpoint permanently blocks every subsequent session close, and Restore hangs with no error and no recovery short of a reload.

Also worth a look: `:286-289` catches everything on close/checkpoint with a comment assuming "endpoint unreachable", so a 403, a 429, or a validation 400 are all indistinguishable from "not deployed yet" — the version silently loses its highlights and nothing reaches `onError` or Sentry.

### High — the `form_key` addition changes auth semantics on an existing endpoint

`integrationClient.ts:710` now appends `form_key` to the envelope file PATCH whenever `meta?.sessionId` is set, i.e. on every in-form docx autosave. Previously this call carried no `form_key`, and the backend's `IsDualAuthenticated` short-circuits to `True` when no form is named — so adding it means `check_form_auth` now runs on every autosave. A missing or expired `feathery_session_token`, an email that left `allowed_emails`, or an auth-type mismatch now produces a **403 on document save**.

Two things make that worse: the call passes `parseResponse=false`, so the centralized `FormAuthenticationError` → re-auth prompt path never runs (the user just gets a silent `error` status), and it combines with the infinite retry above. This tightens auth against the **old** backend too, so it isn't a backward-neutral SDK release — worth deciding whether to gate it on a session cookie actually being present.

### Medium — "no baseline" and "nothing changed" are the same signal

`useDocxHistorySession.ts:199-201`:

```ts
const startSha256 = snap.s0 ? contentHash(normalizeForDiff(JSON.parse(snap.s0))) : finalSha256;
```

When S0 is unavailable this sends `start_sha256 === final_sha256`, which the backend treats as proof the session changed nothing — it **deletes the row and its files** and returns 204. The comment two lines above says the intended degradation is "docx-only (no highlights)", not deletion, and the closing PATCH has already written that row's docx snapshot by then. Reachability of a null `s0` looks narrow, but sending an explicit empty `start_sha256` (or omitting it) when there's no baseline would remove the ambiguity — the backend already requires both to be non-empty.

### Medium

- **`history/sfdtDiff/ir.ts:340-343`** — `contentHash` does `s.split('').reverse().join('')` over the whole serialized document. For a 5 MB SFDT that's a 5-million-element array of one-char strings plus a reverse plus a join: hundreds of MB of transient allocation and a long main-thread stall, with real OOM risk on mobile Safari. It's called 3× per checkpoint/close (`diffSession`, plus both hashes in `closeSession`), and checkpoints piggyback on autosaves every ≥20 s while typing. A backwards `charCodeAt` loop gives the identical value with zero allocation. Also: the wire fields are `final_sha256`/`start_sha256` but this is two concatenated 32-bit FNV-1a hashes, which is misleading for anyone reading the backend column.
- **`useDocxHistorySession.ts:496-514`** — `collectRobinRuns(JSON.parse(ed.serialize()))` runs inside `onEdit`, wired to Syncfusion's `contentChange`, which fires per change rather than edge-collapsed. A Robin turn emits many of these, each paying a full serialize + parse + recursive walk synchronously in the handler. `preTurnSnapshotRef` already captures one serialize per turn; the per-event capture is the expensive part.
- **`timeBudgetMs` isn't an upper bound** — `sfdtDiff/index.ts:71-74` checks it only *between* slices, so one `applySlice` can overrun the 4000 ms budget arbitrarily. `lcs.ts:14-27` can allocate up to 16 MB synchronously per slice. `previewSession` runs the whole diff synchronously inside a click handler, as does `checkpointSession` on autosave. `applyHunks` deep-clones via `JSON.parse(JSON.stringify(...))`, and in `closeSession` that clone exists purely to compute `changes.trackedAuthors`.
- **`useVersionDocument.ts:40-62`** — `versionCache` is module-scoped, bounded to 8 entries of full display SFDT strings, and never cleared on unmount (only a test-only helper). Eight multi-MB strings stay resident after the editor unmounts, on every page that ever opened the viewer. Per-instance, or clear on unmount?
- **`DocumentEditorContainer.tsx:315-357`** — `saveEnvelope` is now driven every ~3 s by autosave, and each invocation also runs the binding write-back: `setFieldValues(newValues, true, true)` + `await client.submitCustom(newValues)`. The comment at `:346-348` explicitly justifies doing this here "rather than on every reconcile" because "pressing Enter should not cost a network round trip" — autosave has now made it a round trip every 3 s plus a whole-form re-render. `setSaving(true/false)` per autosave also disables the Sign/Download button and flickers Save roughly every 3 s while typing.
- **`index.tsx:757-763`, `:784-787`, `:814-818`** — these changed to `history ? await historySession.save() : ...`, and `historySession.save()` returns `Promise<void>`, so `saveResult` is always `undefined` when history is enabled. `(saveResult as DocxSaveResult|undefined)?.file ?? downloadUrl` then silently falls back to the **pre-save** `downloadUrl` from the render closure. It works today only because the S3 key is reused and the fetch is `cache:'no-store'` — load-bearing on an implementation detail. `onTerminalAction(saveResult)` is likewise always called with `undefined`, which is a silent contract change for any external host using `DocxSaveResult`.
- **`useDocxHistorySession.ts:618-626`** returns a fresh object literal every render, and it's a dependency of the effect at `index.tsx:562-573` and of `selectVersion`. They converge, so no loop — but while viewing Current, each history refresh produces a new `saved` object → `setViewingVersion(saved)` → `useVersionDocument` re-fires, and Current deliberately bypasses the cache, so it re-downloads `final_sfdt` + `changes`, re-gunzips, re-parses, re-runs `applyHunks` and re-`open()`s the viewer, flashing the skeleton. `checkpointSession` calls `setSavedAt` → `historyRefreshKey` → refetch, so this happens on every checkpoint. `useMemo` on the returned object, and something other than `savedAt` as the refresh key.
- **Unvalidated remote data** — `listEnvelopeVersions` returns `response.json()` as implicit `any` and satisfies `Promise<DocxVersion[]>` with no validation, and `useVersionDocument.ts:136-138` casts network bytes straight into `ChangeList` before the hunk applier. The surrounding try/catch contains crashes, so it's a typing gap rather than a crash risk, but it's the natural place for a schema check.
- **`restore` isn't idempotent** — the session id is minted inside the adapter per call (`DocumentEditorContainer.tsx:378`), so a user retrying after an apparent failure creates a *second* restore row plus another snapshot. PATCH and close are correctly idempotent via `get_or_create`.

### Smaller notes

- `authors[].kind` is typed `'user' | 'assistant' | 'other'` but `parse_authors` hard-rejects anything outside `user`/`assistant` → 400 → the document never saves and then retries forever. Latent only because nothing currently passes `currentUser`. Either narrow the type or have the backend ignore unknown kinds.
- The backend maps only `you|user|robin|assistant` as author *keys*, but `VersionAuthor.key` is documented as an arbitrary stable identity and `HistoryPanel.tsx:164` already derives keys from `actor_label`. The first email-keyed `currentUser` will silently produce empty `authors`.
- `renameEnvelopeVersion` and `DocxHistoryHost.renameVersion` are wired up but no UI calls them, and `GET .../versions/<id>/` has no caller either — so the backend's re-presigning endpoint is never used, which means a History panel left open past URL expiry has no recovery path. Either ship the UI or drop them from the PR.
- `(initState as any).collaboratorId` ×3 — the field is properly typed at `src/utils/init.ts:37`, so the cast isn't needed.
- `envelopeId` is declared and documented at `:59` as "reset the session when the underlying document changes" but never passed, so the reset effect runs on `openNonce` alone. The reachable paths do bump the nonce, so no live repro — but the guard is inert. Also `close('reset')` cancels without flushing, so regenerate-with-unsaved-edits drops the last ≤3 s.
- `ejLoader.ts:75-95` leaves the 20 s `waitForDocumentLoad` timeout pending after `documentChange` resolves — one orphaned timer per document open.
- `ejLoader.ts:8-20` adds 9 Syncfusion stylesheets from `cdn.syncfusion.com` with no `integrity`/`crossorigin`. Small extension of an existing CDN trust decision rather than a new one, but SRI would be cheap.
- `sessionTracker.ts:92-107` — `noteEdit` closes the idle session then starts a new one, but it runs on `contentChange`, i.e. after the keystroke landed. So the character that reopened editing is captured in the closing session and attributed to the previous author.

### Needs a real browser or a real document — can't settle these from code

1. **SFDT / tracked-change anchoring.** The value of the feature rests on `applyHunks` re-inserting deleted text as synthetic `Deletion` revisions at `BlockPath`s computed against the *accepted* document, with formatting changes faked as `Insertion` runs. Whether those land on the right characters for real documents is only verifiable by opening actual versions. Specifically worth checking: edits inside table cells, edits in a header/footer, and a block deleted at the end of a section (`nextSurvivingPath`).
2. **The `checkRevisionType` monkey-patch** in `useDocxEditor.tsx` — swapped per `renderText` call and restored in a `finally`. Whether ej2 can re-enter or async-defer that renderer call, leaving the stub installed, is undocumented and version-specific.
3. **Does `contentChange` fire strictly post-edit, and after a full batch?** The whole slice-boundary and `preTurnSnapshot` design depends on it, and on the pre-turn snapshot being the right end of the user's slice. The comments assert it confidently; only a real assistant turn on a real document proves the attribution comes out right.
4. **`created` event guarantees** — `VersionViewer.tsx:148-151` awaits a promise resolved only from the `created` listener, and assigns `containerRef.current` only after it. If `created` never fires (detached or zero-size host?), cleanup destroys a `null` ref and the appended container leaks permanently. The `cancelled` check covers "fires late", not "never fires".
5. **Concurrent `open()` on the reused viewer.** The viewer is keyed only by highlight mode, so switching versions reuses one instance, and the effect's `cancelled` flag can't cancel an in-flight `open()`. Two rapid selections issue two opens and both `waitForDocumentLoad` promises resolve on whichever `documentChange` fires first — in the docx fallback branch that could re-open B's content as "populated A". Read-only, so no data loss, but possibly the wrong document under the wrong version bar.
6. **Performance envelope.** The PR's own spike asserts `< 2000ms` for a 400-paragraph document, i.e. a ~2 s synchronous freeze is within the accepted envelope, and preview runs inside a click handler. Someone should decide with a genuine customer-sized contract whether that's acceptable in an embedded SDK or whether the diff belongs in a worker.
7. **Tab-close behaviour.** Close uploads set `keepalive: false` and there's no `pagehide`/`visibilitychange` handler, so closing a tab mid-session drops up to 20 s of document bytes and 20 s of highlights. The comments treat tab-death as an accepted degrade — worth confirming that's deliberate for a form users navigate away from constantly, rather than adding a `sendBeacon`/`keepalive: true` path.

ayeshah:
Follow-up on @trebitowski's [review](https://github.com/feathery-org/feathery-react/pull/1851#pullrequestreview-5236983005), checked against current head `231cb0b5`.

### Resolved findings / specific symptoms

1. **DOCX/SFDT snapshot mismatch during overlapping saves:** both snapshots now start before the first await, and DOCX exports use independent archives. Regression coverage includes typing during Save and overlapping exports through the real Word exporter. This fixes the reported mixed-snapshot corruption; it does **not** yet serialize the complete session-finalization/upload lifecycle.
2. **Missing baseline being treated as an unchanged document:** missing S0 now sends an empty `start_sha256`, rather than copying the final hash, so it no longer supplies false evidence for deleting the version.
3. **Orphaned document-load timeout:** `waitForDocumentLoad` now clears the timeout and removes its listener when it settles.
4. **First resumed edit included in the previous idle session:** the hook disables post-edit idle closure (`closeIdleOnEdit: false`), with regression coverage for a delayed idle timer and an author switch.
5. **Concurrent opens painting the wrong version:** viewer opens are serialized and superseded selections are discarded. There is a regression test for a slow DOCX import followed by a newer selection.
6. **The specific stale-rejection Restore stall:** Restore now uses `saveForRestore()`, which does not re-await an old rejected finalization when no session is open. This is a narrow fix: explicit Save still needs failed-close recovery, and Restore must properly account for an already-running or failed close before replacing the document.

### Partial fixes — not marking these resolved

- **Unmount:** cleanup now cancels the scheduled timer. However, an in-flight save can reject after cancellation and re-arm a retry; final pending edits also need an explicit lifecycle/persistence policy.
- **Concurrent saves:** snapshot consistency is fixed as above, but closing document PATCHes are not yet serialized across sessions.
- **Current preview refresh:** the previous painted preview is retained while the next version loads, but unnecessary current-version fetching/recomputation remains.
- **Restore retries:** duplicate clicks are blocked while a restore runs, but retries after an uncertain network outcome still need a stable idempotency key.

The blocked `flush()` wait, autosave binding commit/gate, permanent-error retry classification, request timeouts, swallowed history-upload errors, and form-authentication/re-auth handling are still open. The medium/smaller findings and remaining real-document checks are not being claimed as resolved. GitHub also still reports merge conflicts, and the PR title still needs its conventional prefix.

Verification on this head: 733 SDK tests passed (one skipped), 41 companion-backend tests passed, TypeScript passed, and lint reported no errors. These passing tests do not cover all of the open findings. Additional controlled scheduler checks reproduced both the unresolved blocked-flush wait and a retry being re-armed after cancellation.

ayeshah:
@trebitowski Thanks for the deep session-lifecycle review — you reviewed `9dd3890c`, and the head is now `4ed3d8a5`; the persistence-hardening commits in between were largely driven by your findings. Where things stand:

**Blocking — `flush()` never settles when the save gate is closed:** fixed — the blocked path settles its waiters with a typed `DocumentPersistenceError('blocked')`, so explicit Save and Restore surface an actionable error instead of hanging. Covered by scheduler specs.

**Blocking — nothing cancels the scheduler on unmount:** fixed — the hook's teardown cancels the scheduler (no orphaned retry timers holding a destroyed editor or slice snapshots).

**Blocking — concurrent `finalizeSession` runs:** fixed structurally — document writes now go through an ordered `documentSaveQueue`: snapshots are captured synchronously, a failed snapshot stays ahead of newer edits until retried, and two rapid saves can no longer interleave so a session's row can't end up holding another session's bytes.

**High — autosave bypassed the binding gate:** `canSave` is now wired from the container into the hook, so autosave respects the same binding soft-gate as explicit saves.

**High — stale rejection aborts the next Restore:** `explicitSave` no longer awaits a retained promise from a previous session — with no open session it flushes/retries pending artifacts and the write queue, so an old idle-close failure can't fail a later restore, and "Document saved" can't flash without a persist.

**High — permanent 4xx retried forever / no queue recovery:** retries now have a ceiling (`maxAttempts`), failed artifact uploads are tracked and retryable from an explicit Save, and the queue settles waiting callers on failure instead of wedging.

**High — `form_key` auth semantics:** decided and made deliberate rather than conditional — the backend now *requires* form context on all version-history endpoints and verifies the envelope's stored form association (with a logged grace path for legacy envelopes that predate the association). The SDK half landed in `40f15a7e`: form_key is sent unconditionally on every version call including the list/detail reads.

**Mediums:** the `contentHash` reverse-allocation is gone (no more per-checkpoint multi-MB transient arrays). The remaining perf-shaped items (diff off the main thread, viewer cache lifetime, per-checkpoint refresh churn) are on the follow-up list rather than blockers — flagging if you disagree with that triage.

**Your "needs a real browser" list:** we ran a manual in-browser pass against the live backend — real-document anchoring of highlights across sessions, the restore round-trip (including restoring a >30-day-old version, which now serves as the retention-window baseline), highlight rendering after restore, and mixed You/Robin attribution all verified working. Known attribution edge cases are tracked as G1–G4 in our QA matrix. Also new since your review: `4ed3d8a5` fixes a regression where Robin auto-opening Suggested changes left a stale version viewer mounted over the live editor (with a canvas ink-density regression test so a blank editor fails CI).

trebitowski (approved):
Can you cleanup the docs file and anything else like it? LGTM otherwise

PR #1854, opened 2026-09-15
feat: redeem one-time and expiring access links

piran777:
Draft: the `@feathery/client-utils` pin in this PR can only be bumped once client-utils publishes, so this stays a draft until then.

## Summary

Client side of one-time-use and time-expiring form links.

- `init()` accepts `linkToken` or reads `_lt` from the URL; a repeated `_lt` is ignored. The normal user-id path still runs, so a token for another form cannot strand a form.
- Every request to the API and static hosts carries `X-Feathery-Link: <token>[.<device secret>]` and, when a collaborator id is held, `X-Feathery-Collaborator`, from one helper applied in the fetch wrapper, offline replay, the in-form assistant's direct fetches, and registered as the `@feathery/client-utils` request header provider. The CDN is never sent credentials.
- A single-use link shows a Continue screen before redemption; the click redeems, stores the device secret, and remounts the form. Redemption is skipped with a message when the browser cannot persist the secret, so a one-time link is never consumed by a device that cannot keep it.
- Blocked states: the backend's 403 messages render through the existing full-page path; the require-link flag renders "Please open this form from your personal link."; a failed or rate-limited redeem shows a retry message.
- After the server confirms the token belongs to the form, `_lt` is removed from the address bar and the token is stored per form, so reloads and new tabs on that device resume the link session from the plain URL. A 403 saying the presented token is spent (expired, used elsewhere, not valid) clears that token and its device secret. Only a request this form actually sent counts, and a newer token the form holds is never touched, so reopening a revoked link from an old email keeps the re-issued one working.
- `_lt` is excluded from saved URL params and added, with `_locale`, to the login redirect's existing `_slug` allowlist; which other params survive login is unchanged.
- Link credentials are page-wide, so one link-bound form per page is supported (hosted forms renders one). A page with two blocks rather than opening one with the other's token.

## Screenshots

Continue screen

Continue screen

Form after Continue

Form after Continue

Other device blocked

Other device blocked

Link expired

Link expired

Link not valid

Link not valid

Plain id bypass blocked

Plain id bypass blocked

Require-link blocked

Require-link blocked

Redeem failed with retry

Redeem failed with retry

Collaboration form

Collaboration form

Redeem flow GIF

Redeem flow GIF

## Verification

- Typecheck, lint, and the full jest suite at the repo's pre-existing baseline (the only failures are the known TableElement and DropdownMultiField suites caused by stale shared dependencies). 160+ new tests cover init, header construction and precedence on every path, storage, URL stripping and resume, the Continue screen states, the redirect allowlist, offline replay, and the client-utils provider.
- Every guard is mutation tested: disabling any one of them fails a test.
- End-to-end with the backend and hosted-forms branches on a local stack (46 Playwright checks), plus a manual browser pass (see the GIF).

## Dependencies

Depends on `@feathery/client-utils` with `setRequestHeaderProvider`. This PR pins the current published version; bump the pin after client-utils publishes, then publish this SDK. Backend enforcement ships in the companion backend PR.

## Related PRs

- [client-utils: request header provider for access-link credentials](https://github.com/feathery-org/client-utils/pull/41)
- [feathery-backend: one-time and expiring form access links](https://github.com/feathery-org/feathery-backend/pull/3887)
- [hosted-forms-next: keep the access link token out of field prefill](https://github.com/feathery-org/hosted-forms-next/pull/271)
- [feathery-frontend: access link status and require-link setting](https://github.com/feathery-org/feathery-frontend/pull/4031)


src/utils/featheryClient/index.ts
```diff
@@ -681,7 +721,15 @@ export default class FeatheryClient extends IntegrationClient {
...
 
     const response = await this._fetch(url, options);
-    if (!response) return [];
+    if (!response) {
+      // A link the server rejects outright (expired, used elsewhere, revoked,
+      // or missing on a link-bound submission) will never open again, so this
+      // device stops holding on to it. `_fetch` has already turned the 403 into
+      // the blocked page, which is what this visit still shows.
+      if (isRejectedLinkMessage(initState.authenticationError))
+        this.forgetRejectedLink();
```
arykhoda: `authenticationError` is global and sticky, and `_fetch` returns `undefined` without sending once it is set. `handleFormAuthenticationError` remounts every form on the page, so after one form's link 403 the next form remounts, its `fetchSession` short-circuits, this still matches the message the other form caused, and `forgetLinkToken(this.formKey)` deletes a token and device secret that were never at fault. For a redeemed single-use link the secret is unrecoverable, so that link cannot be reopened anywhere.

Should this key off this response rather than the flag? The same shared state shows up above: the restore at 681 only runs while the global `initState.linkToken` is empty, so a second link-bound form sends the first one's token instead of its own.

src/auth/internal/utils.ts
```diff
@@ -49,22 +50,29 @@ export function hasAuthGatedSteps(integrations: any): boolean {
...
-    }
-  }
+  const slug =
+    currentParams.get('_slug') ?? pathname.match(/\/to\/([^/]+)/)?.[1];
+  if (slug) redirectParams.set('_slug', slug);
+
+  // Kept for the same reason the user-id rewrite keeps it: the form has to come
+  // back from the auth provider in the language it was opened in
+  const locale = currentParams.get('_locale');
+  if (locale) redirectParams.set('_locale', locale);
```
arykhoda: Rebuilding from scratch fixes the delete-while-iterating bug, but it also changes what survives a login round trip. The old loop dropped params unreliably, so an unpredictable subset of unrelated ones came back; now exactly `_slug`, `_locale` and `_lt` do. A form relying on a utm or other custom param after auth loses it. Was narrowing to those three meant for every caller, or only for the link case?

src/utils/featheryClient/index.ts
piran777: Yes, fixed. The session and redeem fetches now forget a token only when their own request was actually sent, and only the token that request carried; the form's stored token goes only if it's that same token. That also fixed a single-form case: reopening a revoked link from an older email used to wipe the re-issued link this device had redeemed. "Can only be opened from its access link" no longer forgets anything, since it means no usable token was presented rather than that the sent one is dead.

On the second point, link credentials are page-wide (as is the user id), so one link-bound form per page is what's supported, and hosted forms always renders one. A page with two now blocks without touching either form's stored credentials. Per-form credentials would mean passing the form through client-utils' header provider, which only sees the URL, so I'm leaving that as a follow-up.

src/auth/internal/utils.ts
piran777: Not meant for every caller. I went back to master's loop and only added `_lt` and `_locale` to its allowlist. Allowlisted keys are never deleted, so the token and the form's language always survive login, and which other params come back is unchanged. The delete-while-iterating bug is still on master (`_id` and `_cid` can survive login depending on param order), but deciding which params should come back from auth belongs in its own PR.

arykhoda (approved):

PR #1863, opened 2026-09-18
feat: validate conditional Hub cells and improve spreadsheet row entry

trebitowski:
## Changes

Hub spreadsheet cells now evaluate existing conditional constraint rules while editing, including dependencies on hidden fields and custom error messages. Changing a rule input refreshes the cell error and clears stale server constraint messages using field/rule identity instead of display text; staged rows retain nonblocking validation feedback. Errors without constraint metadata remain intact until their own cells change.

Empty spreadsheets retain the Add row footer when row creation is enabled. Enter on the last row inserts a row and focuses the same column, preserving committed edits and undo history and handling native dropdown selections. Insertions that shift existing rows still reset history safely.

Constraint comparisons match the backend's representations for Any booleans/JSON and file arrays, including filenames and quoted values. Boolean input aliases are coerced separately from rule literals; date inputs normalize to calendar dates while literals retain their full timestamps, matching existing backend semantics.

## Validation

- All 386 table tests across 16 suites passed, including conditional rule and attachment parity, structured error identity, literal semantics, empty-state, keyboard/dropdown, and undo/redo regression coverage.
- Type check and Node build passed. Lint passed with one existing warning in `formContext.ts`.
- Rebuilt and served the SDK in the local dashboard and hosted form; both returned HTTP 200.

- [ ] Tested end to end

## Related pull requests

Backend companion: https://github.com/feathery-org/feathery-backend/pull/3914

Shared error transport: https://github.com/feathery-org/client-utils/pull/42

Release dependency: publish the client-utils companion and bump this SDK's pinned 0.23.0 dependency before releasing rejected-save identity support. Successful staged-warning metadata works with the current transport. Local verification used the companion transport build; no unpublished dependency version is pinned in this PR.


src/elements/basic/TableElement/spreadsheet/constraints.ts
```diff
@@ -0,0 +1,169 @@
...
+      type: fields.get(key)?.type ?? 'any',
+      comparator: condition.comparator,
+      value: String(condition.value ?? '')
+    };
+  };
+  return (owner.constraint_rules ?? []).map((rule) => ({
+    when: (rule.when ?? []).map(resolve),
+    constraint: resolve(rule.constraint),
+    message:
+      rule.error_message || `Field \`${owner.key}\` failed a constraint rule`
```
ayeshah, line 63: Clearing a stale server error relies on an exact string match: `mergeCellErrors` drops a server error only when a client-built message equals it exactly (`constraint.message === message`). Client builds `` `Field \`${owner.key}\` failed a constraint rule` ``; backend builds the same from `field.key` (apps/hub/utils.py:494). They match today, so no live bug — but the default-message path breaks silently if either side drifts: reworded text, a different token (key vs label), i18n/wrapping, or the backend changing a custom message. Then the stale error stays until the next save. Safer: match on a structured identity (e.g. `{field_key, rule_index}` or a `code`), not the display string.

src/elements/basic/TableElement/spreadsheet/constraints.ts
```diff
@@ -0,0 +1,169 @@
...
+    const text = String(value);
+    actual = dateValue(condition.type === 'date' ? text.slice(0, 10) : text);
+    expected = dateValue(condition.value);
+    if (Number.isNaN(actual) || Number.isNaN(expected)) return false;
+  } else if (condition.type === 'boolean') {
+    const text = String(value).trim().toLowerCase();
+    if (!['true', 'false', '1', '0', 'yes', 'no', 'y', 'n'].includes(text))
+      return false;
+    actual = ['true', '1', 'yes', 'y'].includes(text);
+    expected = ['true', '1'].includes(condition.value.toLowerCase());
```
ayeshah, line 164: `actual` (line 120) accepts `true/1/yes/y`, but `expected` here only accepts `true/1`. So a comparison value of `yes`/`y` reads as `false`, and a row value of `yes` gives a phantom violation that never matches the server. Fix: use the same list — `['true','1','yes','y'].includes(condition.value.toLowerCase())`.

src/elements/basic/TableElement/spreadsheet/constraints.ts
```diff
@@ -0,0 +1,169 @@
...
+  let actual: string | number | boolean = hubString(value);
+  let expected: string | number | boolean = condition.value;
+  if (condition.type === 'number') {
+    actual =
+      typeof value === 'boolean' ? Number(value) : Number(String(value).trim());
+    expected = condition.value.trim() ? Number(condition.value) : NaN;
+    if (!Number.isFinite(actual) || !Number.isFinite(expected)) return false;
+  } else if (condition.type === 'date' || condition.type === 'datetime') {
+    const text = String(value);
+    actual = dateValue(condition.type === 'date' ? text.slice(0, 10) : text);
```
ayeshah, line 150: The two sides normalize differently: `actual` is sliced to date-only (`text.slice(0,10)`), but `expected` (line 114) isn't. A comparison value with a time part (`2024-01-05T12:00:00`) compares midnight vs noon and disagrees with the backend. Fix: slice `expected` too when `type === 'date'`.

src/elements/basic/TableElement/spreadsheet/constraints.ts
trebitowski: Addressed in SDK 96f204af and backend a7a5a3a92. Constraint errors now carry `{field_key, rule_index}` and the table uses that identity to invalidate stale errors; it no longer compares message text. Tests cover changed server wording and unrelated rules/unclassified errors sharing the same text.

The shared request helper was dropping rejected-response payloads, so https://github.com/feathery-org/client-utils/pull/42 preserves them using FetchError. That companion must be published and the SDK's pinned dependency bumped before releasing rejected-save identity support; this dependency is recorded in the PR description. Staged success warnings already retain the metadata.

src/elements/basic/TableElement/spreadsheet/constraints.ts
trebitowski: Checked this against the actual backend evaluator. Cell input coercion accepts yes/y/t aliases and trims whitespace, but rule literals intentionally use `.lower() in ("true", "1")`, without those aliases or trimming. Expanding only the SDK RHS would diverge from server validation: a literal `yes` currently means false on the server.

Kept the existing literal semantics, documented the distinction, added alias/case/whitespace parity regressions, and fixed the missing t/f aliases on the input side in 96f204af.

src/elements/basic/TableElement/spreadsheet/constraints.ts
trebitowski: Checked the backend ingestion and evaluator paths: date cell values normalize to their calendar date, while constraint literals retain their full ISO timestamp and offset. Thus midnight versus a noon literal is also unequal on the backend. Slicing the RHS would change existing server comparisons.

Kept those semantics and added regressions for midnight/non-midnight literals and timezone offsets that cross calendar dates, plus a code comment explaining the asymmetry (96f204af).

ayeshah (approved):

